PR #100, opened 2014-01-09
Support shims that are sourced in bash

ibuildthecloud:
This change was done specifically to support virtualenvwrapper which
provides scripts that are supposed to be sourced into your current shell.
The BASH_SOURCE variable can be used determine the current file that is
being included.  BASH_LINENO can be used to determine if the current
execution is an invocations or sourcing of a file.  BASH_LINENO will
be 0 in a script invocation because there are no functions currently
on the stack.  Finally, if your a currently sourcing a file, do not
do "set -e" as that will cause the interactive shell to close on any
error.

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -42,10 +42,10 @@ remove_prototype_shim() {
 create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
 #!/usr/bin/env bash
-set -e
+[ "\$BASH_LINENO" == 0 ] && set -e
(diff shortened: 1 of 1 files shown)


blueyed:
@ibuildthecloud 
ping

aiboy:
Any progress so far? Or at least some workaround?

NorthIsUp:
bump?

mhadam:
Wondering if there's any update on this, or if someone could point me in the right direction. In effect, `virtualenvwrapper` must be installed through a python version outside of `pyenv`?

Issue #149, opened 2014-03-28
Python build of 2.7.6 fails on Ubuntu 14.04 with readline 6.3


cloudforrest:
As noted in issue #126, pyenv fails to build python 2.7.6 with readline support on systems that are using the latest readline 6.3, due to the removal of 'old-style' typedefs in the readline libraries by readline's maintainer of late.

Yet another attempt to build 2.7.6, this time using a patch, but to no avail. 
And yes, the test system certainly has libreadline-dev installed.
Here is the latest portion of the build log showing the errors:

```
building 'readline' extension
/usr/bin/gcc -pthread -fPIC -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I. -IInclude -I./Include -I/home/techpraxis/.pyenv/versions/2.7.6/include -I/usr/include/i386-linux-gnu -I/usr/local/include -I/tmp/python-build.20140328104347.15697/Python-2.7.6/Include -I/tmp/python-build.20140328104347.15697/Python-2.7.6 -c /tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c -o build/temp.linux-i686-2.7/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.o
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c: In function ‘setup_readline’:
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c:914:24: error: ‘Function’ undeclared (first use in this function)
     rl_startup_hook = (Function *)on_startup_hook;
                        ^
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c:914:24: note: each undeclared identifier is reported only once for each function it appears in
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c:914:34: error: expected expression before ‘)’ token
     rl_startup_hook = (Function *)on_startup_hook;
                                  ^
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c:916:36: error: expected expression before ‘)’ token
     rl_pre_input_hook = (Function *)on_pre_input_hook;
                                    ^
/tmp/python-build.20140328104347.15697/Python-2.7.6/Modules/readline.c:919:41: error: ‘CPPFunction’ undeclared (first use in this function)
     rl_attempted_completion_function = (CPPFunction *)flex_complete;
[...]

Issue #126, opened 2014-03-09
Building python 2.7/3.3 on Ubuntu using readline 6.3


ghost:
I have the lastest pyenv version installed on my system and I already had built python 2.7.6 and python 3.3.4, but when I was trying to build python 3.2.5 I got this error:

```
Installing Python-3.2.5...

BUILD FAILED

Inspect or clean up the working tree at /tmp/python-build.20140309191454.12907
Results logged to /tmp/python-build.20140309191454.12907.log

Last 10 log lines:
(cd /home/thiago/.pyenv/versions/3.2.5/share/man/man1; ln -s python3.2.1 python3.1)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named readline

The Python readline extension was not compiled. Missing the GNU readline lib?
```

I manually checked for any missing dependency and double checked with pyenv-doctor and all dependencies were installed. So after seek a lot on google, I built 3.2.5 with an old version of readline: 5.2 and after I tested with readline 6.2 and it was ok, I could build 3.2.5 with theses readline versions, but not with 6.3.

Does anyone knows why this happen? Why can't I build CPython 2.X/3.X with readline 6.2 and 5.2 but not with 6.3?

PR #152, opened 2014-03-29
Fix CPython build problem with Readline 6.3

yyuu:
Add some patches for CPython 2.6, 2.7, 3.1, 3.2 and 3.3 to fix build error with Readline 6.3. This should fix the issues reported in #126 and #149.


Paulius-Maruska:
Hello, I'm Ubuntu 14.04 LTS, I got readline 6.3 installed (with apt-get, official sources).

There are no problems with any of the listed versions above, but when I tried to install python 2.5.6 - I got the message `WARNING: The Python readline extension was not compiled. Missing the GNU readline lib?`.

I wish I didn't have to test my code on python 2.5... :/

PR #278, opened 2014-11-29
Use `command` to run pyenv rehash

kevinburke:
In the event that `eval "$(pyenv init -)"` is called from a function named
pyenv (which I do to get pyenv to load lazily in my shell), evaluating the phrase
`pyenv rehash` will cause the outer function to run again (causing an infinite loop).

This change makes it clear you want the command named pyenv and not a function
which may exist in the environment.

I updated (and ran) the tests and also verified that this change works locally in my zsh shell.


yyuu:
This has been merged in #487. Thanks!

PR #290, opened 2014-12-11
Improve `pyenv version`, if there is one missing

blueyed:
Display the other available ones still.


yyuu:
Sorry for delay! I confirmed this is working as I expected :octocat:

PR #304, opened 2015-01-14
Use better image for Travis build status

jbrudvik:


yyuu:
What's the benefit to use SVG instead of PNG?

jbrudvik:
The image that Travis serves through the .svg url looks much clearer than what is served through the .png link — especially on high-resolution displays.

Here's Travis's announcement of the .svg url and explanation (including comparison images):
http://blog.travis-ci.com/2014-03-20-build-status-badges-support-svg/

yyuu:
Thanks for explanation! Got it  :octocat:

PR #312, opened 2015-02-06
Use default environment for MACOSX_DEPLOYMENT_TARGET

lepture:
I have defined MACOSX_DEPLOYMENT_TARGET myself. pyenv should read my
settings as the default value.

I need to build flat 10.6-intel pythons for compatible wheels.
Related article: http://lepture.com/en/2014/python-on-a-hard-wheel


yyuu:
Sorry for delay. I added a fix and some tests for it in 22ecefd5729d1634bd1c498a43aa30ad87dc351c :disappointed_relieved:

PR #313, opened 2015-02-08
fish shell/ pyenv init information / direct user to place new code at end of file

derwolfe:
Thanks for pyenv - it is great!

I use the fish shell and was having issues getting it to work. I, not thinking, placed the command 

```status --is-interactive; and . (pyenv init -|psub)```

before some of my path jockeying in the config.fish file. After moving the command to the end the config, after my path manipulation logic, pyenv worked perfectly.

This PR changes the informational message for fish shell users by telling them to place the command *at the end of the file*.

Thanks again!

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -50,15 +50,13 @@ if [ -z "$print" ]; then
   esac
 
   { echo "# Load pyenv automatically by adding"
+    echo "# the following to the end of ${profile}:"
+    echo
     case "$shell" in
     fish )
-      echo "# the following to the end of ${profile}:"
-      echo
       echo 'status --is-interactive; and . (pyenv init -|psub)'
       ;;
     * )
-      echo "# the following to ${profile}:"
-      echo
       echo 'eval "$(pyenv init -)"'
       ;;
     esac
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -50,13 +50,15 @@ if [ -z "$print" ]; then
   esac
 
   { echo "# Load pyenv automatically by adding"
-    echo "# the following to ${profile}:"
-    echo
     case "$shell" in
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks for your PR.

I think it not only makes sense for fish, but all shells. Can you update the PR to just change the message for all shells?

derwolfe:
I've updated the PR to apply this message to all shells.

derwolfe:
@blueyed - I just wanted to check in. Is there anything I am missing that is blocking this from being merged in?

Thanks!

blueyed:
Sorry, forgot about it.

I've amended it in e9aa6cf - just a single commit, with fixed commit message.

Thanks!

derwolfe:
Great - Thanks!

yyuu:
I am sorry for not checking this out :disappointed_relieved: 

I think that we don't always have to place pyenv configuration "at the end of" shell's profile. Because there are bunch of shell tools similar to pyenv out there, we can put only one of them at the end, though. Where to put pyenv configuration in shell configuration would be heavily depending what is in user's shell profile. I think that one thing pyenv have to notice to users is that it will manage PATH during the pyenv-init. I believe that managing shell PATH is basically user's responsibility. Put pyenv configuration at the end of shell's profile isn't always true and may confuse users.

blueyed:
You are right.  It should have been "towards the end" probably?

yyuu:
"towards the end" sounds relatively better than "to the end". Although, this is a documentation issue and it might be sufficient if we wrote some doc in README. Anyway it'd be better to send PR to rbenv as well if we need to modify pyenv-init.

yyuu:
Added a note in README where to place `pyenv init -` in the shell configuration file. Anyway, thanks for sending a PR :octocat:

Issue #326, opened 2015-03-02
Use portable PyPy versions


BengtOFFIS:
Currently, pyenv downloads a PyPy binary that is compatible with Ubuntu only on any Linux system. That leads to errors and should be avoided. One way to do this would be to detect the Linux variant and install an appropriate binary. That would however put a lot of burden onto pyenv, because there are a lot of Linux distributions out there and all would need to be detected, a binary version would need to be found and everything maintained. Instead I would suggest to use [the portable PyPy binaries linked at the PyPy Download page](https://github.com/squeaky-pl/portable-pypy). They are only a bit bigger than the source distributions and all errors that occur with them should be reported over at the project portable-pypy.

yyuu:
Sounds interesting. I'd merge if someone sent me a PR, or I may be write it later.

BengtOFFIS:
Thanks for commenting on this. I am glad you like the idea. Sorry, I fear I can not do this myself. I tried but failed to modify the install scripts to download the portable PyPy. If someone wants to give this a shot, there is [a list of all versions of this portable PyPy distribution](https://bitbucket.org/squeaky/portable-pypy/downloads) on BitBucket.

olliewalsh:
> If someone wants to give this a shot

@BengtOFFIS did you miss the pull request I opened a few hours ago?

BengtOFFIS:
Oh, yes, I must have left this tab open. I will have a look at your PR.

PR #329, opened 2015-03-10
portable-pypy support

olliewalsh:
Issue #326: have this from last year, only tested on debian wheezy.

### DIFF
--- a/plugins/python-build/share/python-build/pypy-portable-2.3.1
+++ b/plugins/python-build/share/python-build/pypy-portable-2.3.1
@@ -1,0 +1,16 @@
+case "$(pypy_architecture 2>/dev/null || true)" in
+"linux" )
+  install_package "pypy-2.3.1-linux_i686-portable" "https://bitbucket.org/squeaky/portable-pypy/downloads/pypy-2.3.1-linux_i686-portable.tar.bz2" "pypy" verify_py27 ensurepip
+  ;;
+"linux64" )
+  install_package "pypy-2.3.1-linux_x86_64-portable" "https://bitbucket.org/squeaky/portable-pypy/downloads/pypy-2.3.1-linux_x86_64-portable.tar.bz2" "pypy" verify_py27 ensurepip
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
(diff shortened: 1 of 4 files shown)


BengtOFFIS:
This adds support for pypy2-2.3.1, pypy2-2.4.0, pypy3-2.3.1 and pypy3-2.4.0, which is already worth pulling. However the current PyPy2 release is 2.5.0 so that version would be most useful to have. Also why not support [all past versions of this portable pypy distibution](https://bitbucket.org/squeaky/portable-pypy/downloads)?

Should the portable pypy have its own namespace? That would actually *not* solve the issue I reported in #326. Perhaps the Ubuntu-specific pypy should move to its own namespace, too. So pyenv would install PyPy from source by default, which should always work. If possible, use the precompiled variants.

I already tested the portable distribution under Fedora. I checked out your branch and all versions downloaded and installed as expected. I had to install virtualenvwrapper manually, to avoid error messages when opening a new terminal.

yyuu:
@olliewalsh :+1:

@BengtOFFIS Lack of some versions isn't a problem. Anyone can write patch and send some PRs.

Replace `pypy` namespace by `pypy-portable` would be nice for Linux variants other than Debian/Ubuntu, but it might not be nice for platforms other than Linux such like OS X. Just introducing new namespace of `pypy-portable` would be the simplest way to deal with....

yyuu:
@BengtOFFIS Pushed 292813eca39d3dd18fde2e6c3040071607b25e51 to install `pypy-portable` on unsupported platforms. This might be the best for your needs.

olliewalsh:
> Also why not support all past versions of this portable pypy distibution?

Been months since I looked at this but IIRC the older portable pypy packages did not create the pypy symlink so they would not work with pyenv - this could be fixed up after install I suppose.

olliewalsh:
Closing this PR as it's superceded by 
https://github.com/yyuu/pyenv/commit/292813eca39d3dd18fde2e6c3040071607b25e51

yyuu:
@olliewalsh 292813eca39d3dd18fde2e6c3040071607b25e51 hasn't been merged into master branch yet. Anyway I will merge them later. Thanks for sending PR :octocat:

PR #336, opened 2015-03-24
Update pyenv

autolife:
so you run the pyenv command in anywhere path


yyuu:
I couldn't understand what you want to fix. Please explain the problem in detail if you wanna fix something.

PR #340, opened 2015-03-26
Add pypy-2.5.1

joshfriend:
Portable distros for linux are not yet available, so I opted to print an error for that case.


yyuu:
:+1:

PR #341, opened 2015-03-26
fix pypy-2.5.0-src build directory

joshfriend:
When I tried to install pypy-2.5.0-src I got this error:

```
pushd: pypy-pypy-c6ad44ecf5d8: No such file or directory
```

Looks like when 2.5.0 was added in b9ae43a9, they forgot to change the build directory and just copied the one from pypy-2.4.0-src.


yyuu:
Good catch! Thanks :octocat:

PR #345, opened 2015-03-30
Fix require_distro

joshfriend:
The `require_distro` function loops over all distro names passed to it, but constantly checks the current distro against the first argument. Thus, some pythons (e.g. pypy in this case) will fail to install the official binaries unless you are using whatever the first listed supported distro is, even though the binary supports a wide range of distros (like 12.04 thru 14.04).

On a related note, `pypy-2.5.x` binary versions for `linux64` did not have the correct list of supported distros.


yyuu:
good catch :+1:

PR #347, opened 2015-03-31
fix url and checksum for pypy-2.5.1-src

joshfriend:
Sorry to blow up your inbox today...

Looks like the tag/branch for 2.5.1 moved after I computed the expected checksum. When I downloaded the release, it was at commit [9c4588d](https://bitbucket.org/pypy/pypy/commits/9c4588d), but is now at [e3d046c](https://bitbucket.org/pypy/pypy/commits/e3d046c). I also see a few commits in the log indicating that the tag for 2.5.1 was moved.

At any rate, I switched the url to the bitbucket downloads page since that is what http://pypy.org/download.html links to. This will ensure that the file we download won't change unless they re-upload it manually.


yyuu:
Thanks for many of improvements :octocat: 

this change looks reasonable. I will try to change the download links of older PyPy releases as well.

PR #350, opened 2015-04-07
Add the environment variable PYENV_NOREHASH to disable automatic rehashing

mcallaway:
There are times when we do not want this to happen, such as when we are using 'fpm' to package python modules.

In my case, I am using "pyenv" alongside "rbenv" (and "plenv" and "renv"). In my ruby environment, I'm using "fpm" (https://github.com/jordansissel/fpm). Let's say I have python 2.7.9 in pyenv, and I want to package "pyramid" into a .deb.

```
fpm --debug -f 
    --python-pypi https://pypi.python.org/simple \
    --python-install-lib /opt/pyenv/.pyenv/versions/2.7.9/lib/python2.7/site-packages \
    --python-bin /opt/pyenv/.pyenv/shims/python \
    --python-pip /opt/pyenv/.pyenv/shims/pip -s python -t deb pyramid
```

This will fail because I'm not root and it will fail the auto rehash. With this patch, I can now do this:

```
PYENV_NOREHASH=1 fpm --debug -f 
    --python-pypi https://pypi.python.org/simple \
    --python-install-lib /opt/pyenv/.pyenv/versions/2.7.9/lib/python2.7/site-packages \
    --python-bin /opt/pyenv/.pyenv/shims/python \
    --python-pip /opt/pyenv/.pyenv/shims/pip -s python -t deb pyramid
```

### DIFF
--- a/pyenv.d/exec/pip-rehash/pip
+++ b/pyenv.d/exec/pip-rehash/pip
@@ -19,8 +19,9 @@ export PATH="${PYENV_BIN_PATH}:${PATH}"
 STATUS=0
 "$PYENV_COMMAND_PATH" "$@" || STATUS="$?"
 
-# Run `pyenv-rehash` after a successful installation.
-if [ "$STATUS" == "0" ]; then
+# Run `pyenv-rehash` after a successful installation,
+# unless rehash is explicitly disabled.
+if [ "$STATUS" == "0" -a -z "$PYENV_NOREHASH" ]; then
   case "$1" in
   "install" | "uninstall" ) pyenv-rehash;;
   esac


yyuu:
This would work as expected with current impl of pyenv. I cannot guarantee if `pyenv-rehash` won't run from other place, though.

Just for this case, it'd be sufficient to specify executables in `/opt/pyenv/.pyenv/2.7.9/bin` as `--python-bin` and `--python-pip` as similar as `--python-install-lib`.

```sh
fpm --debug -f 
    --python-pypi https://pypi.python.org/simple \
    --python-install-lib /opt/pyenv/.pyenv/versions/2.7.9/lib/python2.7/site-packages \
    --python-bin /opt/pyenv/.pyenv/versions/2.7.9/bin/python \
    --python-pip /opt/pyenv/.pyenv/versions/2.7.9/bin/pip -s python -t deb pyramid
```

swashcap:
Running into this:

```
caught per expectationpyenv: cannot rehash: /home/rof/.pyenv/shims/.pyenv-shim exists
```

We’re executing multiple python child processes in parallel-ish. Having a `PYENV_NOREHASH` would help us out a ton!

native-api:
The real problem seems to be that Pyenv tries to rehash when it can't/doesn't need to.

I'm not familiar with workings of `fpm`. What does it do under the hood that triggers a rehash? Could you e.g. provide a trace of Pyenv invocations that trigger the issue?

native-api:
> The real problem seems to be that Pyenv tries to rehash when it can't/doesn't need to.

If this is the case, disabling rehashing would be treating the symptoms rather than the cause -- not a good thing for your long-term health.

PR #355, opened 2015-04-14
2.7.9 - Stop '/Library/Python/X.X/site-packages' from being added to sys.path

mrjefftang:
This is just a copy/paste of https://github.com/yyuu/pyenv/pull/292 to support 2.7.9 to solve the issue in  https://github.com/yyuu/pyenv/issues/282


yyuu:
:octocat:

PR #357, opened 2015-04-15
Performance tweaks

blueyed:
Mainly for https://github.com/yyuu/pyenv/issues/356.

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -12,7 +12,7 @@ if [ -n "$PYENV_DEBUG" ]; then
   set -x
 fi
 
-if enable -f "${0%/*}"/../libexec/pyenv-realpath.dylib realpath 2>/dev/null; then
+if enable -f "${BASH_SOURCE[0]%/*}"/../libexec/pyenv-realpath.dylib realpath 2>/dev/null; then
   abs_dirname() {
     local path="$(realpath "$1")"
     echo "${path%/*}"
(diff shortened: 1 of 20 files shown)


yyuu:
Pretty cool :+1: Better performance would be really nice.

Although, separating `bash` processes would be better for simplicity even if it's not the best for performance. I'm not sure performance is how much important for `pyenv` since `pyenv` is basically a command-line tool and will be ran only if some python apps are invoked. Most of time would be consumed by application itself.

blueyed:
> pyenv is basically a command-line tool and will be ran only if some python apps are invoked

Yes, but e.g. with Vim etc, `python` might be called several times, and then it adds up - especially with things like version checking during startup etc, which are expected to be really fast.

I guess this should be first discussed / agreed upon for rbenv.
(btw: see #360 for better merging in the future.. :octocat:)

blueyed:
@yyuu 
Performance is merely an issue, if it gets called often, e.g. with Neovim, where the python interpreter is called multiple times during startup.  But also for things like displaying the pyenv/Python version in your prompt etc.

I've forwarded it to rbenv (https://github.com/sstephenson/rbenv/issues/736), and that seemed to trigger the "anyenv" rewrite in Go.. :)

That means that it won't make it into rbenv.

Given the common merge ancestor we now have, do you think it's worthwile for pyenv, until anyenv would/could replace it possibly?

yyuu:
I think we'd be better to wait for/contribute to anyenv. Until then, I want to keep this not merged to avoid unexpected compatibility issues....

blueyed:
Makes sense, closing it for now.

PR #358, opened 2015-04-15
src/configure: fix sed expression

blueyed:
This has been fixed for rbenv in
https://github.com/sstephenson/rbenv/commit/050f750.

Ref: https://github.com/sstephenson/rbenv/pull/706


yyuu:
:+1:

PR #368, opened 2015-04-29
pip-rehash: handle versions in commands, like "pip2" and "pip3.4"

blueyed:
I've came up with the regexp before I've noticed that only pip and easyinstall are affected, so this could get optimized / simplified maybe.

Also the exporting of `PYENV_REHASH_REAL_COMMAND` feels dirty, but is required for the inner communication.

Fixes https://github.com/yyuu/pyenv/issues/367

### DIFF
--- a/pyenv.d/exec/pip-rehash.bash
+++ b/pyenv.d/exec/pip-rehash.bash
@@ -1,5 +1,13 @@
 PYENV_PIP_REHASH_ROOT="${BASH_SOURCE[0]%/*}/pip-rehash"
-if [ -x "${PYENV_PIP_REHASH_ROOT}/${PYENV_COMMAND##*/}" ]; then
-  PYENV_COMMAND_PATH="${PYENV_PIP_REHASH_ROOT}/${PYENV_COMMAND##*/}"
+PYENV_REHASH_COMMAND="${PYENV_COMMAND##*/}"
+
+# Remove any version information, from e.g. "pip2" or "pip3.4".
+if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.\d)?$ ]]; then
+  PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
+fi
+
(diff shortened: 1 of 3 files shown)


pyenv.d/exec/pip-rehash.bash
```diff
@@ -1,5 +1,13 @@
 PYENV_PIP_REHASH_ROOT="${BASH_SOURCE[0]%/*}/pip-rehash"
-if [ -x "${PYENV_PIP_REHASH_ROOT}/${PYENV_COMMAND##*/}" ]; then
-  PYENV_COMMAND_PATH="${PYENV_PIP_REHASH_ROOT}/${PYENV_COMMAND##*/}"
+PYENV_REHASH_COMMAND=${PYENV_COMMAND##*/}
```
yyuu: Please add surrounding quotes.

blueyed: Will do.

blueyed:
Is it fine otherwise?
I think it should be as light as possible, because it gets used often.

I wonder if the `python` in `create_prototype_shim` should be adjusted, too? (https://github.com/yyuu/pyenv/blob/master/libexec/pyenv-rehash#L49)

yyuu:
Sorry for delay. Looks fine for me now.

blueyed:
@yyuu 

What do you think about this question from me?

> I wonder if the python in create_prototype_shim should be adjusted, too? 

I've opened a PR for rbenv: https://github.com/sstephenson/rbenv/pull/735

yyuu:
@blueyed 
It'd be nice if pyenv-rehash supports the names of executable like `pythonX` and `pythonX.Y` in addition to standard `python`. It's not complete, though. There might be other names like `pypy`, `jython` and others.

I don't think we need to handle all the cases here since I prefer simplicity. An executable named `python` will be treated as the name of `python` in pyenv, and it's `ruby` in rbenv. Basically I think it's sufficient, at least for now.

PR #376, opened 2015-05-12
Fix broken Anaconda3 2.2.0 URLs

thomasjo:
The current Anaconda3 2.2.0 URLs return 404. This PR fixes the problem. Not sure if the current ones ever worked.


yyuu:
Umm. Good catch :+1: Thanks.

PR #378, opened 2015-05-16
pyenv-version-file-read: use sed instead of type/head/awk

blueyed:
I was seeing the following occasionally in scripts:

> …/.pyenv/libexec/pyenv-version-file-read: line 12: type: write error: Broken pipe

This patch hopefully improves/fixes this, and it seems better anyway to
just use sed here.

I've not tested this with OS X's default `sed`.


yyuu:
I confirmed this is working expectedly at least on `sed` of my OS X 10.10.3. BTW, how I can reproduce the "Broken pipe"? I have never seen it on my installation...

blueyed:
I don't know how to reproduce it either.

I have seen it in logs from scripts, where I have redirected output to the log.
But iit does not happen very often.

PR #379, opened 2015-05-18
pyenv-rehash: look for python*

blueyed:
This was rejected for rbenv, where it does not make much sense
(https://github.com/sstephenson/rbenv/pull/735).

Ref: https://github.com/yyuu/pyenv/pull/368#issuecomment-102806837

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -54,6 +54,19 @@ else
 fi
 export PYENV_ROOT
 
+# Transfer PYENV_FILE_ARG (from shims) into PYENV_DIR.
+if [ -z "${PYENV_DIR}" ]; then
+  if [ -n "${PYENV_FILE_ARG}" ]; then
+    if [ -L "${PYENV_FILE_ARG}" ]; then
+      PYENV_DIR="$(abs_dirname "${PYENV_FILE_ARG}")"
+    else
+      PYENV_DIR="${PYENV_FILE_ARG%/*}"
+    fi
+    export PYENV_DIR
+    unset PYENV_FILE_ARG
+  fi
+fi
+
 if [ -z "${PYENV_DIR}" ]; then
   PYENV_DIR="$(pwd)"
 else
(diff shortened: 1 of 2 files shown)


yyuu:
@blueyed Could you explain the actual use case scenario of this?

blueyed:
@yyuu 
The "use case" is that `python2` and `python3` get handled like `python` for this code path (`PYENV_DIR` gets set).

yyuu:
Idealistically all `python`, `python2` and `python3` should be handled evenly. Although, at least with current impl, `python` is the special name for executable in pyenv (and `ruby` for rbenv). Honestly I'm not sure what is the benefit of setting `PYENV_DIR` on invocation of `python2` and `python3` in addition to `python`....

blueyed:
> `python` is the special name for executable in pyenv 

So it's just for internal calls?
Then this PR is void - and I might add a comment instead.  It wasn't really clear to me what this was doing.  I've thought it would be triggered by calling `python3 some.py`, too.

yyuu:
Kind of internal call, or some kind of legacy code. I've just forgotten what that's for....

The PYENV_DIR will be read by ./libexec/pyenv and pyenv will once try to chdir to the directory if it is possible or not. However it will return to original cwd once it suceeded to chdir. So if I understand correctly it isn't so meaningful.

blueyed:
Re-opening.  See https://github.com/yyuu/pyenv/issues/404 for where this is relevant.

yyuu:
Good catch :+1:

I confirmed the behaviour reported on #404, and confirmed that this changes the behaviour to respect local version at symlink destination.

I believe your changes seem to be reasonable and useful for most cases, though it changes current behaviour slightly. Please add tests for this feature, then I'll merge this. (and, I think it's worth to reopen PR on rbenv as well)

yyuu:
I added tests in 438e828eb507536eebe42adee3e896bceb165d8c and merged into master. Thanks for PR.

PR #396, opened 2015-06-25
Update python-build

johnseekins:
Look in the "PYTHON_BUILD_CACHE_PATH" directory for cached files. Not in the current directory.

Also, don't overwrite a cached file that is already there.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -404,7 +404,9 @@ reuse_existing_tarball() {
   local checksum="$2"
 
   # Reuse existing file in build location
-  if [ -e "$package_filename" ] && verify_checksum "$package_filename" "$checksum"; then
+  if [ -f "$PYTHON_BUILD_CACHE_PATH/$package_filename" ] && verify_checksum "$package_filename" "$checksum"; then
+    echo "Found a cached copy of $package_filename in $PYTHON_BUILD_CACHE_PATH"
+    echo "Copying and un-tarring"
     return 0
   fi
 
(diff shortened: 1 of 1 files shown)


yyuu:
I'm not sure what you want to do. `PYTHON_BUILD_CACHE_PATH` isn't current directory by default. It's at `${PYENV_ROOT}/cache`.

https://github.com/yyuu/pyenv/blob/114063495f08a5c3f53d9eba8238e32959832296/plugins/python-build/bin/pyenv-install#L178

johnseekins:
Okay.
Here's the problem I was running in to:

I downloaded CPython 2.7.10 *before* running pyenv for the first time. I copied the 2.7.10 tarball in $(PYENV_ROOT)/cache (where we both agree it should look for files) and it didn't recognize it was there.
At first, I assumed this was because I named the file incorrectly. But the file downloaded and saved by pyenv was named exactly the same.
By making sure that the code in plugins/python-build/bin/python-build definitely looked in the cache directory:
OLD CODE:
if [ -e "$package_filename" ] && verify_checksum "$package_filename" "$checksum"; then
NEW CODE:
if [ -f "$PYTHON_BUILD_CACHE_PATH/$package_filename" ] && verify_checksum "$package_filename" "$checksum"; then

Basically, you only look in the current directory for cached files.

Later in that same function, you *do* define $package_filename to be the full path. So I suppose an alternative would be to just move that to the top of the function.

Honestly, as I look at this, I'm going to change my pull request some, but I do think it's valid.

PR #401, opened 2015-07-06
Allow per python installation system libraries.

karolciba:
Set LD_LIBRARY_PATH to python installation root path, by this
allow system to load libraries installed not only in default
system paths. Therefore allowing user to keep different
versions for each python installation.

### DIFF
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -32,6 +32,7 @@ fi
 export PYENV_VERSION
 PYENV_COMMAND_PATH="$(pyenv-which "$PYENV_COMMAND")"
 PYENV_BIN_PATH="${PYENV_COMMAND_PATH%/*}"
+PYENV_LIB_PATH="${PYENV_BIN_PATH%/*}/lib"
 
 OLDIFS="$IFS"
 IFS=$'\n' scripts=(`pyenv-hooks exec`)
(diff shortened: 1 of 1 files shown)


karolciba:
If one would like to compile and install, for example wxPython, and keep this library only for targeted python installation (not system, global wide) eg:
python build-wxpython.py --build_dir=../bld --prefix=$HOME/.pyenv/versions/2.7.10/ --install

Then only way to use this library is to set LD_LIBRARY_PATH in shell, which seems strange and inconvenient:
LD_LIBRARY_PATH=$HOME/.pyenv/versions/2.7.10/lib/ python demo/demo.py

thomasjo:
Setting `LD_LIBRARY_PATH` is generally considered harmful. It should only ever be set when experimenting during development. It breaks all kinds of environments, especially ones with multiple versions of the same library. 

Link against the library properly instead.

yyuu:
Agreed with @thomasjo. Setting `LD_LIBRARY_PATH` is harmful. Plus, it will be effective only for GNU/Linux.

In python-build, it sets RPATH to executables to skip setting up `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH`. I think same would be available for other shared libraries.

https://github.com/yyuu/pyenv/blob/36c5cbf74f484016ebc49cb4961982483fd88525/plugins/python-build/bin/python-build#L1749-L1755

Nevertheless you need to set `LD_LIBRARY_PATH`, you can setup it in custom _exec_ hook of pyenv. See also https://github.com/yyuu/pyenv/wiki/Authoring-plugins#pyenv-hooks about hooks.

PR #405, opened 2015-07-08
Use $PWD instead of $(pwd)

blueyed:
The former should be faster, because it does not invoke a subshell.

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -33,7 +33,7 @@ resolve_link() {
 }
 
 abs_dirname() {
-  local cwd="$(pwd)"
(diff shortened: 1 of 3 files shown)


yyuu:
Is there rbenv PR?

blueyed:
Not that I know of - haven't checked.

yyuu:
I couldn't find one in rbenv PRs. I think same can be applied to rbenv and it is definitely useful for them. Please send a PR to them as well.

blueyed:
Ok, will do - but not during the next days probably.
Will close this PR here for now - it should get merged then from rbenv.

PR #410, opened 2015-07-17
Display version origin with non-installed versions

blueyed:
This is useful as an indicator where it is coming from.

Submitted for rbenv at: https://github.com/sstephenson/rbenv/pull/756,
but pyenv has more tests in this regard.


yyuu:
This must be merged by #487

yyuu:
This has been merged in #487. Thanks!

PR #471, opened 2015-11-04
List Anaconda-style virtual environments as a version in pyenv

yyuu:
List virtualenv created in `envs` directory as a version, to improve the interoperability between Anaconda/Miniconda.

```sh
% pyenv versions
  system
* 3.4.3 (set by /home/yyuu/.pyenv/version)
  3.4.3/envs/bar
  3.4.3/envs/foo
  miniconda3-3.16.0
  miniconda3-3.16.0/envs/foo
```


thomasjo:
Nice! I actually had to stop using Anaconda due to the problems I've had with its virtualenvs in conjunction with pyenv. This looks like a step in the right direction!

PR #488, opened 2015-11-21
always handle first file argument as `PYENV_FILE_ARG`

### LINKED ISSUE #479: python shim switches versions depending on python's command line argument
There's something weird happening with the `python` shim as of pyenv 20151105.

Let's say I have this folder structure:

```
35/
    .python-version                   sets a local python 3.5.0
    test.py
    27/
        .python-version               sets a local python 2.7.10
        dummy.txt                       just an empty file
```

All `test.py` does is print the current python version:

```
import sys

print(sys.version)
```

Here's what happens if I call test.py by itself:

```
~/35$ python test.py 
3.5.0 (default, Nov 10 2015, 13:26:40) 
[GCC 4.8.4]
```

As expected, since that's the local version I set with `~/35/.python-version`.

But if I just add the dummy file on the command line, pyenv switches to a different version!

```
~/35$ python test.py 27/dummy.txt 
2.7.10 (default, Nov 10 2015, 15:29:47) 
[GCC 4.8.4]
```

I found this while working in a virtual environment and had to read in a file that was outside the environment. My script failed because it tried to import a package that did not exist in my global python.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28198931-python-shim-switches-versions-depending-on-python-s-command-line-argument?utm_campaign=plugin&utm_content=tracker%2F282009&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282009&utm_medium=issues&utm_source=github).
</bountysource-plugin>

yyuu:
This should fix #479.

TODO: 
1. ~~write some tests~~
2. ~~submit same fix to rbenv~~

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -57,24 +57,17 @@ else
 fi
 export PYENV_ROOT
 
-# Transfer PYENV_FILE_ARG (from shims) into PYENV_DIR.
 if [ -z "${PYENV_DIR}" ]; then
+  # Transfer PYENV_FILE_ARG (from shims) into PYENV_DIR.
   if [ -n "${PYENV_FILE_ARG}" ]; then
-    if [ -L "${PYENV_FILE_ARG}" ]; then
-      PYENV_DIR="$(abs_dirname "${PYENV_FILE_ARG}")"
-    else
-      PYENV_DIR="${PYENV_FILE_ARG%/*}"
-    fi
-    export PYENV_DIR
+    PYENV_DIR="$(abs_dirname "$(dirname "${PYENV_FILE_ARG}")/${PYENV_FILE_ARG##*/}")"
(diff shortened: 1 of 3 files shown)


yyuu:
@blueyed can you review this if you have time?

yyuu:
Submitted to rbenv as sstephenson/rbenv#829

blueyed:
:+1:

native-api (changes requested):
As per https://github.com/pyenv/pyenv/issues/1813#issue-807820300, this is not an adequate fix. The first non-option argument may be an argument to some option, and the set of such options is version- and implementation-specific.

PR #502, opened 2015-12-14
Fix error about symbolic link

mengbo:
Fix the error below

```

$ mkdir testdir; echo "print 'hello'" > testdir/test.py; ln -s test.py testdir/test; python testdir/test
/Users/meng/.pyenv/libexec/pyenv: line 43: cd: test.py: Not a directory
hello

```

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -40,7 +40,7 @@ abs_dirname() {
   local path="$1"
 
   while [ -n "$path" ]; do
-    cd "${path%/*}"
+    [ -d "${path%/*}" ] && cd "${path%/*}"
     local name="${path##*/}"
     path="$(resolve_link "$name" || true)"
   done


yyuu:
1. There is another copy of the function in python-build. It'd be better to update both.
  * https://github.com/yyuu/pyenv/blob/v20151210/plugins/python-build/bin/python-build#L60-L72
2. Please send similar fix to https://github.com/rbenv/rbenv and https://github.com/rbenv/ruby-build

yyuu:
@mengbo Could you rebase to latest master to fix conflict?

And, it's really helpful for both rbenv and pyenv if the `abs_dirname` can handle filename without `/` (==> a file in current directory). Do you have any idea to implement it?

will-Do:
This also fix https://github.com/yyuu/pyenv/issues/580

brbsix:
This fixes #588 as well. Thanks

brbsix:
Any chance this can be merged?

PR #522, opened 2016-01-09
Prevent duplicate PATH entries

oddraisin:
Prevents PATH population in the event the pyenv is initialized twice (e.g. once in shell create, a second time in tmux/screen session).


yyuu:
the PATH you modified is one used only from pyenv internally, and will not expose to user's shell. Anyway please open PR to https://github.com/rbenv/rbenv if you want to change common part of rbenv/pyenv.

Issue #534, opened 2016-02-02
add (optional) option use aria2 to manager/accelerate download [$10]


luzfcb:
if you show me where in the code is located the logic of the download process, I would be happy to send a pull-request

https://aria2.github.io/

another reason I want a download manager is because where I live (northern region of Brazil), sometimes the Internet connection is unstable, so it's impossible to get finished download, because the connection drops for a few seconds. A download manager like [aria2](https://aria2.github.io/) can deal well with connection failure problems

```bash
fabio@luzfcb:~$ pyenv install 2.7.10 -v
/tmp/python-build.20160202070442.19150 ~
Downloading Python-2.7.10.tgz...
-> https://www.python.org/ftp/python/2.7.10/Python-2.7.10.tgz
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
error: failed to download Python-2.7.10.tgz

BUILD FAILED (Ubuntu 14.04 using python-build 20151222-8-g5c5205e)

Inspect or clean up the working tree at /tmp/python-build.20160202070442.19150
Results logged to /tmp/python-build.20160202070442.19150.log

Last 10 log lines:
/tmp/python-build.20160202070442.19150 ~
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104

```

yyuu:
@luzfcb How about adding `--retry` for `curl`? I couldn't find equivalent for `wget`, though.

As a temporal workaround, you can place downloaded archives inside `$(pyenv root)/cache` to prevent `pyenv install` to download archives every time. `python-build` (the backend of `pyenv install`) will use cached archive if it is available

https://github.com/yyuu/pyenv/blob/425e44d064f934b78c8858b6d3ad597f46a06fa8/plugins/python-build/bin/python-build#L416

luzfcb:
I believe that the equivalent ```curl --retry``` for `wget` is:

```
       -t number
       --tries=number
           Set number of tries to number. Specify 0 or inf for infinite retrying.  The default is to retry 20 times, with the exception of fatal errors
           like "connection refused" or "not found" (404), which are not retried.
```

but aria2 is more better que wget, for people like me who suffer from bad connections.

yyuu:
Hmm. It seems aria2 has a feature to resume after failed download. It'd be beneficial to add support in https://github.com/rbenv/ruby-build

luzfcb:
`wget` and `aria2` has "`--continue`" option

[...]

PR #539, opened 2016-02-08
Use `aria2c` as the download if available (#534)

yyuu:
Use `aria2c` as HTTP/HTTPS downloader if it is available. This should solve #534.


yyuu:
rbenv/ruby-build#904 has been merged!

PR #542, opened 2016-02-15
Compile with `--enable-unicode=ucs4` by default for CPython

yyuu:


yyuu:
I'll merge this anyways. I may be going to revert this if any problem reported.

izayoi5776:
looks like can not compile to ucs2 even --enable-unicode=ucs2

PR #543, opened 2016-02-15
Update build scripts for Continuum products

kuanyili:
Download links are changed to HTTPS. This is important especially for latest version of miniconda as there is no SHA256 to keep users from MITM attack.

Anaconda 2.5.0 is added as well.


yyuu:
Good catch :+1: Thanks~

Issue #546, opened 2016-02-29
Missing function fetch_nightly_tarball required for pypy-latest


hackaugusto:
The commit c8ab429d7b0990d901494f9bcd95f6baa5cc51af removed the function `fetch_nightly_tarball`

These recipes use `install_nightly_package`:
- plugins/python-build/share/python-build/pypy-c-nojit-latest
- plugins/python-build/share/python-build/pypy-c-jit-latest

hackaugusto:
I tested all the urls in the `pypy-c-jit-latest`, the compressed file itself has the same name as in the url, but the uncompressed folder has the actual commit, so the recipe won't work by just adding `install_nightly_package` back.

hackaugusto:
Additionally [build_package_pypy_builder](https://github.com/yyuu/pyenv/blob/90e6e30d63a3911f9fad3774e0b94c6050566735/plugins/python-build/bin/python-build#L939) has `python` hardcoded, this could be either python 3 or python 2, if it's python 3 the installation will fail:

```
/tmp/python-build.20160229180917.22779/pypy-dev /tmp/python-build.20160229180917.22779 /mnt/extra/third_party/python/pyenv
  File "rpython/bin/rpython", line 17
    print __doc__
                ^
SyntaxError: Missing parentheses in call to 'print'
```

yyuu:
Gimme PRs

yyuu:
For RPython's Python issue, there is a workaround in `pyenv-install` which will find some Python 2.7 from pyenv. Obviously it will not work with bare `python-build`, though.

https://github.com/yyuu/pyenv/blob/a6f1f480eb6e607d2e5a012873bc7737737dd97f/plugins/python-build/bin/pyenv-install#L203-L224

hackaugusto:
Thanks!

PR #547, opened 2016-03-01
Added pypy-dev special case in pyenv-install to use py27

hackaugusto:
Related to #546

pypy-dev compiles the pypy project too, so it should use a Python2.7 interpreter


yyuu:
Good catch :+1:

PR #549, opened 2016-03-02
rbenv 1.0+

yyuu:
Import changes from rbenv 1.0 (actually I imported changes from the current HEAD instead of tag). Still need testing.


yyuu:
I couldn't find any problem with this at least for now. Please let me know if there's some trouble with this.

PR #577, opened 2016-04-15
Add Jython 2.7.1 beta3

anthrotype:
Hi,

I just copied the build script for [jython-2.7.1b2](https://github.com/yyuu/pyenv/blob/master/plugins/python-build/share/python-build/jython-2.7.1b2) and changed the URL and SHA256 hash.

Thank you,

Cosimo


yyuu:
:+1: Thanks~

PR #592, opened 2016-05-08
Added Portable PyPy 5.1.1

Samureus:

PR #593, opened 2016-05-08
Pypy 5.1.1 src

Samureus:
Meh, this still depends on Portable PyPy 5.1.1 being merged first, sorry, I can send the PR later if needed or delete the previous one.


yyuu:
I'll merge this in favor of #592 since I couldn't find any significant difference between them. Creating topic branch per PR could be better.

Issue #594, opened 2016-05-09
pyenv shims openssl


mfdj:
Question, perhaps a suggestion about openssl. Today I ran `which openssl` and was surprised to see `~/.pyenv/shims/openssl`.

I don't use python regularly but I do use ruby and rbenv regularly so I was surprised to see a tool not written in or specific to python managed by pyenv. I'm not against this, but I think it might be nice to give user's a heads up as it feels like an exceptional side-effect of a language-specific management tool.

This plus the fact that openssl is a keg-only brew: [which can be problematic to have in the PATH?](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-build-software-outside-Homebrew-with-Homebrew-keg-only-dependencies.md). FWIW, I imagine it's more about the numerous header-files than the single `openssl` binary in the path.

Given those two points I'd roughly suggest it might be worth mentioning in the docs somewhere (I couldn't find anything in the repo or wiki). Thanks.

yyuu:
pyenv will create shim for `openssl` if you're installing Anaconda installed. Though it'll be just ignored by pyenv's hook. See output of `pyenv which openssl`.

https://github.com/yyuu/pyenv/blob/2017dd0c25467caffd6b9d6749f80dc502128842/pyenv.d/which/anaconda.bash

```sh
% which openssl
/home/yyuu/.pyenv/shims/openssl
% pyenv which openssl
/usr/bin/openssl
```

mfdj:
Ah, side-effect of Anaconda specifically. Thanks for the clarification and snippet.

PR #595, opened 2016-05-10
Skip creating shims for system executables bundled with Anaconda (#594)

yyuu:
Anaconda bundles some of system executables (e.g. `curl`, `openssl`, etc.). Though for most use case, users will not want to use them until they really want them. This lets pyenv to skip generating shims for those executables.

With this, we'd be able to remove a hook script for `pyenv which` eventually.


thomasjo:
Awesome! 🎉

PR #606, opened 2016-05-21
Don't attempt to read .python-version directories.

roolebo:
My $HOME is located on a share: /company/username. Someone created /company/.python-version directory there and I was getting an error on each pyenv activation:
cut: /company/.python-version: No such file or directory.

That happened because pyenv-version-file wasn't distinguishing regular files and directories, it printed existing matching path and fed it to libexec/pyenv-version-file-read which spilled the error.

### DIFF
--- a/libexec/pyenv-version-file
+++ b/libexec/pyenv-version-file
@@ -9,7 +9,7 @@ target_dir="$1"
 find_local_version_file() {
   local root="$1"
   while ! [[ "$root" =~ ^//[^/]*$ ]]; do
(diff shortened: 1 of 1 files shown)


yyuu:
Could you send same fix to https://github.com/rbenv/rbenv ?

blueyed:
Forwarded: https://github.com/rbenv/rbenv/pull/1003

blueyed:
@yyuu 
Should we merge it here already anyway?

PR #615, opened 2016-05-28
"." has been deprecated in the fish shell in favor of explicit "source"

cha-ron:
running the command suggested by `pyenv init` as it stands (with `.` instead of `source`) in the fish shell doesn't work - fish complains about the [deprecation](https://github.com/fish-shell/fish-shell/issues/310) and otherwise fails. this fixes the problem.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -70,7 +70,7 @@ if [ -z "$print" ]; then
     echo
     case "$shell" in
     fish )
-      echo 'status --is-interactive; and . (pyenv init -|psub)'
+      echo 'status --is-interactive; and source (pyenv init -|psub)'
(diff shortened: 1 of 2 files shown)


yyuu:
Good catch. Could you send same fix to https://github.com/rbenv/rbenv ?

cha-ron:
done

yyuu:
In this case, since it's obvious, I'll merge this now regardless of rbenv's PR status. I may be going to revert in case of trouble, though.

yyuu:
Finished applying similar fix to pyenv-virtualenv. I believe there's no remaining thing in pyenv and plugins. Please let me know (or send PR) if you realize something remaining.

PR #617, opened 2016-05-31
Test with Travis environments

jayvdb:
This change is [Reviewable](https://reviewable.io/reviews/yyuu/pyenv/617)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,149 @@
 sudo: false
-install: git clone --depth 1 https://github.com/sstephenson/bats.git
-script: make test
 language: c
+
+# virtualenv versions:
+# precise: 1.7.1.2
+# trusty: 1.11.4
+# precise: 12.0.6
+# trusty: 13.1.0
+
+addons:
+  apt:
+    # Note that it is not possible to include `virtualenv` from
+    # source `debian-sid`, as pulls in many other debian-sid packages
+    # which conflicts with the Ubuntu packages.
+    packages:
+      - python-pip
+
 env:
-  - PYENV_NATIVE_EXT=
-  - PYENV_NATIVE_EXT=1
+  global:
+    - VIRTUAL_ENV_BASE=~/virtualenv
+    - VIRTUAL_ENV_VERSION_SPEC='<13'
+
+matrix:
+  include:
+    # Plain bats
+
+    - env: PYENV_NATIVE_EXT=
+      addons: false
+    - env: PYENV_NATIVE_EXT=1
+      addons: false
+
+    # CPython
+
+    - env: TESTENV=2.7.11,3.2.6,3.4.4
+      dist: precise
+
+    # Stackless
+
+    - env: TESTENV=stackless-3.4.1,stackless-dev
+      dist: precise
+
+    # PyPy
+
+    - env: TESTENV=pypy-2.4.0,pypy-5.1.1,pypy-c-jit-latest,pypy-c-nojit-latest
+      dist: precise
+
+    # Jython
+
+    - env: TESTENV=jython-2.7.0,jython-dev VIRTUAL_ENV_VERSION_SPEC='==1.7.1.2'
+      dist: precise
+
+    # Anaconda/Miniconda
+
+    - env: TESTENV=anaconda3-4.0.0,miniconda3-4.0.5 VIRTUAL_ENV_ADD_LIB_PATH=1
+      dist: precise
+
+# Failures (must all be replicated into the allow_failures section)
+
+    # Miniconda 2
+
+    - env: TESTENV=miniconda2-4.0.5
+      dist: precise
+
+    # IronPython
+
+    - env: TESTENV=ironpython-2.7.5
+      dist: precise
+      sudo: required
+      addons:
+        apt:
+          packages:
+            - python-pip
+            - mono-xbuild
+            - mono-dmcs
+            - mono-complete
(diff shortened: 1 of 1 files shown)


joshfriend:
Wow this makes the time to run tests unbearably long 😉

jayvdb:
We can remove some of the failures by fixing the failures.
i.e. some of the failures are ilustrative of problems and failed attempts to workaround the problem.
Elapsed time is quite short actually; some projects have several hours elapsed.
If time is a problem, I can set it up to do one successful test job, and do a random config in that job.

.travis.yml
```diff
@@ -1,7 +1,230 @@
...
+
+    - env: TESTENV=pypy-5.1.1 DIST=precise
+      dist: precise
+
+    - env: TESTENV=pypy-5.1.1 DIST=trusty
+      dist: trusty
+
+    # Miniconda
+
+    - env: TESTENV=miniconda3-4.0.5 DIST=precise LD_LIBRARY_PATH=/opt/python/3.5.0/lib/
```
jayvdb: The need for LD_LIBRARY_PATH here seems like it could be fixed by pyenv, but also fixed upstream...?

jayvdb:
Something I havent tried yet is whether the latest version of virtualenv addresses any of these failures.

.travis.yml
```diff
@@ -1,7 +1,230 @@
...
+        apt:
+          packages:
+            - python-virtualenv
+            - mono-xbuild
+            - mono-dmcs
+            - mono-complete
+
+install:
+  - git clone --depth 1 https://github.com/sstephenson/bats.git
+  - |
```
jayvdb: Ironpython install a binary `python` which needs to be run by mono, or use this trick on a sudo enable worker, which is slower than the precise containers. Otherwise virtualenv tries to run it, and fails.
Is it possible to detect this and install a script `python` which invokes mono?

.travis.yml
```diff
@@ -1,7 +1,230 @@
...
+    - env: TESTENV=miniconda2-4.0.5 DIST=precise
+      python: 2.7.9
+      dist: precise
+    - env: TESTENV=miniconda2-4.0.5 DIST=trusty
+      python: 2.7.9
+      dist: trusty
+
+    # Stackless
+
+    - env: TESTENV=stackless-dev DIST=precise
```
jayvdb: Fix : https://github.com/yyuu/pyenv/pull/618

jayvdb:
Elapsed is now 11 mins, with more test environments included.

The previous problem with pypy-dev has been fixed, by installing 2.7.11 first.  Now it dies during the build, and has been excluded.

Note that it is not possible to include `virtualenv` from apt source `debian-sid`, as pulls in many other Debian sid packages which conflict with the Ubuntu packages.

I tested using a the latest virtualenv, and some of the failures are fixed by it, namely 3.4.4 and stackless on `language: generic`.
I have a slight preference to keep them as failures until we know why they are failing on the older virtualenv, but I can integrate the virtualenv upgrade code if we want more successes before this is merged.

yyuu:
@jayvdb just a silly question, but what do you want to test by this? How do you think what is insufficient with stub based testing?

jayvdb:
Well, integration testing finds different types of problems, especially ones that havent been seen before.

Obviously the -dev series of versions will benefit the most from integration tests, as the contents of those branches are unpredictable.  And as seen by #618, integration tests will find these problems quickly when upstream changes.

There is also the benefit that defining integration tests highlights minimum supported versions of other programs that are used together with pyenv, such as virtualenv.  Currently it is obvious that the `virtualenv` with Ubuntu precise and trusty is not supported by all pyenv install-able versions.  There is no hint of that in the README ;-)

Ideally we can install minimum versions of each dependency and run `pyenv install x` for every install-able version, and while it may take a while it should install them all, and basic operations should be possible with each of them.  When they all can be installed and used, and there are unittests for the known problems found, this integration testing could done by a weekly build (not yet possible with Travis) instead of with each commit, as it is only looking for unexpected breakages.

Until then, I believe it is useful to show the known good environments and the known broken environments, so we can focus on fixing these problems, and/or raising them upstream.  What is the point of shipping these install-able versions if we cant get them to work ourselves?

Whether or not you merge, I'm going to keep working on fixing these broken versions, and this set of integration tests will get quicker as more versions have been fixed and dependencies reduced.

.travis.yml
jayvdb: https://github.com/pypa/virtualenv/issues/899

jayvdb:
Testing only on precise without any `language: python`.
Elapsed down to 7.5 mins.

.travis.yml
```diff
@@ -1,7 +1,149 @@
...
+    # Anaconda/Miniconda
+
+    - env: TESTENV=anaconda3-4.0.0,miniconda3-4.0.5 VIRTUAL_ENV_ADD_LIB_PATH=1
+      dist: precise
+
+# Failures (must all be replicated into the allow_failures section)
+
+    # Miniconda 2
+
+    - env: TESTENV=miniconda2-4.0.5
```
jayvdb: this is also fixed with `VIRTUAL_ENV_ADD_LIB_PATH=1`
now just need to get ironpython working... ;-)

yyuu:
This could help preventing the issues like #638.

native-api:
Superceded by 8ed7912759c211d8bf60dff7666cc7b8b6e57373

PR #634, opened 2016-06-21
Use `curl` stub instead of `aria2c` during tests

yyuu:
Sometime `aria2c` requires some _special_ customization (e.g. #620) to call it. Using `curl` during tests could make test impl simpler and understandable..


yyuu:
I'll merge this to apply some `aria2c` related fixes. I may revert this later though.

PR #643, opened 2016-06-27
Add CPython 3.4.5 and 3.5.2

zmwangx:
http://blog.python.org/2016/06/python-352-and-python-345-are-now.html.


robsonpeixoto:
How we can backport this patch ?
bug: https://bugs.python.org/issue26664
patch: https://hg.python.org/cpython/rev/8d7bde14d7a4

yyuu:
`pyenv install` (and `python-build`) supports `-p` which let python-build to read patch from standard input. `cat foo.patch | pyenv install -v 3.4.5` allows users to apply custom patch when building.

https://github.com/yyuu/pyenv/blob/0e298b691d8490064cc069a56be91597afdb590b/plugins/python-build/bin/pyenv-install#L18
https://github.com/yyuu/pyenv/blob/0e298b691d8490064cc069a56be91597afdb590b/plugins/python-build/bin/python-build#L8

I can bundle the patch as a default patch if there's working configuration.

robsonpeixoto:
Thank @yyuu. It's works really well.
`curl https://hg.python.org/cpython/raw-rev/8d7bde14d7a4 | pyenv install -p 3.5.2`

> I can bundle the patch as a default patch if there's working configuration.

I'll love it =D

yyuu:
@robsonpeixoto could you send me a PR if possible? Just putting the patch file under $(pyenv root)/plugins/python-build/share/python-build/patches/Python-5.2 should let python-build to apply the patch before building.

robsonpeixoto:
I'll

robsonpeixoto:
https://github.com/yyuu/pyenv/pull/725

PR #648, opened 2016-06-29
pypy-5.3.1: Remove stray text

JayH5:
Pretty sure these lines are not supposed to be there... causing install to fail for me.


yyuu:
Good catch :octocat:

alex:
Any chance of making a release for this? Thanks!

PR #654, opened 2016-07-04
Use xz archive instead of gzip archive

yyuu:
XZ could be better from perspective of download size.

* [x] Check if the archive is actually available
* [x] Update documentations about dependencies
  * We'd need to add `xz-utils`

Because XZ is _relatively new_, changing the compression algorithm may cause problem on very old system. Please comment to issue if you have any concern about this.

### DIFF
--- a/plugins/python-build/share/python-build/2.7.10
+++ b/plugins/python-build/share/python-build/2.7.10
@@ -1,4 +1,4 @@
 #require_gcc
 install_package "openssl-1.0.2g" "https://www.openssl.org/source/openssl-1.0.2g.tar.gz#b784b1b3907ce39abf4098702dade6365522a253ad1552e267a9a0e89594aa33" mac_openssl --if has_broken_mac_openssl
 install_package "readline-6.3" "http://ftpmirror.gnu.org/readline/readline-6.3.tar.gz#56ba6071b9462f980c5a72ab0023893b65ba6debb4eeb475d7a563dc65cafd43" standard --if has_broken_mac_readline
-install_package "Python-2.7.10" "https://www.python.org/ftp/python/2.7.10/Python-2.7.10.tgz#eda8ce6eec03e74991abb5384170e7c65fcd7522e409b8e83d7e6372add0f12a" ldflags_dirs standard verify_py27 ensurepip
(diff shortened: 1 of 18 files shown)


thomasjo:
Awesome! 🎉

tsahara:
Build fails on 2 BSD-derived systems.

 - On OpenBSD 5.9, "tar" does not understand "-J" and it does not have any option to decompress "xz" files.  But "xz" command can be installed by OpenBSD ports.
 - On NetBSD6, "tar" does not understand "-J" but it can decompress .xz archives with "--xz" option.  "xz" command is installed by default as "/usr/bin/xz".

yyuu:
@tsahara Umm. I just guessed that it should work fine even on BSD since it worked on OS X 10.11.5 (it must be similar to FreeBSD toolchain). I should have been more careful about diversity between BSD families 😞 

Actually I'm not familiar to those BSD families.... Could you check the following if you get a chance?

* Do OpenBSD and NetBSD (and FreeBSD, and ...DragonFlyBSD?) have GNU tar package?
* Do these `gtar` executable can understand `-J` option?

If nothing helps, it might be better to consider rolling back to `.tar.gz`....

yyuu:
The footprint of `.tar.xz` could be smaller than `.tar.gz` for most cases. Plus, I guess majority of pyenv users are on GNU/Linux or OS X where `tar` supports `-J` by default. Generally speaking, using `.tar.xz` could be just better as long as `tar` supports it.

It must be possible to implement that conditional switching of download sources (`tar.xz` vs. `tar.gz`) based on availability of `tar -J`. I'll try to implement it if I get a chance.

tsahara:
> Do OpenBSD and NetBSD (and FreeBSD, and ...DragonFlyBSD?) have GNU tar package?
> Do these gtar executable can understand -J option?

- OpenBSD (5.9) has `gtar` package and it understands `-J` option.  It is GNU tar 1.28, installed to `/usr/local/bin/gtar`.
- NetBSD (6.1.5) has `gtar` package and it understands `-J` option too.  It is GNU tar 1.28, installed to `/usr/pkg/bin/gtar`.
- FreeBSD(10.3)'s `tar` can extract `.tar.xz` archives with/without `-J` option (it is smart enough to recognize compression algorithms automatically so `-J` option is ignored when extracting archives).
- I don't have any DragonFlyBSD hosts :)

PR #663, opened 2016-07-25
Use `.tar.xz` archive only if `tar` supports `-J` (especially on *BSD)

yyuu:
`tar` on some BSD systems may not support `-J` (`xz`). As reported in #654, using `.tar.xz` causes some errors on tarball extraction on NetBSD/OpenBSD. (OTOH, it seems `tar` of FreeBSD/OS X supports `-J`)

Generally speaking `.tar.xz` archives are way smaller than `.tar.gz` and it'd be better to use it as long as the `tar` itself supports it. This workaround will let python-build to use only on the platform where `tar` can work with `-J`.

(cc: @tsahara)

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -473,6 +473,11 @@ download_tarball() {
   fi
 }
 
+has_tar_xz_support() {
+  local tarxz="$(mktemp "XXXXXXXX.tar.xz")"
+  tar Jcf "${tarxz}" /dev/null 1>/dev/null 2>&1
+}
+
 fetch_git() {
   local package_name="$1"
   local git_url="$2"
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -474,8 +474,7 @@ download_tarball() {
 }
 
 has_tar_xz_support() {
-  local tarxz="$(mktemp "XXXXXXXX.tar.xz")"
-  tar Jcf "${tarxz}" /dev/null 1>/dev/null 2>&1
+  tar Jc /dev/null 1>/dev/null 2>&1
 }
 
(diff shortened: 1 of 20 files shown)


yyuu:
@tsahara could you check if this works on NetBSD/OpenBSD?

plugins/python-build/bin/python-build
```diff
@@ -473,6 +473,11 @@ download_tarball() {
   fi
 }
 
+has_tar_xz_support() {
+  local tarxz="$(mktemp "XXXXXXXX.tar.xz")"
```
tsahara: It does not work on NetBSD/OpenBSD.  They assume "XXXXXXXX" is at the end of the template string.

NetBSD:
```
% mktemp XXXXXXXX.tar.xz
XXXXXXXX.tar.xz
```

OpenBSD:
```
% mktemp XXXXXXXX.tar.xz
mktemp: insufficient number of Xs in template `XXXXXXXX.tar.xz'
```

But we don't have to create a temporary file to check if `tar` command supports `-J` option.  With `-f -` option, `tar` writes an archive to stdout.  I tried ``tar Jcf - /dev/null 1>/dev/null 2>&1`` and it works on FreeBSD/NetBSD/OpenBSD/macOS/Ubuntu ($?=0 on FreeBSD/macOS/Ubuntu and $?=1 on NetBSD/OpenBSD).

yyuu: Ah, good catch. I'll fix the code to not to use temporary file.

tsahara:
Thanks for the quick fix!  Build succeeded on both NetBSD and OpenBSD (I built python 3.5.2 on NetBSD and 2.7.12 on OpenBSD) at ad9805a.

PR #669, opened 2016-08-15
rbenv 20160815

yyuu:
`git merge rbenv/master -s recursive -X rename-threshold=5%`


yyuu:
It should be safe to merge since most of changes are mine :octocat:

PR #674, opened 2016-08-16
Add default-packages to .gitignore

jossy:
The plugin pyenv-default-packages uses `$(pyenv root)/default-packages` [as configuration file](https://github.com/jawshooah/pyenv-default-packages#usage). 
Since this plugin is listed as approved, I assume it makes sense to have the file permanently ignored by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /src/Makefile
 /src/*.o
 /bats/
+/default-packages


AlJohri:
this seems like a good idea 👍 .

@yyuu any chance `pyenv-default-packages` can become an officially endorsed package mirroring `rbenv-default-gems`? I just tried the latest version on [https://github.com/jawshooah/pyenv-default-packages](git) and it works fine. just needs update git tags and *ideally* to be uploaded as a proper package to homebrew. would be happy to maintain or make any PRs necessary!

yyuu:
Generally speaking, adding something to .gitignore doesn't mean _endorsment_. Anyway this is harmless and will merge.

AlJohri:
@yyuu makes sense. should I make another issue regarding pyenv-default-packages potentially coming under the `pyenv` organization? it seems to need very little maintenance. same code has been working since 2016.

PR #680, opened 2016-08-23
make all http source urls https

hawkowl:
Even if pyenv checks a hash for most things (which it doesn't for jython's hg), it doesn't say that to the user (which just tells them it's downloading a HTTP url then unzips and compiles it), and the cost of HTTPS nowadays is so nil that pyenv might as well use it for all the URLs.


hawkowl:
(related; https://github.com/yyuu/pyenv/issues/679)

joshfriend:
It's worth noting that ftpmirror.gnu.org redirects you to a mirror that does not use HTTPS even if you access ftpmirror.gnu.org using HTTPS.

Are you sure you covered everything? I scanned the file list quickly and noticed a few things like stackless and pypy were missing.

hawkowl:
@joshfriend yeah, that's suboptimal, re the GNU thing. From a security perspective we're still relying on the hash to be checked, but if GNU one day fixes their mirroring then it'll work just fine going forward :)

re stackless and pypy, heh my sed script crashed out at "patches" and didn't keep going :) It does appear that pypy's buildbot doesn't support HTTPS though  -- I've filed https://bitbucket.org/pypy/pypy/issues/2376/buildbotpypyorg-is-not-available-over about that, and just fixed the stackless ones.

yyuu:
We'd be happy if you send us another PR once PyPy's buildbot supports HTTPS :octocat:

hawkowl:
Thanks @yyuu! :)

PR #687, opened 2016-09-05
Only mv bin to bin.orig the first time, otherwise the mv will fail

s1341:
build_package_symlink_version_suffix is called twice. The second time, the mv will fail (as bin has already been moved to bin.orig), resulting in the build failing.

This patch performs the mv only the first time.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1470,7 +1470,9 @@ build_package_symlink_version_suffix() {
   if [[ "$PYTHON_CONFIGURE_OPTS" == *"--enable-framework"* ]]; then
     if [ -e "${PREFIX_PATH}/bin" ]; then
       # Always create `bin` as symlink to framework path if the version was built with `--enable-frameowrk` (#590)
-      mv -f "${PREFIX_PATH}/bin" "${PREFIX_PATH}/bin.orig"
+      if [ ! -e "${PREFIX_PATH}/bin.orig" ]; then
+        mv -f "${PREFIX_PATH}/bin" "${PREFIX_PATH}/bin.orig"
+      fi
     fi
     # Only symlinks are installed in ${PREFIX_PATH}/bin
     ln -fs "${PREFIX_PATH}/Python.framework/Versions/Current/bin" "${PREFIX_PATH}/bin"


plugins/python-build/bin/python-build
```diff
@@ -1470,7 +1470,9 @@ build_package_symlink_version_suffix() {
   if [[ "$PYTHON_CONFIGURE_OPTS" == *"--enable-framework"* ]]; then
     if [ -e "${PREFIX_PATH}/bin" ]; then
       # Always create `bin` as symlink to framework path if the version was built with `--enable-frameowrk` (#590)
-      mv -f "${PREFIX_PATH}/bin" "${PREFIX_PATH}/bin.orig"
+      if [ ! -e "${PREFIX_PATH}/bin.orig" ]; then
```
yyuu: I guess what you need to test here is `[ -e "${PREFIX_PATH}/bin" ]`. `mv -f` won't fail even if there's existing `${PREFIX_PATH}/bin.orig`.

s1341: I can try that.

s1341: That doesn't work. We need to check for bin.orig, otherwise the mv fails.

yyuu: I overlooked that there `[ -e "${PREFIX_PATH}/bin" ]` exists at L1471.... Obviously we don't need to duplicate the line.

s1341:
Try on OSX.

yyuu:
Because this is a feature related to framework installation, obviously I'm using OS X.

```sh
%  sw_vers
ProductName:    Mac OS X
ProductVersion: 10.11.6
BuildVersion:   15G31
```

Again, I couldn't reproduce the issue. I created dummy definition of `issue68` with the following content. Obviously it's running `build_package_symlink_version_suffix` 3 times.

```sh
echo $PREFIX_PATH

mkdir -p "${PREFIX_PATH}/bin" "${PREFIX_PATH}/lib"

cat <<EOS > "${PREFIX_PATH}/bin/python3.5"
#!/usr/bin/env bash
echo "$0 $@"
EOS
chmod a+x "${PREFIX_PATH}/bin/python3.5"

build_package_symlink_version_suffix
build_package_symlink_version_suffix
build_package_symlink_version_suffix

find "${PREFIX_PATH}"
```

Then, I ran it with using `python-build` and it just succeeded.

```sh
% PYTHON_CONFIGURE_OPTS="--enable-framework" ~/.pyenv/plugins/python-build/bin/python-build -v issue687 /tmp/issue68
/tmp/issue687
/tmp/issue687
/tmp/issue687/bin
/tmp/issue687/bin.orig
/tmp/issue687/bin.orig/python3.5
/tmp/issue687/include
/tmp/issue687/lib
/tmp/issue687/Python.framework
/tmp/issue687/Python.framework/Versions
/tmp/issue687/Python.framework/Versions/Current
/tmp/issue687/Python.framework/Versions/issue687
/tmp/issue687/Python.framework/Versions/issue687/include
/tmp/issue687/Python.framework/Versions/issue687/lib
/tmp/issue687/Python.framework/Versions/issue687/share
/tmp/issue687/share
```

Is there anyone who can reproduce the problem?

0xd3d0:
@yyuu your test works, because in the second time that build_package_symlink_version_suffix the bin directory doesn't exist, so the mv operation doesn't occur.

With a real target installation, like the following command, this is not the case:
```
PYTHON_BUILD_CACHE_PATH=/Users/blah/.sas/pyenv/buildcache PYTHON_CONFIGURE_OPTS="--enable-framework --enable-unicode=ucs2" pyenv install 2.7.8 -vfk
```

Which outputs:
```
mv: /Users/blah/.sas/pyenv/versions/2.7.8/bin and /Users/blah/.sas/pyenv/versions/2.7.8/bin.orig/bin are identical

BUILD FAILED (OS X 10.11.6 using python-build 20160726-7-g740fbe4)
```
This occurs after the pip installation succeeds...

s1341:
@yyuu any idea how to proceed here?

yyuu:
@s1341 sorry for delay. @0xd3d0's information helped me to identify the issue. I'll prepare some workaround for this maybe by the end of today in Japan time.

yyuu:
@s1341 @0xd3d0 I've merged #700 as a workaround for `mv` issue. I confirmed it worked with 2.7.8 on OS X 10.11.6 at least on my environment. Please let me know if you guys still see similar issues.

s1341:
The patch seems to work. Thanks!

Issue #607, opened 2016-05-23
clang: Not a directory


gatspy:
use pyenv shell
install pyqt5 and sip ,when execute make command prompt the following error message.
/usr/local/Cellar/pyenv/20160509/libexec/pyenv: line 43: cd: clang: Not a directory

joshfriend:
#502, #580, #588, 

Please remember to search the issues list first :)

Issue #588, opened 2016-05-03
False-positive shim error


brbsix:
I've got a Python [script](https://github.com/brbsix/pygmentize-addons/blob/master/pygshit) that's basically `less` and `cat` plus syntax highlighting. I was using it to view python's bash-completion script _/usr/share/bash-completion/completions/python2_ and encountered the following error:

```
/home/user/.pyenv/libexec/pyenv: line 43: cd: python: Not a directory
```

The bug is reproducible with the following command:

```
python - '/usr/share/bash-completion/completions/python2' <<EOF
pass
EOF
```

For some reason `pyenv` seems to think the file is something that it's not. The error message seems to come from the following function:

```
abs_dirname() {
  local cwd="$PWD"
  local path="$1"

  while [ -n "$path" ]; do
    cd "${path%/*}"
    local name="${path##*/}"
    path="$(resolve_link "$name" || true)"
  done

  pwd
  cd "$cwd"
}
```

Which was called by:

```
# Transfer PYENV_FILE_ARG (from shims) into PYENV_DIR.
if [ -z "${PYENV_DIR}" ]; then
  if [ -n "${PYENV_FILE_ARG}" ]; then
    if [ -L "${PYENV_FILE_ARG}" ]; then
      PYENV_DIR="$(abs_dirname "${PYENV_FILE_ARG}")"
    else
      PYENV_DIR="${PYENV_FILE_ARG%/*}"
    fi
    export PYENV_DIR
    unset PYENV_FILE_ARG
  fi
fi
```

I'm also able to reproduce the error by creating a link named `python` (or anything matching `python*` , the pattern found in `pyenv-rehash`'s `create_prototype_shim` function) and calling it. e.g.

```
touch somefile
ln -s somefile python
python /absolute/path/to/somefile
```

I realize this is not a particular common scenario, but is there any way to fix this?
##

yyuu:
I guess this the message was printed due to wrongly configured `$PATH`, as same reason as #580. Please ensure `$(pyenv root)/shims` is located before some system path (e.g. `/usr/bin`) in your `$PATH`.

brbsix:
I just installed pyenv in a clean Ubuntu 16.04 Vagrant box and am experiencing the error there as well.

This is the tail of *~/.bashrc*:

    PATH="$HOME/.pyenv/bin:$PATH"

    eval "$(pyenv init -)"
    eval "$(pyenv virtualenv-init -)"

This is `$PATH`:

[...]

Issue #580, opened 2016-04-21
pyenv and ansible-playbook: Not a directory error


YAmikep:
I have ansible and pyenv installed.

When using the `ansible-playbook` command, I get this error message:

```
vagrant@dev:~$ ansible-playbook
/home/vagrant/pyenv/libexec/pyenv: line 43: cd: ansible: Not a directory
```

Any idea why? 
##

yyuu:
Please show me the output of 'PYENV_DEBUG=1 ansible-playbook'. I guess you have stale shim script.

YAmikep:
@yyuu here is the output:
[pyenv_debug.txt](https://github.com/yyuu/pyenv/files/230935/pyenv_debug.txt)

yyuu:
@YAmikep Sorry for delay....

At the beginning of the debug print, I found `PYENV_FILE_ARG=/usr/bin/ansible-playbook` in the script. It must mean you have `ansible-playbook` executables in both `/usr/bin` and your `/home/vagrant/pyenv` and the former has precedence in your `${PATH}` configuration.

Basically I'd recommend you to have pyenv's shims path at almost top of your `${PATH}`. Giving precedence to system paths (e.g. `/usr/bin`, `/bin`) might cause problem like yours. Please review your `${PATH}` and try again.

```
+ program=python
+ [[ python = \p\y\t\h\o\n* ]]
+ for arg in '"$@"'
+ case "$arg" in
+ '[' -f /usr/bin/ansible-playbook ']'
+ export PYENV_FILE_ARG=/usr/bin/ansible-playbook
+ PYENV_FILE_ARG=/usr/bin/ansible-playbook
+ break
+ export PYENV_ROOT=/home/vagrant/pyenv
+ PYENV_ROOT=/home/vagrant/pyenv
+ exec /home/vagrant/pyenv/libexec/pyenv exec python /usr/bin/ansible-playbook
+ [pyenv:23] enable -f /home/vagrant/pyenv/libexec/../libexec/pyenv-realpath.dylib realpath
+ [pyenv:29] '[' -z '' ']'
++ [pyenv:31] type -p greadlink readlink
++ [pyenv:31] head -1
+ [pyenv:31] READLINK=/bin/readlink
+ [pyenv:32] '[' -n /bin/readlink ']'
+ [pyenv:53] '[' -z /home/vagrant/pyenv ']'
+ [pyenv:56] PYENV_ROOT=/home/vagrant/pyenv
[...]

PR #689, opened 2016-09-06
Resolve symlinks when symlink points to file in same directory

mdippery:
I encountered a bug in `abs_dirname` when using pyenv. I had a file structure that looked like this:

    /dept/is/dev/mdippery/opt/bin/asciidoc -> ../Cellar/asciidoc/bin/asciidoc
    /dept/is/dev/mdippery/opt/Cellar/asciidoc/bin/asciidoc -> asciidoc.py
    /dept/is/dev/mdippery/opt/Cellar/asciidoc/bin/asciidoc.py

pyenv's `abs_dirname` was encountering an error when it tried to resolve this system of symlinks. When `abs_dirname /dept/is/dev/mdippery/opt/bin/asciidoc` called, the following steps would happen:

1. `$path` was set to `/dept/is/dev/mdippery/opt/bin/asciidoc`.
2. Because `[ -n $path ]` was true, pyenv would enter the for loop on Line 42 of `~/.pyenv/libexec/pyenv`.
3. `$parent` would be set to `/dept/is/dev/mdippery/opt/bin`.
4. pyenv would `cd` to `$parent`.
5. `resolve_link asciidoc` would be called, and return `../Cellar/asciidoc/bin/asciidoc`, which would be set to `$path`.
6. The for loop would start again, since `[ -n $path ]` is still true.
7. `$parent` would be set to `../Cellar/asciidoc/bin`.
8. pyenv would `cd` to `$parent`.
9. `resolve_link asciidoc` would be called, and return `asciidoc.py`, which would be set to `$path`.
10. The for loop would start again, since `[ -n $path ]` is still true.
11. _Here's the bug:_ `$parent` would be set to `asciidoc.py`, since that is what `${path%/*}` returns when `$path` is simply `asciidoc.py`, with no parent information.
12. pyenv would naïvely assume that is a directory, and attempt to `cd` to it. _This causes an error to be thrown._

The simplest solution seems to be to check whether the "basename" of a path is the same as its "parent" (i.e., there is a symlink that points to another file in the same directory), and break from the loop if that is the case.

This fixes a bug that is referenced by #580, #588, and #607. A fix was also supplied in #502, but was never merged. (This fix is a bit different and may work better.)


mdippery:
Almost exactly the same bug is described in [rbenv#868](https://github.com/rbenv/rbenv/pull/868) as well.

yyuu:
Could you update the changes as exactly same as rbenv#868 to avoid conflict?

mdippery:
I'm not sure that rbenv#868 actually works properly, since it won't actually break out of the loop.

mdippery:
Actually, I take that back. I can update it so it looks the same as rbenv#868, except that rbenv#868 was never merged into rbenv, either, so there's not actually a conflict.

mikegleasonjr:
I have this wrong output also when running `ansible` from source. `ansible-playbook` is an alias of `ansible` in the same directory.

blueyed:
Yeah, came across my old PR (https://github.com/rbenv/rbenv/pull/868) for this again.
Seems to be copied code from bats after all: https://github.com/sstephenson/bats/pull/224.
There are even more places where this is used as a copy in pyenv.. :/

libexec/pyenv
```diff
@@ -40,7 +40,9 @@ else
     local path="$1"
 
     while [ -n "$path" ]; do
-      cd "${path%/*}"
+      local parent="${path%/*}"
+      if [ "$path" = "$parent" ]; then break; fi
+      cd "$parent"
```
blueyed: Breaking is wrong here: the local file might be a symlink itself that needs to get resolved.

blueyed:
A PR with a test is ready in rbenv itself now: https://github.com/rbenv/rbenv/pull/868#issuecomment-310792150.

blueyed:
btw: a workaround, which also improves performance is to run `make` in pyenv's `src` dir, which will generate `libexec/pyenv-realpath.dylib` to skip this method altogether.

jonganc:
How about simply changing `cd "${path%/*}"` to `[[ "$path" = */* ]] && cd "${path%/*}` (i.e. only change directory if `path` is not in the current directory)?

blueyed:
Yeah, I have to pick that up again.
Maybe somebody can help over there?

jonganc:
If I understood the rbenv discussion, mislav didn't think this was a relevant use case. On the pyenv side at least, I should point out that the bug shows up in the last few releases of ansible (i.e. note that in the [bin directory](https://github.com/ansible/ansible/tree/v2.3.2.0-1/bin), most executables are symlinks to `ansible`), so it's not some obscure issue.

PR #695, opened 2016-09-09
Added pypy2-5.4.1-src

Samureus:
Tested on Ubuntu 16.04 up to the point where it starts translating.


yyuu:
@Samureus just in case, could you add an _alias_ definition named `pypy-5.4.1-src` like this?

https://github.com/yyuu/pyenv/blob/45173d1f67b91f6623f9d29660dede5d590b42ec/plugins/python-build/share/python-build/pypy-5.4-src

Samureus:
Meh forgot about that one, added file through Github.

PR #696, opened 2016-09-09
Added pypy-portable-5.4.1 now

Samureus:
Tested on Ubuntu 16.04 64-bit.


yyuu:
Thanks for swift action!

PR #714, opened 2016-09-24
Added stackless-2.7.9

Samureus:
Tested on Ubuntu 16.04 64-bit. Please note that there's are also MacOS X dmg packages listed at

https://bitbucket.org/stackless-dev/stackless/downloads?tab=downloads

but I don't have a Mac to test with and I don't know much Travis-fu to confirm it will catch any mistakes.


yyuu:
In python-build, we're building stackless from source. You don't need to care about dmg.

PR #715, opened 2016-09-24
Stackless 3.4.2

Samureus:
Tested on Ubuntu 16.04 64-bit up to the point where it installs.

By the way, I did not include a checksum as I could not find any at the repository, but since Bitbucket uses HTTPS anyway that should be ok?


yyuu:
Could you add SHA256 checksum based on archive you downloaded from bitbucket? Idealy it'd be better to use checksum provided by origin, but if there none exists, sometime I'm generating checksome from downloaded archive by myself. It'd be just better than nothing, though.

Samureus:
Added the checksum. Will update the other 2 PRs.

yyuu:
@Samureus From the diff, it looks like you're trying to replace `stackless-2.7.9` by `stackless-3.4.2`. I guess this isn't what you intended to do. Could you review your changes again?

Samureus:
Sorry for the delay, just downloaded https://bitbucket.org/stackless-dev/stackless/get/v3.4.2-slp.tar.bz2 and ran it again:

```
$ sha256sum stackless-dev-stackless-587417070fe4.tar.bz2 
c3dbdde536fa6bf366fdbc3e957596487608ba11009758ce599357512eb74103  stackless-dev-stackless-587417070fe4.tar.bz2
```
But now git got picky. Let me close and re-send.

PR #718, opened 2016-09-25
Add Pyston 0.5.1

Samureus:
Tested on Ubuntu 16.04 up to the point where the Numpy tutorial at https://docs.scipy.org/doc/numpy-dev/user/quickstart.html works:

```python
$ python
Pyston v0.5.1 (rev abb4b132d260e5858a458462f71fd94955647195), targeting Python 2.7.7
>> import numpy
>> import numpy as np
>> a = np.arange(15).reshape(3, 5)
>> a
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14]])
>> a.shape
(3, 5)
>> a.ndim
2
>> a.dtype.name
'int64'
>> a.itemsize
8
>> type(a)
<type 'numpy.ndarray'>
>> b = np.array([6, 7, 8])
>> b
array([6, 7, 8])
>> type(b)
<type 'numpy.ndarray'>
>> 
```
This ended up a whole more hackish than I'd like, comments appreciated.

### DIFF
--- a/plugins/python-build/share/python-build/pyston-0.5.1
+++ b/plugins/python-build/share/python-build/pyston-0.5.1
@@ -5,8 +5,15 @@ case "$(pyston_architecture 2>/dev/null || true)" in
   # pyston targets python 2.7.7 and does not have ensurepip, when attempting to
   # run https://bootstrap.pypa.io/get-pip.py it dumps core, see
   # https://github.com/dropbox/pyston/issues/1373
-  # FIXME: more ugly hacks because Olenna Tyrell would approve
+  # FIXME ugly hack below adapted from
+  # https://github.com/dropbox/pyston/blob/master/docker/pyston/Dockerfile
+  ( cd "${PREFIX_PATH}/" && ./pyston virtualenv/virtualenv.py . )
+  # activate the virtualenv so that we're using the correct pip and site-packages location
+  ( source "${PREFIX_PATH}/bin/activate" )
+  # FIXME: if this fails it should not break the installation
   ( cd "${PREFIX_PATH}/bin" && ./pip install https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz && ./pip install git+git://github.com/numpy/numpy@v1.11.0 )
+  # deactivate the virtualenv
+  ( deactivate )
   ;;
 * )
   { echo
--- a/plugins/python-build/share/python-build/pyston-0.5.1
+++ b/plugins/python-build/share/python-build/pyston-0.5.1
@@ -11,7 +11,7 @@ case "$(pyston_architecture 2>/dev/null || true)" in
 * )
   { echo
     colorize 1 "ERROR"
-    echo ": A Pyston 0.5.1 binary is not available for $(pypy_architecture 2>/dev/null || true)."
+    echo ": A Pyston 0.5.1 binary is not available for $(pyston_architecture 2>/dev/null || true)."
     echo
   } >&2
   exit 1
--- a/plugins/python-build/share/python-build/pyston-0.5.1
+++ b/plugins/python-build/share/python-build/pyston-0.5.1
@@ -1,0 +1,19 @@
+case "$(pypy_architecture 2>/dev/null || true)" in
+"linux64" )
+  install_package "pyston-0.5.1-linux64" "https://github.com/dropbox/pyston/releases/download/v0.5.1/pyston-0.5.1-linux64.tar.gz#4b0d2ad2c19f6393b79adbb9312649d38cb8cb3daf0c8fdf8631465c7761bb79" "pyston" verify_py27
+  # disabling ensurepip avoids having installation aborted because of a coredump
+  # pyston targets python 2.7.7 and does not have ensurepip, when attempting to
+  # run https://bootstrap.pypa.io/get-pip.py it dumps core, see
+  # https://github.com/dropbox/pyston/issues/1373
+  # FIXME: more ugly hacks because Olenna Tyrell would approve
+  ( cd "${PREFIX_PATH}/bin" && ./pip install https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz && ./pip install git+git://github.com/numpy/numpy@v1.11.0 )
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
+    echo ": A Pyston 0.5.1 binary is not available for $(pypy_architecture 2>/dev/null || true)."
+    echo
+  } >&2
+  exit 1
+  ;;
+esac
--- a/plugins/python-build/share/python-build/pyston-0.5.1
+++ b/plugins/python-build/share/python-build/pyston-0.5.1
@@ -6,11 +6,11 @@ case "$(pyston_architecture 2>/dev/null || true)" in
   # run https://bootstrap.pypa.io/get-pip.py it dumps core, see
   # https://github.com/dropbox/pyston/issues/1373
   # activate the virtualenv so that we're using the correct pip and site-packages location
-  ( source "${PREFIX_PATH}/bin/activate" )
+  # ( source "${PREFIX_PATH}/bin/activate" )
   # FIXME: if this fails it should not break the installation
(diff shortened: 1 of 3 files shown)


yyuu:
@Samureus Great addition of new kind of python! I feel `numpy` must be an optional package which should be managed by users of pyenv, though. I guess we should remove `numpy` part from this. How do you think?

Samureus:
I agree that adding numpy increases the build time but:

- From what I could gather from the README, pyston 0.5.1 only works with specific versions of virtualenv (12.1.dev0), a custom version of cython (0.24) and numpy (1.11.0).
- Which is why I chose to have it all built-in, so that users can start using it "batteries included".
- It also has the advantage that since the numpy module is installed inside `.pyenv/versions/pyston-0.5.1/site-packages/numpy/`, if I understand virtualenv correctly, any virtualenv created will copy the right numpy installation.

Just to test a silly idea, I went ahead and created a virtualenv using the downloaded 0.5.1 file, and this is what happened:

```
(pyston-test) samo@Bruinen:~$ pip install --user https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz
Collecting https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz
  Downloading https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz (1.7MB)
    100% |████████████████████████████████| 1.7MB 131kB/s 
Building wheels for collected packages: Cython
  Running setup.py bdist_wheel for Cython ... done
  Stored in directory: /home/samo/.cache/pip/wheels/f7/48/1e/98aadc70b64899c309ec1dd31540c8551d7ab38db8bf4eba42
Successfully built Cython
Installing collected packages: Cython
Successfully installed Cython-0.24
(pyston-test) samo@Bruinen:~$ pip install --user git+git://github.com/numpy/numpy@v1.11.0
Collecting git+git://github.com/numpy/numpy@v1.11.0
  Cloning git://github.com/numpy/numpy (to v1.11.0) to /tmp/pip-B9r9aI-build
Installing collected packages: numpy
  Found existing installation: numpy 1.11.1
    Uninstalling numpy-1.11.1:
      Successfully uninstalled numpy-1.11.1
  Running setup.py install for numpy ... done
Successfully installed numpy-1.11.0
(pyston-test) samo@Bruinen:~$ python
Pyston v0.5.1 (rev abb4b132d260e5858a458462f71fd94955647195), targeting Python 2.7.7
>> import numpy as np
  File "<stdin>", line 1, in <module>
ImportError: No module named numpy
>> 
```
Note that I had to use pip install --user and even after that it didn't work.

What I can think of right now:

- I can split it into `pyston-0.5.1` and `pyston-0.5.1-numpy`, with the later installing numpy into pyston-0.5.1. Maybe I can put a message to point users to use `pyston-0.5.1-numpy` if they want to save themselves the hassle.
- We can also wait until the Pyston guys create another point release which should resolve the core dump, which prevents installing the pip bootstrap.

yyuu:
@Samureus Gotcha, I've got your point. I think it's acceptable to include `numpy` as a part of `pyston` installation at present since it's still in its early stage.

plugins/python-build/share/python-build/pyston-0.5.1-numpy
```diff
@@ -0,0 +1 @@
+source "${BASH_SOURCE%/*}/pyston-0.5.1"
```
yyuu: I don't think we need separate definitions of `pyston-0.5.1` and `pyston-0.5.1-numpy`. Just make `pyston-0.5.1` as "battery included" at least for now.

kmod:
This looks cool!  FYI we improved our virtualenv compatibility post-0.5.1, so now we don't need a custom virtualenv script anymore.  We also changed some of our output directories to match CPython (ie we now use `lib/python2.7/site-packages` instead of just `site-packages`).  And the new version of Cython should have Pyston support, and hopefully the next version of numpy will use the new version of Cython, which was the reason we couldn't use the default numpy.

So hopefully soon using Pyston will become much easier :)

plugins/python-build/share/python-build/pyston-0.5.1-numpy
Samureus: Done in https://github.com/yyuu/pyenv/pull/718/commits/8acb4ebaba1010112fd6b48c3c2311894b98ffc9

plugins/python-build/share/python-build/pyston-0.5.1
```diff
@@ -0,0 +1,19 @@
+case "$(pypy_architecture 2>/dev/null || true)" in
```
Samureus: Done in https://github.com/yyuu/pyenv/pull/718/commits/d24ce2494bc4d2890368b9f15c9d5531e61dcb31 and https://github.com/yyuu/pyenv/pull/718/commits/4e4bd420202bcc16da696002171fafd007dae28f

plugins/python-build/share/python-build/pyston-0.5.1
```diff
@@ -0,0 +1,19 @@
...
+  # pyston targets python 2.7.7 and does not have ensurepip, when attempting to
+  # run https://bootstrap.pypa.io/get-pip.py it dumps core, see
+  # https://github.com/dropbox/pyston/issues/1373
+  # FIXME: more ugly hacks because Olenna Tyrell would approve
+  ( cd "${PREFIX_PATH}/bin" && ./pip install https://github.com/dropbox/pyston/releases/download/v0.5.1/Cython-0.24-pyston.tar.gz && ./pip install git+git://github.com/numpy/numpy@v1.11.0 )
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
+    echo ": A Pyston 0.5.1 binary is not available for $(pypy_architecture 2>/dev/null || true)."
```
yyuu: @Samureus `s/pypy_architecture/pyston_architecture/`

Samureus: Oops, done in https://github.com/yyuu/pyenv/pull/718/commits/7a957c54889da314a50ae8e970246f80e5386e60

Samureus:
Aaaand I broke the numpy build! It looks like installing gfortran and libatlas-base-dev since yesterday borked the setup script, I uninstalled them and now it complains about

`
    creating build/temp.linux-x86_64-2.7/numpy/random/mtrand
    
    compile options: '-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE=1 -D_LARGEFILE64_SOURCE=1 -Inumpy/core/include -Ibuild/src.linux-x86_64-2.7/numpy/core/include/numpy -Inumpy/core/src/private -Inumpy/core/src -Inumpy/core -Inumpy/core/src/npymath -Inumpy/core/src/multiarray -Inumpy/core/src/umath -Inumpy/core/src/npysort -I/home/samo/.pyenv/versions/pyston-0.5.1/from_cpython/Include -Ibuild/src.linux-x86_64-2.7/numpy/core/src/private -Ibuild/src.linux-x86_64-2.7/numpy/core/src/private -Ibuild/src.linux-x86_64-2.7/numpy/core/src/private -c -Werror=implicit-function-declaration'
    
    ccache: numpy/random/mtrand/mtrand.c
    
    numpy/random/mtrand/mtrand.c:49:27: fatal error: longintrepr.h: No existe el archivo o el directorio
    
    compilation terminated.
    
    numpy/random/mtrand/mtrand.c:49:27: fatal error: longintrepr.h: No existe el archivo o el directorio
    
    compilation terminated.
`
I went back to https://github.com/dropbox/pyston/tree/master/from_cpython/Include and longintrepr.h is missing, see https://github.com/dropbox/pyston/issues/1376.

I'm also following the instructions in http://www.scipy.org/scipylib/building/linux.html#debian-ubuntu but between the missing header and wrapping my head about selecting the correct BLAS library, this is getting messy.

@yyuu and @kmod what do you think, should I drop bundling numpy alltogether, or just wait until the Pyston people have another release? I still believe an early release is good for letting people test.

Samureus:
I think I have an idea what's going on, apparently it's trying to use cython 0.24.1 which I installed on my machine, instead of the one I specified, I need to test this later.

Samureus:
@yyuu Ok I give up on bundling numpy, at least it installs pyston for now. If you like, I can add a warning message about pip and numpy or redirect them to just use the docker image with numpy.

yyuu:
@Samureus thanks for your effort. I think we don't need warnings at least for now since every users of Pyston would know it's still in earlier stage of development.

PR #725, opened 2016-10-03
Add patch to fix 3.5.2 bug - https://bugs.python.org/issue26664

robsonpeixoto:


yyuu:
Thanks for your help :+1:

PR #726, opened 2016-10-05
Stop using ucs4 as default unicode encoding on OS X

yyuu:
This PR is follow up of discussion in https://github.com/yyuu/pyenv/issues/257

This lets python-build to stop configuring `--enable-unicode=ucs4`. It means that CPython built on OS X will use narrow encoding (UCS2) after this.

At least for now, this doesn't care platforms other than OS X; all BSDs will be managed just as similar as Linux distros in terms of encoding perspective. This may cause some binary incompatibility issue on some BSD families. I may be going to tweak the lines again if I got some issues from them.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1934,7 +1934,9 @@ fi
 
 # Compile with `--enable-unicode=ucs4` by default (#257)
 if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
-  package_option python configure --enable-unicode=ucs4
+  if ! is_mac; then
+    package_option python configure --enable-unicode=ucs4
+  fi
 fi
 
 # SSL Certificate error with older wget that does not support Server Name Indication (#60)
--- a/plugins/python-build/test/build.bats
+++ b/plugins/python-build/test/build.bats
@@ -67,6 +67,10 @@ assert_build_log() {
   stub_make_install
   stub_make_install
 
+  # yyuu/pyenv#257
+  stub uname '-s : echo Linux'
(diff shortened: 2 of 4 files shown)


matthew-brett:
Tested on my laptop - looks good to me - thanks.

cancan101:
When I reinstalled python, I get:
```
  File "/Users/alex/.pyenv/versions/2.7.12/bin/virtualenv", line 7, in <module>
    from virtualenv import main
  File "/Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/site-packages/virtualenv.py", line 16, in <module>
    import base64
  File "/Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/base64.py", line 9, in <module>
    import struct
  File "/Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/struct.py", line 1, in <module>
    from _struct import *
ImportError: dlopen(/Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_struct.so, 2): Symbol not found: _PyUnicodeUCS2_AsEncodedString
  Referenced from: /Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_struct.so
  Expected in: flat namespace
 in /Users/alex/.pyenv/versions/2.7.12/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_struct.so
```

cancan101:
Looks like a full uninstall and then reinstall fixed.

PR #731, opened 2016-10-11
Add CPython 3.6.0b2 and remove 3.6.0b1

zmwangx:
https://www.python.org/downloads/release/python-360b2/.


joshfriend:
Checksums look correct :+1:

PR #811, opened 2017-01-17
Add CPython 3.5.3

browniebroke:
https://www.python.org/downloads/release/python-353/


plugins/python-build/share/python-build/3.5.3
```diff
@@ -0,0 +1,8 @@
+#require_gcc
+install_package "openssl-1.0.2g" "https://www.openssl.org/source/openssl-1.0.2g.tar.gz#b784b1b3907ce39abf4098702dade6365522a253ad1552e267a9a0e89594aa33" mac_openssl --if has_broken_mac_openssl
+install_package "readline-6.3" "https://ftpmirror.gnu.org/readline/readline-6.3.tar.gz#56ba6071b9462f980c5a72ab0023893b65ba6debb4eeb475d7a563dc65cafd43" standard --if has_broken_mac_readline
+if has_tar_xz_support; then
+  install_package "Python-3.5.3" "https://www.python.org/ftp/python/3.5.3/Python-3.5.3.tar.xz#f33c27745f2bd87344be790465ef984a972fd539dc83bd4f61d4242c607ef1ee" ldflags_dirs standard verify_py35 ensurepip
```
yyuu: It seems SHA2 of both tar.gz and tar.xz are configured as same. I think at lease one checksum is wrong.

browniebroke: Indeed. I've copied the URLs that were on the Python page and got the archives with curl, and I realised they were actually stored under /ftp/python/3.5.2/ on the Python website, which is odd. Anyway, it fixed now, but it looks like they were both wrong.

andriisoldatenko:
+1

yyuu:
Thanks~

PR #818, opened 2017-01-22
simple init.zsh

hackaugusto:
Simple init.zsh that can be used by zgen/antigen to add pyenv to the PATH. This removes from the step 2 of the installation the need to manually change the path, eg.:

`echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile`

### DIFF
--- a/_pyenv
+++ b/_pyenv
@@ -1,0 +1,33 @@
+#compdef pyenv
+
+_pyenv() {
+  _arguments -s '*::pyenv command:_pyenv_command'
+}
+
+(( ${+functions[_pyenv_command]} )) || _pyenv_command() {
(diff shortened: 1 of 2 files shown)


yyuu:
I don't want to merge _platform_ specific changes into pyenv core to keep repository content solid. I guess this feature can be implemented as a separated repository (e.g. pyenv-zgen).

hackaugusto:
What about the completion function? `_pyenv` is the same functionality as [pyenv.zsh](https://github.com/yyuu/pyenv/blob/99d16707e372143fb35d822c26fe8427719b903c/completions/pyenv.zsh) using the shell's newer completion system.

PR #831, opened 2017-02-09
Fix fish subcommand completion

maxnordlund:
This allows subcommand style plugins to properly autocomplete.
Existing commands are not affected. 

Example, say you have support for `pyenv foo bar --flag`, then
this allows the last `--flag` argument to be properly completed.

### DIFF
--- a/completions/pyenv.fish
+++ b/completions/pyenv.fish
@@ -18,5 +18,6 @@ end
 
 complete -f -c pyenv -n '__fish_pyenv_needs_command' -a '(pyenv commands)'
 for cmd in (pyenv commands)
-  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a "(pyenv completions $cmd)"
+  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a \
+    "(pyenv completions (commandline -opc)[2..-1])"
 end


yyuu:
It seems completion function for zsh is doing similar thing. Good catch :+1:

https://github.com/yyuu/pyenv/blob/f36d735e27a7619da02a4bb45c9b4a4958b5a888/completions/pyenv.zsh#L14

completions/pyenv.fish
```diff
@@ -18,5 +18,6 @@ end
 
 complete -f -c pyenv -n '__fish_pyenv_needs_command' -a '(pyenv commands)'
 for cmd in (pyenv commands)
-  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a "(pyenv completions $cmd)"
+  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a \
+    "(pyenv completions (commandline -opc)[2..-1])"
```
yyuu: Wouldn't it be better to check length of `(commandline -opc)`like zsh completion's doing? In case of `(commandline -opc)` == 2, I guess it should return `(pyenv commands)` instead of `(pyenv completions (command -opc)[2..-1])`. Please see also zsh's implementation.

https://github.com/yyuu/pyenv/blob/f36d735e27a7619da02a4bb45c9b4a4958b5a888/completions/pyenv.zsh#L11-L15

maxnordlund: That is already handled in the `__fish_pyenv_needs_command`/`__fish_pyenv_using_command `, but if it was just one completion then yes.

maxnordlund:
> @maxnordlund could you open similar PR to https://github.com/rbenv/rbenv as well? Because pyenv's sharing most of code with that, contributing to upstream will be welcomed.

I'll do it right away 👍

completions/pyenv.fish
yyuu: Ah, understood. LGTM :+1:

PR #873, opened 2017-03-22
Add CPython 3.6.1 and remove 3.6.1rc1

zmwangx:
https://www.python.org/downloads/release/python-361/


joshfriend:
Awesome, thank you!

PR #883, opened 2017-03-27
Remove prototype shim if rehash is interrupted

sschaetz:
We use pyenv in our CI and the CI system can kill a job at any point in time (force push, new commit). I think we had a situation where the job was killed while pyenv was rehashing and the file was not removed. 

All subsequent builds failed with `pyenv: cannot rehash: [...].pyenv/shims/.pyenv-shim exists`

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -27,8 +27,9 @@ set -o noclobber
 set +o noclobber
 
 # If we were able to obtain a lock, register a trap to clean up the
-# prototype shim when the process exits.
-trap remove_prototype_shim EXIT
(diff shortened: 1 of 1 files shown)


sschaetz:
Any thoughts?

yyuu:
Sorry for delay. I didn't have enough time for investigation. Though IIRC trapping EXIT must be sufficient to hook up all exit cases except SIGKILL. I'm not sure if there's actually some cases which will not be covered by EXIT hook.

sschaetz:
I think you're right. The issue we saw was very likely due to SIGKILL.

PR #905, opened 2017-05-07
Updated README with correct path for zsh

vmesel:


blueyed:
That really depends, and makes it slower in general.

tukushan:
Installing into `~/.zshenv` didn't work for me but `~/.zshrc` did.

blueyed:
@tukushan 
Have you logged out and in again?

andrebautista:
Can confirm this change is necessary. It seems as though the .zshenv paths are being added after usr/bin and usr/local/bin. When I move them to the .zshrc file the pyenv paths are added before and the correct python version is used.

anton-petrov:
Already corrected.

PR #907, opened 2017-05-09
Fixes pip-rehash to rehash if pip was called with a flag

LarsFronius:
`pip -v install foobar` or `pip -q install foobar` did not trigger a rehash before. Now it should have the same behaviour as `pip install foobar`.


yyuu:
Good catch 👍

PR #921, opened 2017-05-29
micropython-dev: Set better sys.path

scop:
The default was `~/.micropython:/usr/lib/micropython`


scop:
Rebased; could you have a look at applying this? Thanks!

yyuu:
Confirmed this worked for me. Thanks!

PR #927, opened 2017-06-05
Merge rbenv master

blueyed:
As of rbenv/rbenv@755c820.

    Conflicts:
            README.md
            libexec/pyenv---version
            libexec/pyenv-init
            libexec/rbenv
            libexec/rbenv-sh-shell
            libexec/rbenv-which
            test/init.bats
            test/shell.bats

I've added the "rbenv*" files and renamed them, so that they should not conflict with "deleted by us" in the future.

### DIFF
--- a/libexec/pyenv---version
+++ b/libexec/pyenv---version
@@ -12,7 +12,7 @@
 set -e
 [ -n "$PYENV_DEBUG" ] && set -x
 
-version="1.0.10"
+version="1.1.0"
 git_revision=""
 
 if cd "${BASH_SOURCE%/*}" 2>/dev/null && git remote -v 2>/dev/null | grep -q pyenv; then
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -86,8 +86,8 @@ mkdir -p "${PYENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "setenv PATH '${PYENV_ROOT}/shims' \$PATH"
-  echo "setenv PYENV_SHELL $shell"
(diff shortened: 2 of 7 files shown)


libexec/pyenv-init
```diff
@@ -97,10 +97,7 @@ esac
 
 completion="${root}/completions/pyenv.${shell}"
 if [ -r "$completion" ]; then
-  case "$shell" in
-  fish ) echo ". '$completion'" ;;
-  *    ) echo "source '$completion'" ;;
-  esac
+  echo "source '$completion'"
```
blueyed: This seems odd, but comes from rbenv/rbenv@8b0b51a (https://github.com/rbenv/rbenv/pull/873).

alex: As of fish 2.6, both appear to work. I could have sworn with fish 2.5 that `source` was the preferred spelling.

blueyed: Yes, seems to be good - but it confused me a bit.
Probably `source` was not available / preferred in older fish versions?!

libexec/pyenv---version
```diff
@@ -12,7 +12,7 @@
 set -e
 [ -n "$PYENV_DEBUG" ] && set -x
 
-version="1.0.10"
+version="1.1.0"
```
yyuu: 👍

alex:
🎉  is there a plan to issue a release with this change it?

jackfischer:
@alex I don't know if you are using pyenv through homebrew but in trying to get this I just discovered you can run `brew upgrade pyenv --HEAD` for a quick fix.

PR #941, opened 2017-06-23
python-build: fix/improve is_mac function

blueyed:
shellcheck was not able to parse it before.

There are no tests, but I've checked it using:

    uname_s() {
      echo "Darwin"
    }
    osx_version() {
      echo 1000
    }
    is_mac() {
      [ "$(uname_s)" = "Darwin" ] || return 1
      (( $# == 0 )) || eval "[ '$(osx_version)' $* ]"
    }

    is_mac; echo $?
    is_mac -gt 1000; echo $?
    is_mac -lt 1000; echo $?
    is_mac -gt 999; echo $?

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -101,9 +101,12 @@ os_information() {
   fi
 }
 
+# Is this MacOS?
+# This can be called with arguments for test/[[ optionally, in which case it is
+# tested against the version from osx_version.
 is_mac() {
-  [ "$(uname -s)" = "Darwin" ] || return 1
-  [ $# -eq 0 ] || [ "$(osx_version)" "$@" ]
+  [[ "$(uname -s)" == 'Darwin' ]] || return 1
+  (( $# == 0 )) || eval "[[ '$(osx_version)' $* ]]"
 }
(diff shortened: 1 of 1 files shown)


blueyed:
For reference, the shellcheck issue:
```
:!shellcheck plugins/python-build/bin/python-build
                                                  

In plugins/python-build/bin/python-build line 104:
is_mac() {                                        
         ^-- SC1009: The mentioned parser error was in this brace group.


In plugins/python-build/bin/python-build line 106:                      
  [ $# -eq 0 ] || [ "$(osx_version)" "$@" ]       
                  ^-- SC1073: Couldn't parse this test expression.
                                      ^-- SC1072: Expected comparison operator (don't wrap commands in []/[[]]). Fix any mentioned problems and try again.

shell returned 1                                                                                                                                          
```

native-api:
What's the real issue here? `python-build` isn't checked with Shellcheck as part of CI AFAIK so it's not obliged to pass it.

The change could be accepted regardless if it makes the code better (cleaner, more readable) -- but this doesn't look like the case, either: the existing and proposed code are roughly the same in terms of readability.

PR #965, opened 2017-08-14
Added 3.4.7

jamesstidard:
First time adding a version so worth double checking


joshfriend:
Nice first contribution, Thanks a lot!

PR #970, opened 2017-08-19
Option to disable implicit whence command.

vadviktor:
Hey,

there is a shell prompt builder that I write I Go which also has a python plugin https://github.com/bullettrain-sh/bullettrain-go-python.

The problem I am facing is that whenever I am testing e.g. `python3 --version`  in a virtualenv or setup that has no python3, the whence command kicks in a search if it can help the user. In many cases it's great and helps a lot.

In my case, it is merely a performance bottleneck and useless in my scenario. Therefore I kindly ask if I can share this little option (which be also beneficial for other shell themes out there) to disable the implicit search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,8 +195,8 @@ easy to fork and contribute any changes back upstream.
     $ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
     $ echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
     ```
-    **Zsh note**: Modify your `~/.zshenv` file instead of `~/.bash_profile`.  
-    **Ubuntu and Fedora note**: Modify your `~/.bashrc` file instead of `~/.bash_profile`.  
+    **Zsh note**: Modify your `~/.zshenv` file instead of `~/.bash_profile`.
+    **Ubuntu and Fedora note**: Modify your `~/.bashrc` file instead of `~/.bash_profile`.
     **Proxy note**: If you use a proxy, export `http_proxy` and `HTTPS_PROXY` too.
 
 3. **Add `pyenv init` to your shell** to enable shims and autocompletion.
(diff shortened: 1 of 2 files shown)


blueyed:
What about doing this automatically based on if it is used interactively (`if [ -t 0 ]`)?

vadviktor:
@blueyed sorry but I am not too friendly with shell programming (that's why I write my stuff in other languages mostly) and I am not familiar at all with using `-t`. Could you point me to some pages where I can see it in action or having it explained in details, please? Otherwise I can't decide if it's the right tool or even, how to test it if it's right.

vadviktor:
Ok I have found a place http://tldp.org/LDP/abs/html/intandnonint.html#II2TEST
Testing it now.

vadviktor:
@blueyed So it does have the same expected output BUT not the expected performance. Using 
```
if [ -t 0 ]; then
    exit 127
fi
```

is as slow as not using it. On the other hand, testing for the env variable has both the expected output and more importantly, performance.

blueyed:
Please make sure it is hit.

It seems to be hit with ´]echo $(pyenv which foo </dev/null)`.

But it shows that you might want to use `if ! [[ -t 1 ]]` - i.e. if stdout is not connected to a TTY.

vadviktor:
Hey @blueyed,

thanks for your help, I am using 

```shell
if ! [[ -t 1 ]]; then
  exit 127
fi
```

as you suggested and it really gives me the performance and output I'd need. Though I don't understand how can this be what I am looking for, as now, issuing just `python3` on a python 2 only virtualenv won't suggest me virtualenvs where it exists. I think this method would kill this feature, as far as I can see it working locally.
Do you think there is more to be done to have it being suppressed only when I need it to be?

Thanks!

vadviktor:
Hey, I am still looking for a way not to kill the basic feature of whence, but be able to disable automatically being called when one is sure they don't need it. I'd really help to build snappier prompt builders. Can you get back to me on this matter, please?

PR #971, opened 2017-08-23
Added CPython 3.5.4

claco:


claco:
First contrib. Hopefully Travis double checks the checksums.

PR #977, opened 2017-09-07
Update `pyenv init` line in README.md

cia-rana:
I added ```[ -x ~/.pyenv/bin/pyenv ]``` to the front of ```pyenv init```  in case ```pyenv``` is not found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo '[ -x $PYENV_ROOT/bin/pyenv ] && eval "$(pyenv init -)"' >> ~/.bash_profile
+    $ echo 'eval "$([[ -x $PYENV_ROOT/bin/pyenv ]] && pyenv init -)"' >> ~/.bash_profile
     ```
     **Zsh note**: Modify your `~/.zshenv` file instead of `~/.bash_profile`.  
     **Ubuntu and Fedora note**: Modify your `~/.bashrc` file instead of `~/.bash_profile`.
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo '[[ -x $PYENV_ROOT/bin/pyenv ]] && eval "$($PYENV_ROOT/bin/pyenv init -)"' >> ~/.bash_profile
+    $ echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
     ```
     **Zsh note**: Modify your `~/.zshenv` file instead of `~/.bash_profile`.  
     **Ubuntu and Fedora note**: Modify your `~/.bashrc` file instead of `~/.bash_profile`.
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo 'eval "$(pyenv init -)"' >> ~/.bash_profile
+    $ echo '[ -x ~/.pyenv/bin/pyenv ] && eval "$(pyenv init -)"' >> ~/.bash_profile
```
blueyed: It might be somewhere else..
You might want to include `$PYENV_ROOT` in the check and/or check for `PYENV_ROOT` being non-empty and use it in the `eval`.

README.md
```diff
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo 'eval "$(pyenv init -)"' >> ~/.bash_profile
+    $ echo '[ -x $PYENV_ROOT/bin/pyenv ] && eval "$(pyenv init -)"' >> ~/.bash_profile
```
blueyed: As for Bash and other modern shells, using `[[` instead of `[` is better.
And then also use it in the `eval` - otherwise it might still fail if `$PYENV_ROOT` is not in the `$PATH`.

README.md
```diff
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo '[ -x $PYENV_ROOT/bin/pyenv ] && eval "$(pyenv init -)"' >> ~/.bash_profile
+    $ echo 'eval "$([[ -x $PYENV_ROOT/bin/pyenv ]] && pyenv init -)"' >> ~/.bash_profile
```
blueyed: More like `[[ -x $PYENV_ROOT/bin/pyenv ]] && eval "$($PYENV_ROOT/bin/pyenv init -)" I've meant.

blueyed:
(NOTE: please squash-merge this in case you agree with it)

README.md
```diff
@@ -203,7 +203,7 @@ easy to fork and contribute any changes back upstream.
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
    configuration file since it manipulates `PATH` during the initialization.
     ```sh
-    $ echo 'eval "$(pyenv init -)"' >> ~/.bash_profile
+    $ echo '[[ -x $PYENV_ROOT/bin/pyenv ]] && eval "$($PYENV_ROOT/bin/pyenv init -)"' >> ~/.bash_profile
```
yyuu: 1. The existence of `pyenv` executable doesn't mean we can skip `pyenv-init`. We can skip `pyenv-init` only if it has already source'd (=~ `pyenv` function exists in current shell)
1. `$PYENV_ROOT` may not be declared unless `pyenv-init` has source'd. It'd be better to use `pyenv root` since it must work as long as `$PATH` has set up

If you really want to avoid sourcing pyenv multiple times, what you really need to do must be follows. I'm not sure if it worth doing this like this though.

```sh
if ! declare -f pyenv 1>/dev/null 2>&1; then
  eval "$(pyenv init -)"
fi
```

blueyed: @yyuu 
It is not about avoiding sourcing it multiple times, but in case pyenv is not installed, e.g. when using the same dotfiles somewhere else (?!).

cia-rana: That’s exactly what I wanted to say! I created this pull request because an error about `pyenv` occurs when I use my dotfiles in an environment that `pyenv` isn't installed.

yyuu: If so, what you need to do before `pyenv init` is to check the availability of `pyenv` command. I think relying on `PYENV_ROOT` here could not be a good idea since not all people are setting it.

```sh
if command -v pyenv 1>/dev/null 2>&1; then
  eval "$(pyenv init -)"
fi
```

blueyed:
Woah, that looks a bit ugly..
And then it also has not been squash-merged?!

yyuu:
Ah, sorry. I mistakenly merged without squashing :disappointed:

PR #984, opened 2017-09-20
Add CPython 3.6.3rc1 and 3.7.0a1

zmwangx:
https://www.python.org/downloads/release/python-363rc1/
https://www.python.org/downloads/release/python-370a1/


joshfriend:
Thank you!

Issue #98, opened 2014-01-04
Fails for/with `virtualenv --python="$(which python)" FOO`


blueyed:
When using the above code to create a virtualenv, it fails:

```
❯ virtualenv --python="$(which python)" FOO
Running virtualenv with interpreter /home/user/.pyenv/shims/python
New python executable in FOO/bin/python
ERROR: The executable FOO/bin/python is not functioning
ERROR: It thinks sys.prefix is u'/usr' (should be u'/home/user/src/pip-tools/FOO')
ERROR: virtualenv is not compatible with this system or executable
```

Since `virtualenv` itself is shimmed, a solution to this might be to replace instances to itself accordingly.

For example, the following tests make use of it:

```
https://github.com/nvie/pip-tools/blob/master/tests/dump.t
```

tgalery:
hi guys, sorry to disturb you with this, but I'm pretty much having this issue when passing a miniconda python bin, e.g.

```
$ virtualenv env2 -p /home/ec2-user/.pyenv/versions/miniconda-3.10.1/bin/python --system-site-packages
Running virtualenv with interpreter /home/ec2-user/.pyenv/versions/miniconda-3.10.1/bin/python
New python executable in env2/bin/python
ERROR: The executable env2/bin/python is not functioning
ERROR: It thinks sys.prefix is u'/usr' (should be u'/home/ec2-user/ner/env2')
ERROR: virtualenv is not compatible with this system or executable
```

Any pointers?

yyuu:
It might be an issue of [virtualenv](https://github.com/pypa/virtualenv) with miniconda's python executable. Anyway virtualenv is virtualenv, not pyenv.

eregon:
@yyuu 
> It seems that the CPython's `sys.executable` requires the path to the `python` executable should be in `PATH`.

If that's a case, it sounds like a CPython bug.
But it doesn't seem to be the case for me:
```
$ which python    
/usr/bin/python
$ PATH=/nothing /usr/bin/python -c 'import sys; print(sys.executable)'
/usr/bin/python
```
Prepending `/usr/bin` causes many issues (#789, #1169), can we avoid it?

blueyed:
> If that's a case, it sounds like a CPython bug.

Good point.

But it's still the case:

```
% (PATH=/nothing exec -a python /usr/bin/python -c 'import sys; print("exe", repr(sys.executable))';)                                                                            
exe ''
```
Python 3.7.4.

[...]

PR #987, opened 2017-09-21
prevent changing PATH in pyenv shims

### LINKED ISSUE #985: pyenv shims always change PATH
The pyenv shims for any executable change the PATH for the current process and child processes. This affects usage of any python program that starts child processes like `envdir`, `pew` and can be demonstrated using `python` itself.

The shim will prepend the `bin` of the current activated python and also the pyenv `libexec` folder. For example: `/Users/johria/.pyenv/versions/3.6.2/bin:/usr/local/Cellar/pyenv/1.1.3/libexec` will get prepended to the `PATH`.

This can most easily be demonstrated via the shim for `python` itself:

checking the current PATH:
```
$ echo "$PATH"
bin:/Users/johria/.nodenv/shims:/Users/johria/.rbenv/shims:/Users/johria/.pyenv/shims:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

checking the PATH modified by the pyenv shim:
```
$ python -c "import os; print(os.environ['PATH'])"
/Users/johria/.pyenv/versions/3.6.2/bin:/usr/local/Cellar/pyenv/1.1.3/libexec:bin:/Users/johria/.nodenv/shims:/Users/johria/.rbenv/shims:/Users/johria/.pyenv/shims:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

as you can see pyenv has changed the PATH

is there any way to allow shims to work without changing the PATH? I realize this is the root of my issues. 

cc @blueyed @yyuu

AlJohri:
fixes #985. this introduces two changes:

- remove `libexec` from PATH
- remove actual python binary location from PATH

cc @blueyed cc @yyuu for review

also do either of you remember some more details regarding #98? was this specific to virtualenv? I don't see any documentation saying `sys.executable` **needs** to point to the python binary:

https://docs.python.org/3/library/sys.html#sys.executable
>A string giving the absolute path of the executable binary for the Python interpreter, on systems where this makes sense. If Python is unable to retrieve the real path to its executable, sys.executable will be an empty string or None.

reading through some of the virtualenv sourcecode, the issue seemed to be specific to virtualenv

to elaborate, this is the quintessential bug from changing of the path: https://github.com/jezdez/envdir/issues/62 (tl;dr: programs that create subshells end up with a different PATH)

### DIFF
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -44,7 +44,12 @@ shift 1
 
 # reset path to original path
 # https://github.com/pyenv/pyenv/issues/985
-export PATH="$PYENV_ORIGINAL_PATH"
+# if installing package via pip/conda/easy_install, do not reset path as pyenv-which will be needed for auto rehash
+if [[ "$PYENV_COMMAND" != "pip" && "$PYENV_COMMAND" != "conda" && "$PYENV_COMMAND" != "easy_install" ]]; then
+  if [ -z "$PYENV_ORIGINAL_PATH" ]; then
+  	export PATH="$PYENV_ORIGINAL_PATH"
+  fi
+fi
 
 if [ "$PYENV_COMMAND" = "virtualenv" ]; then
   # CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -41,7 +41,15 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-# CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
-# https://github.com/pyenv/pyenv/issues/98
-export PATH="${PYENV_BIN_PATH}:${PATH}"
+
+# reset path to original path
+# https://github.com/pyenv/pyenv/issues/985
+export PATH="$PYENV_ORIGINAL_PATH"
+
+if [ "$PYENV_COMMAND" = "virtualenv" ]; then
+  # CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
+  # https://github.com/pyenv/pyenv/issues/98
+  export PATH="${PYENV_BIN_PATH}:${PATH}"
+fi
+
 exec -a "$PYENV_COMMAND" "$PYENV_COMMAND_PATH" "$@"
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -86,6 +86,7 @@ bin_path="$(abs_dirname "$0")"
 for plugin_bin in "${PYENV_ROOT}/plugins/"*/bin; do
   PATH="${plugin_bin}:${PATH}"
 done
+export PYENV_ORIGINAL_PATH="$PATH"
 export PATH="${bin_path}:${PATH}"
 
 PYENV_HOOK_PATH="${PYENV_HOOK_PATH}:${PYENV_ROOT}/pyenv.d"


blueyed:
Great!
I cannot reproduce #98 with this PR (`virtualenv --python="$(which python)" FOO`.
Using system Python 3.6.2 and virtualenv 15.1.0.

libexec/pyenv-exec
```diff
@@ -41,7 +41,20 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-# CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
-# https://github.com/pyenv/pyenv/issues/98
-export PATH="${PYENV_BIN_PATH}:${PATH}"
+
+# reset path to original path
+# https://github.com/pyenv/pyenv/issues/985
+# if installing package via pip/conda/easy_install, do not reset path as pyenv-which will be needed for auto rehash
```
blueyed: > pyenv-which will be needed for auto rehash

This should be handled in a better way, e.g. by exporting/setting the location for it explicitly or something similar.

blueyed:
> I cannot reproduce

should have looked at the diff before.. ;) (it gets special-cased still)

blueyed:
Maybe some more advanced exec would help here?
https://docs.python.org/3.6/library/os.html#os.execl ?

And maybe just using `exec -a` does it?

>     exec [-cl] [-a name] [command [arguments]]
      If command is specified, it replaces the shell.  No new process is created.  The argu‐
      ments become the arguments to command.  If the -l option is supplied, the shell places
      a  dash  at  the  beginning  of  the  zeroth argument passed to command.  This is what
      login(1) does.  The -c option causes command to be executed with an empty environment.
      If  -a  is supplied, the shell passes name as the zeroth argument to the executed com‐
      mand.  If command cannot be executed for some reason, a non-interactive  shell  exits,
      unless  the  execfail  shell option is enabled.  In that case, it returns failure.  An
      interactive shell returns failure if the file cannot be executed.  If command  is  not
      specified, any redirections take effect in the current shell, and the return status is
      0.  If there is a redirection error, the return status is 1.

libexec/pyenv-exec
AlJohri: Okay, so looking at the auto rehash more carefully, it seems like it needs access to `pyenv-which` and `pyenv-rehash`. I was trying to keep modifications localized in my initial PR.

I can instead, where possible, make `pyenv-xxx` into an absolute reference since these are pre-packaged plugins so I know exactly where they are. I can export the path to the main pyenv plugin directory and re-use it in the plugin (or get the path to the main pyenv plugin directory again within the plugin itself). Does this sound like a better approach to you?

It seems as though the PATH changing is primarily due to the flexibility of running `pyenv-xyz` arbitrarily in plugins.

blueyed:
I've just seen an issue where
`bash /home/user/.pyenv/libexec/pyenv-exec python3.6 -c import sys;out=sys.stdout;getattr(out, "buffer", out).write(sys.prefix.encode("utf-8"))` was spinning with `tox -e py36-1.11-stripe128` in the pinax-stripe project.
tox was hanging at the beginning "py36-1.11-stripe128 create: /home/user/Vcs/pinax-stripe/.tox/py36-1.11-stripe128"

AlJohri:
cool, I'll debug with the pinax project, I got it running locally with vanilla pyenv.

but I'm not sure I understand how `exec -a` would help in this case? It seems as though `exec -a` is already being used in the last line of `pyenv-exec`:

```
exec -a "$PYENV_COMMAND" "$PYENV_COMMAND_PATH" "$@"
```

blueyed:
> exec -a

Haven't noticed - but maybe it works better without it then?!

libexec/pyenv-exec
blueyed: Cannot `PYENV_ROOT` be used to build the absolute path?

blueyed:
Check out https://github.com/pyenv/pyenv/pull/1169.

It uses the full path with `exec -a`, and therefore Python does not look up the (relative) argv0 in $PATH, and we do not need to mangle the $PATH anymore.

I think it would be good to additionally do not export/have libexec and plugin's bins in there, which could be solved by saving and using the original PATH before the `exec` - this however is not that critical.

franknarf8:
Hello,

We provisionned our build servers to use `pyenv` to manage different python versions and it worked fantastically until we stumbled upon this issue.

Across our CI/Buildchain, we are using different build tools (including external ones) requiring different versions of python (python2&python3). 

Now we have the issue that we are running a top-level tool with pyenv's python 3.5.5 (which prepends its `[...]/bin` directory to the PATH), this tool is calling an external building tool which at some point also calls `python` (without specifying the version e.g. `python2`, but since python3.5.5's `[...]/bin` directory is in the path, `python` resolves to python 3.5.5). The thing is, this external building tool is expecting `python` to be python 2, not python 3.5.5.

We did set `pyenv global 2.7.15 3.5.5` initially expecting that `python` would always default to python 2.7.15 and it took us a while to figure out that pyenv was causing this.

Is there a way to disable shims altering PATH? Which workaround would you advise in this situation?

Thanks!

\- Frank

efiop:
@AlJohri @blueyed Hi guys! Any progress on this? We are running into this issue on our `dvc run` command https://github.com/iterative/dvc/issues/2506 and are using a workaround https://github.com/iterative/dvc/pull/2525 for now, but we would be interested in a proper long-term solution and would be happy to help to get this fixed. Is https://github.com/pyenv/pyenv/pull/1169 a better solution? If so, do you need any help with it? Should we close this PR in favor of https://github.com/pyenv/pyenv/pull/1169 ?

blueyed:
@efiop 
Thanks for the interest.
IIRC #1169 would be better if it fixes the same issue.
Help over there would be appreciated, e.g. with regard to (missing?) tests and/or review - have not looked into it in detail now.

blueyed:
Closing in favor of #1169 for now.

PR #992, opened 2017-09-28
python-build: add anaconda[23]-5.0.0

shoichiaizawa:
This PR consists of a single commit adding two new files, which allow the user to install both **Anaconda2 5.0.0** and **Anaconda3 5.0.0**.

The tests have been done only on macOS, but it should also work on Linux; please check if it is buildable on Linux.

The SHA-256 checksum for each Anaconda shell file is copied from the following pages:

- https://docs.anaconda.com/anaconda/install/hashes/Anaconda2-5.0.0-Linux-x86.sh-hash
- https://docs.anaconda.com/anaconda/install/hashes/Anaconda2-5.0.0-Linux-x86_64.sh-hash
- https://docs.anaconda.com/anaconda/install/hashes/Anaconda2-5.0.0-MacOSX-x86_64.sh-hash
- https://docs.anaconda.com/anaconda/install/hashes/Anaconda3-5.0.0-Linux-x86.sh-hash
- https://docs.anaconda.com/anaconda/install/hashes/Anaconda3-5.0.0-Linux-x86_64.sh-hash
- https://docs.anaconda.com/anaconda/install/hashes/Anaconda3-5.0.0-MacOSX-x86_64.sh-hash

### DIFF
--- a/plugins/python-build/share/python-build/anaconda2-5.0.0
+++ b/plugins/python-build/share/python-build/anaconda2-5.0.0
@@ -1,0 +1,19 @@
+case "$(anaconda_architecture 2>/dev/null || true)" in
+"Linux-x86" )
+  install_script "Anaconda2-5.0.0-Linux-x86" "https://repo.continuum.io/archive/Anaconda2-5.0.0-Linux-x86.sh#a3ed8769d20d55a41c04cf7c04e81c95974ea8eb614afab7bbc0c06fa6a52437" "anaconda" verify_py27
+  ;;
+"Linux-x86_64" )
+  install_script "Anaconda2-5.0.0-Linux-x86_64" "https://repo.continuum.io/archive/Anaconda2-5.0.0-Linux-x86_64.sh#58a7117f89c40275114bf7e824a613a963da2b0fe63f2ec3c1175fea785b468e" "anaconda" verify_py27
+  ;;
+"MacOSX-x86_64" )
+  install_script "Anaconda2-5.0.0-MacOSX-x86_64" "https://repo.continuum.io/archive/Anaconda2-5.0.0-MacOSX-x86_64.sh#d85198c63657924fae11b6ea5961f50d81d09a1185d6f0a9a9d5bc69eb788ccc" "anaconda" verify_py27
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
+    echo ": The binary distribution of Anaconda2 is not available for $(anaconda_architecture 2>/dev/null || true)."
+    echo
+  } >&2
+  exit 1
+  ;;
+esac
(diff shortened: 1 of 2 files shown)


shoichiaizawa:
I am experiencing some bug that I am not sure what is caused by.

Up until now, I have been using the Homebrew version of pyenv, and since I wanted to make this PR, I decided to use pyenv with the manual git checkout.

After adding the two files (`anaconda2-5.0.0` and `anaconda3-5.0.0`) to the `python-build` directory, firstly I installed `anaconda3-5.0.0` followed by `3.6.2` and `2.7.14`. These python versions were installed in this order.

The installations for those three versions succeeded but I encountered an error when I changed the Python global version to `3.6.2` or `2.7.14`. And with `3.6.2` or `2.7.14` selected, when I ran some commands generating coloured output, it got greyed out, except for using the firstly installed anaconda3-5.0.0, with which output was coloured as expected; here is the error:

### Running `nvm ls` causes `pyenv: tput: command not found`:

Here is an expected output:

coloured-nvm-ls

Here is a greyed output:

greyed-nvm-ls

### Running [`archey`](https://github.com/obihann/archey-osx) command causes the apple logo in grey instead of 5 colours:

Here is an expected output:
  
archey-coloured

Here is a greyed output:

archey-greyed

This did not happen in the Homebrew version of pyenv.

I am not sure this is reproducible or unique to my dev environment 😢 

Any thoughts regarding this?

shoichiaizawa:
Hmm, it seems that I have the same problem for the Homebrew version of pyenv 😢 And the order of the Python version installations did not matter in the end, but this exact problem occurred only for `anaconda2-5.0.0` and `anaconda3-5.0.0`. With `anaconda3-4.4.0` set as global, I did not encounter this whatsoever.

I suspect that there is something wrong with the latest Anaconda distributions but I am not certain of it.

I hope I am clear enough to you with all the explanations I made.

yyuu:
@shoichiaizawa

Anaconda's prefix may contain some executables which is a part of base systems, like `tput` (a part of `ncurses`). And, the existence of pyenv's shim scripts for these executable may conceal these executables in your base system. This must be the cause of the problem with `nvm`.

The possible/available workarounds for the issues is follows.

1. Prevent creating pyenv's shim script for these system executables
2. Fallback to non-pyenv executables if the given command was not found in active pyenv versions

The 1 is implemented as [a hook script of pyenv](https://github.com/pyenv/pyenv/blob/master/pyenv.d/rehash/conda.bash). By default, pyenv is configured to skip creating shim scripts [based on the blacklist](https://github.com/pyenv/pyenv/blob/master/pyenv.d/rehash/conda.d/default.list). Adding some text file at `$(pyenv root)/pyenv.d/rehash/conda.d/*.list` and add `tput` in it should let pyenv to skip creating shim script for `tput`. (or, it might be better to add it in the default blacklist; PR welcome)

The 2 is implemented as a plugin named [pyenv-which-ext](https://github.com/pyenv/pyenv-which-ext). It will try to look up the given command from original `$PATH` (=~ outside of pyenv) if it was not found inside current active pyenv version. This implicit behavior may be confusing somehow, but it'll work automatically without managing the blacklist by yourself.

I believe either will work for you. Please give it a try later.

PR #997, opened 2017-10-03
Turn off auto-update for conda builds.

chebee7i:
Unless specified otherwise, conda will auto-update to the latest version, resulting in a conda install that differs from what was requested. This commit makes the necessary change for the user, but I could also imagine that we might want the user to specify this environment variable themselves (or through a .condarc file).


blueyed:
@chebee7i 
Please update/reply.

PR #1010, opened 2017-10-09
Fix broken download link

jab:
Looks like 5045d88 contained some typos, including the URL for the pypy-5.9-linux_x86_64 download, so that request results in a 404 (there is a spurious "-1" in the filename).


jab:
Thanks for merging so quickly!
Come to think of it, could some basic tests (along the lines of checking that `curl -I` gives 200 for each link) be worth adding to prevent typos like this from going unnoticed in the future?

blueyed:
Would make sense.. maybe even only for files changed in a PR.

(Related IIRC): https://github.com/pyenv/pyenv/pull/617

PR #1023, opened 2017-10-29
Clean up some pypy build defs

joshfriend:
I was thinking that we could remove all the pypy definitions for pypy-portable since the definitions for mainstream pypy already will install pypy-portable if the platform is not supported by the official binaries.

Thoughts?


Samureus:
Let's do this.

Issue #993, opened 2017-09-29
macOS High Sierra: ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?


uber1geek:
I am having this issue on mac os high sierra(10.13) even with setting CFLAGS and LDFLAGS correctly using brew --prefix

BruceFrankWang:
@uber1geek 

I can finish this step successfully.

first:
brew install readline openssl xz

and Homebrew said both readline and openssl is keg-only. so I wrote a .sh file to export environment variables:

# ~/.env-exporter
# zlib
export CFLAGS="-I$(xcrun --show-sdk-path)/usr/include"

# Add readline path to environment variable
export CFLAGS="-I$(brew --prefix readline)/include $CFLAGS"
export LDFLAGS="-L$(brew --prefix readline)/lib $LDFLAGS"

# Add openssl path to environment variable
export CFLAGS="-I$(brew --prefix openssl)/include $CFLAGS"
export LDFLAGS="-L$(brew --prefix openssl)/lib $LDFLAGS"

# Add SQLite path to environment variable
export CFLAGS="-I$(brew --prefix sqlite)/include $CFLAGS"
export LDFLAGS="-L$(brew --prefix sqlite)/lib $LDFLAGS"


every time I want install a Python:
 # source ~/.env-exporter
# pyenv install -v 2.7.13

I failed at the step of installing pip, cause of could not build the pyexpat module.

Hope helpful, and wish I can install 2.7.13 & 3.6.2

natecox:
I'm currently experiencing this issue on a clean install of High Sierra. I was able to get openssl recognized using the above script, however I'm now failing because `pyexpat` can't be found (as referenced in the bottom of the last post). Is this a known issue?

uber1geek:
I would suggest removing the old installation and install using homebrew. The problem turned out to be the openssl hadn't actually successfully installed (even though homebrew reported that it had). run brew uninstall openssl && brew install openssl && CFLAGS="-I$(brew --prefix openssl)/include" LDFLAGS="-L$(brew --prefix openssl)/lib" pyenv install 3.6.2

uber1geek:
or simply,  `brew install pyenv`

BruceFrankWang:
@uber1geek You are right. Do not install pyenv via pyenv-installer, instead of brew.

jlott:
[...]

Issue #950, opened 2017-07-17
ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?


jingle0927:
➜  ~ pyenv install 3.5.2
Installing Python-3.5.2...
patching file Lib/venv/scripts/posix/activate.fish
ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?

Please consult to the Wiki page to fix the problem.
<a  target="_blank">https://github.com/pyenv/pyenv/wiki/Common-build-problems</a>

BUILD FAILED (Deepin 15.4 using python-build 20160602)

Inspect or clean up the working tree at /tmp/python-build.20170717074120.10900
Results logged to /tmp/python-build.20170717074120.10900.log

Last 10 log lines:
(cd /home/zonzely/.pyenv/versions/3.5.2/share/man/man1; ln -s python3.5.1 python3.1)
if test "xupgrade" != "xno"  ; then \
        case upgrade in \
                upgrade) ensurepip="--upgrade" ;; \
                install|*) ensurepip="" ;; \
        esac; \
         ./python -E -m ensurepip \
                $ensurepip --root=/ ; \
fi
Ignoring ensurepip failure: pip 8.1.1 requires SSL/TLS

ChrisAitken:
What os and version are you using ? 

Also the fix is right there in the message you https://github.com/pyenv/pyenv/wiki/Common-build-problems#error-the-python-ssl-extension-was-not-compiled-missing-the-openssl-lib just copy and paste the script for your specific os

elderone:
Debian GNU/Linux 9.1 (stretch)
pyenv 1.1.3
`CFLAGS=-I/usr/include/openssl LDFLAGS=-L/usr/lib pyenv install -v 3.2.3`

elderone:
```
building '_ssl' extension
gcc -pthread -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/usr/include/openssl -I/home/elderone/.pyenv/versions/3.2.3/include -fPIC -IInclude -I. -I./Include -I/home/elderone/.pyenv/versions/3.2.3/include
 -I/usr/include/x86_64-linux-gnu -I/usr/local/include -I/tmp/python-build.20170807132419.25380/Python-3.2.3 -c /tmp/python-build.20170807132419.25380/Python-3.2.3/Modules/_ssl.c -o build/temp.linux-x86_64-3.2/tm
[...]

PR #1025, opened 2017-10-30
Ignore LibreSSL bundled with macOS 10.13

jamieconnolly:
I've encountered the same error about a missing OpenSSL lib that has affected other users on macOS 10.13 (See #950 and #993). I've also seen the workaround to this problem (See [Common-build-problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems#error-the-python-ssl-extension-was-not-compiled-missing-the-openssl-lib)).

I haven't encountered this issue with `rbenv`, so I went searching and they've handled the issue inside `ruby-build`.

It seems that High Sierra now bundles LibreSSL, and it's this check that is causing `python-build` to not locate the openssl package.

I've added in the same check that `ruby-build` does, and it works on my local machine. I've not done any tests on this as I have no idea where to start.

See rbenv/ruby-build#1095 and rbenv/ruby-build#1104.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1385,8 +1385,9 @@ use_homebrew_readline() {
 }
 
 has_broken_mac_openssl() {
-  is_mac &&
-  [[ "$(/usr/bin/openssl version 2>/dev/null || true)" = "OpenSSL 0.9.8"?* ]] &&
+  is_mac || return 1
+  local openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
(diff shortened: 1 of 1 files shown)


joshfriend:
Interesting observation, thanks!

I think we usually fix these things by merging the upstream changes from rbenv. When was the last time that happened, @yyuu?

yyuu:
I don't remember when was the last time I merged the changes from ruby-build.... 😵 At least I can say this change is equivalent to ruby-build and must be harmless. I'll merge this soon.

PR #1031, opened 2017-11-08
Get number of processors using getconf over cpuinfo grep

scop:


yyuu:
@scop the usage of getconf(1) looks nice :+1: I think this is beneficial to get merged into upstream https://github.com/rbenv/ruby-build. Could you open a PR to it as well?

scop:
https://github.com/rbenv/ruby-build/pull/1150

PR #1032, opened 2017-11-08
Add install_git tag support

scop:
...and use it to add micropython-1.9.3.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -482,6 +482,13 @@ fetch_git() {
   local git_url="$2"
   local git_ref="$3"
 
+  local fetch_args=(${git_ref})
+  if [ ${#fetch_args[@]} -eq 1 ]; then  # just a branch name
+    fetch_args[0]="+${fetch_args[0]}:${fetch_args[0]}"
+  elif [[ ${#fetch_args[@]} -eq 2 && ${fetch_args[0]} == tag ]]; then
+    git_ref="${fetch_args[1]}"
+  fi
+
   echo "Cloning ${git_url}..." >&2
 
   if type git &>/dev/null; then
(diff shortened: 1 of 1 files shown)


plugins/python-build/bin/python-build
```diff
@@ -482,6 +482,13 @@ fetch_git() {
   local git_url="$2"
   local git_ref="$3"
 
+  local fetch_args=(${git_ref})
+  if [ ${#fetch_args[@]} -eq 1 ]; then  # just a branch name
+    fetch_args[0]="+${fetch_args[0]}:${fetch_args[0]}"
+  elif [[ ${#fetch_args[@]} -eq 2 && ${fetch_args[0]} == tag ]]; then
```
joshfriend: why does this have to be different from fetching a branch? both branches and tags are just refs to specific commits and you should be able to get either one with the same command.

scop: Fetching `+tagname:tagname` doesn't work for tags, which is what python-build does with the refs. git fetch man page documents `tag <tag>` to fetch up to _tag_, and this is what this patch accomplishes.

Granted, it doesn't look at all clean to me and I thought it would've been easier to implement, but this was the smallest change I got working. (Tested with a bunch of micropython fetches and builds.)

plugins/python-build/share/python-build/micropython-1.9.3
```diff
@@ -0,0 +1,2 @@
+#require_gcc
+install_git micropython-1.9.3 https://github.com/micropython/micropython "tag v1.9.3" micropython
```
yyuu: For this specific case, what we need to do is not checking out git tag, downloading archive of the tagged release, in my opinion. We can download the archive of v1.9.3 from https://github.com/micropython/micropython/releases/download/v1.9.3/micropython-1.9.3.tar.gz

scop:
Ok, micropython 1.9.3 moved to #1033. In order to test this implementation, here is its git tag based install recipe:

```
install_git micropython-1.9.3 https://github.com/micropython/micropython "tag v1.9.3" micropython
```

PR #1036, opened 2017-11-13
Fix $PWD changes when executing Python

zachriggle:
Currently, pyenv uses `cd` in some spots to determine real paths, by abusing `$PWD`.

This works in the general case, but fails when `PWD=/proc/self/cwd`.  When `pyenv` attempts to change directories back to the "old" directory, it fails to do so.

This Pull Request wraps a few of these cases in sub-shells, and changes the code in one instance that uses `$OLDPWD`.

Specifically, this issue becomes apparent because it breaks the build for Android Open Source Project when `pyenv` is on `$PATH`, even when `pyenv global system` is in use.  This is because Clang upstream compilers now use `#!/usr/bin/env python` instead of `#!/usr/bin/python` in their shebang for the Python wrapper scripts.

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -36,9 +36,10 @@ else
   }
 
   abs_dirname() {
-    local cwd="$PWD"
     local path="$1"
 
+    # Use a subshell to avoid changing the current path
+    (
     while [ -n "$path" ]; do
       cd "${path%/*}"
       local name="${path##*/}"
@@ -46,7 +47,7 @@ else
     done
 
     pwd
-    cd "$cwd"
+    )
   }
 fi
 
@@ -72,11 +73,13 @@ fi
 
 if [ -z "${PYENV_DIR}" ]; then
   PYENV_DIR="$PWD"
-else
-  cd "$PYENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$PYENV_DIR'"
-  PYENV_DIR="$PWD"
-  cd "$OLDPWD"
 fi
+
+if [ ! -d "$PYENV_DIR" ] || [ ! -e "$PYENV_DIR" ]; then
+  abort "cannot change working directory to \`$PYENV_DIR'"
+fi
+
+PYENV_DIR=$(cd "$PYENV_DIR" && echo "$PWD")
 export PYENV_DIR
 
 
--- a/libexec/pyenv-hooks
+++ b/libexec/pyenv-hooks
@@ -37,10 +37,10 @@ resolve_link() {
 }
 
 realpath() {
-  local cwd="$PWD"
   local path="$1"
   local name
-
+  # Use a subshell to avoid changing the current path
+  (
   while [ -n "$path" ]; do
     name="${path##*/}"
     [ "$name" = "$path" ] || cd "${path%/*}"
@@ -48,7 +48,7 @@ realpath() {
   done
 
   echo "${PWD}/$name"
-  cd "$cwd"
+  )
 }
 fi
 
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -44,10 +44,11 @@ if ! enable -f "${BASH_SOURCE%/*}"/pyenv-realpath.dylib realpath 2>/dev/null; th
   }
 
   realpath() {
-    local cwd="$PWD"
     local path="$1"
     local name
 
+    # Use a subshell to avoid changing the current path
+    (
     while [ -n "$path" ]; do
       name="${path##*/}"
       [ "$name" = "$path" ] || cd "${path%/*}"
@@ -55,7 +56,7 @@ if ! enable -f "${BASH_SOURCE%/*}"/pyenv-realpath.dylib realpath 2>/dev/null; th
     done
 
     echo "${PWD}/$name"
-    cd "$cwd"
+    )
   }
 fi
 
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -60,9 +60,10 @@ resolve_link() {
 }
 
 abs_dirname() {
-  local cwd="$(pwd)"
   local path="$1"
 
+  # Use a subshell to avoid modifying the current path
+  (
   while [ -n "$path" ]; do
     cd "${path%/*}"
     local name="${path##*/}"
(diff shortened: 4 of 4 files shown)


libexec/pyenv
```diff
@@ -36,17 +36,18 @@ else
   }
 
   abs_dirname() {
-    local cwd="$PWD"
     local path="$1"
 
+    # Use a subshell to avoid changing the the current path
```
blueyed: s/the the/the/

libexec/pyenv
```diff
@@ -72,11 +73,10 @@ fi
 
 if [ -z "${PYENV_DIR}" ]; then
   PYENV_DIR="$PWD"
-else
-  cd "$PYENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$PYENV_DIR'"
-  PYENV_DIR="$PWD"
-  cd "$OLDPWD"
 fi
+
+PYENV_DIR=$(cd "$PYENV_DIR" && echo "$PWD") || abort "cannot change working directory to \`$PYENV_DIR'"
```
blueyed: Why is this necessary after all?
Could also check only if it is a directory then?!

zachriggle: No idea, I'm trying to respect the current code as closely as possible

zachriggle:
Merging to "upstream" `rbenv` only makes sense if it's possible to merge from `rbenv` to `pyenv` (which I don't think is the case).  I don't use `rbenv` and don't have a good way to test the changes.  I'll open an issue on the `rbenv` issue tracker pointing at this pull request.

libexec/pyenv
zachriggle: Done

libexec/pyenv
zachriggle: I've added that code back in to test `-d` and `-x`.  I'm not sure why the check is there in the first place, but there is a test for this specific check.

blueyed:
> if it's possible to merge from rbenv to pyenv (which I don't think is the case)

It is.. I've put some effort into it a while ago, so that it works actually..!

yyuu:
Interesting 👀 

However, I'm still not understanding what is causing the problem in Android Open Source Project.... `pwd` will change inside pyenv before executing `python` itself. Though it should be restored to the original directory before pyenv actually start executing `python`.

Could you provide some minimal script to reproduce the problem...? I've tried writing some, but I could not find issues so far.

```sh
#!/usr/bin/env bash
set -e -o pipefail
export PYENV_ROOT="$(mktemp -d /tmp/pyenv.XXXXXXXX)"
git clone --depth=1 https://github.com/pyenv/pyenv.git "${PYENV_ROOT}"
export PATH="${PYENV_ROOT}/bin:/usr/bin:/bin"
eval "$(pyenv init -)"
# create a dummy version to let pyenv to generate a shim script for python
mkdir -p "${PYENV_ROOT}/versions/foo/bin"
echo -e "#!/bin/sh\necho python \$@" > "${PYENV_ROOT}/versions/foo/bin/python"
chmod 755 "${PYENV_ROOT}/versions/foo/bin/python"
pyenv global system
pwd
cd /proc/self/cwd
python -c 'import os;print(repr(os.getcwd()))'
rm -fr "${PYENV_ROOT}"
```

The `pwd` of `python` process is set as it should be, in my understanding.

```
% ./x.sh
Cloning into '/tmp/pyenv.u1wafgCG'...
remote: Counting objects: 577, done.
remote: Compressing objects: 100% (398/398), done.
remote: Total 577 (delta 268), reused 262 (delta 89), pack-reused 0
Receiving objects: 100% (577/577), 253.72 KiB | 307.00 KiB/s, done.
Resolving deltas: 100% (268/268), done.
/home/yyuu
'/home/yyuu'
```

zachriggle:
The problem is that the `realpath` functions in `pyenv` changes directories before `python` is ever executed:

```
sh -c 'PWD=/proc/self/cwd; cwd="$PWD"; realpath .; cd /; realpath .; cd "$cwd"; realpath .'
# /home/pwntools
# /
# /
```

libexec/pyenv
```diff
@@ -72,11 +73,13 @@ fi
 
 if [ -z "${PYENV_DIR}" ]; then
   PYENV_DIR="$PWD"
-else
-  cd "$PYENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$PYENV_DIR'"
-  PYENV_DIR="$PWD"
-  cd "$OLDPWD"
 fi
+
+if [ ! -d "$PYENV_DIR" ] || [ ! -e "$PYENV_DIR" ]; then
```
yyuu: I think just checking with `[ ! -d "${PYENV_DIR}" ]` must be sufficient here.

libexec/pyenv
```diff
@@ -72,11 +73,13 @@ fi
...
+  abort "cannot change working directory to \`$PYENV_DIR'"
+fi
+
+PYENV_DIR=$(cd "$PYENV_DIR" && echo "$PWD")
```
yyuu: Just in case, please add quoting around the `$(...)`.

```sh
PYENV_DIR="$(cd "$PYENV_DIR" && echo "$PWD")"
```

libexec/pyenv
blueyed: > to get the absolute path

This is what I've meant mostly.
At least there does not seem to be a failing test when removing the `PYENV_DIR=$(cd "$PYENV_DIR" && echo "$PWD")`.
Since this spawns a subshell, it is not really cheap, and should be avoided if not necessary.

libexec/pyenv
blueyed: Might not be necessary altogether: https://github.com/pyenv/pyenv/pull/1036#discussion_r151895707

yyuu: I should have followed the discussion little more.... Anyway it's good enough to me to proceed for merging~

yyuu:
Since pyenv is used in various CI provider, I've merged this into pyenv's master to restore the corrupted build of the AOSP.

I or someone who can work for it may be going to port this change to https://github.com/rbenv/rbenv to keep them _compatible_. Because the use of `cd` there in rbenv/pyenv is not anything essential, I don't foresee any issues merging this into rbenv as well.

zachriggle:
Thanks!

PR #1051, opened 2017-12-05
Check wget version iff wget is going to be used

LWisteria:
Do not check when tarball is cached

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -330,6 +330,11 @@ http() {
   elif type curl &>/dev/null; then
     "http_${method}_curl" "$url" "$file"
   elif type wget &>/dev/null; then
+    # SSL Certificate error with older wget that does not support Server Name Indication (#60)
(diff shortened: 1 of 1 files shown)


plugins/python-build/bin/python-build
```diff
@@ -2051,12 +2056,6 @@ if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
   fi
 fi
 
-# SSL Certificate error with older wget that does not support Server Name Indication (#60)
-if ! command -v curl 1>/dev/null 2>&1 && [[ "$(wget --version 2>/dev/null || true)" = "GNU Wget 1.1"[0-3]* ]]; then
```
yyuu: Honestly speaking I prefer to perform checking beforehand. However, the condition here is not appropriate anymore since it's not checking `aria2c`.

At least I can say that to check the availability of wget only if it is going to be used is the appropriate behavior.

PR #1055, opened 2017-12-13
Fix pyenv install on Solaris / Illumos

klimkjar:
"pyenv install" crashes on Solaris with an empty log file. Adding support for the proper Solaris getconf call in num_cpu_cores fixed it. Tested and working under OmniOS CE r151024.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -155,8 +155,11 @@ num_cpu_cores() {
   Darwin | *BSD )
     num="$(sysctl -n hw.ncpu 2>/dev/null || true)"
     ;;
+  SunOS )
+    num="$(getconf NPROCESSORS_ONLN 2>/dev/null)"
+    ;;
   * )
-    num="$({ getconf _NPROCESSORS_ONLN || getconf NPROCESSORS_ONLN ||
+    num="$({ getconf _NPROCESSORS_ONLN ||
              grep -c ^processor /proc/cpuinfo; } 2>/dev/null)"
     num="${num#0}"
     ;;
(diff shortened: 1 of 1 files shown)


yyuu:
I remembered that the code around there has changed recently in #1031

I tried to look up if there's some POSIX standard or something like that for it, but I couldn't find it so far....

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/getconf.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/sysconf.html

plugins/python-build/bin/python-build
```diff
@@ -156,7 +156,7 @@ num_cpu_cores() {
     num="$(sysctl -n hw.ncpu 2>/dev/null || true)"
     ;;
   * )
-    num="$({ getconf _NPROCESSORS_ONLN ||
+    num="$({ getconf _NPROCESSORS_ONLN || getconf NPROCESSORS_ONLN ||
```
yyuu: The `*` case looks like it is assuming the system is very similar to Linux.... To add support for Solaris/Illumos, I prefer to add a case condition for Solaris.

klimkjar: Sure - I'll add a SunOS case statement instead.

plugins/python-build/bin/python-build
```diff
@@ -155,6 +155,9 @@ num_cpu_cores() {
   Darwin | *BSD )
     num="$(sysctl -n hw.ncpu 2>/dev/null || true)"
     ;;
+  SunOS )
+    num="$(getconf NPROCESSORS_ONLN 2>/dev/null)"
```
yyuu: Just in case, please add `|| true` to `getconf` command line.

yyuu:
BTW, the `python-build` script is inheriting most of lines with upstream https://github.com/rbenv/ruby-build. Since this will be worth to merge for ruby-build too, could you prepare a patch and send a PR to it as well once you get a chance?

klimkjar:
I've added a fallback to true as well. I will have a look at the ruby-build script. Thank you for your work on pyenv!

PR #1059, opened 2017-12-25
Add ability to build using readline and openssl from pkgsrc

jdputsch:
This is a simple change to pyenv's python-build plugin allowing it to use "pkgsrc" instead of "homebrew" for the source of "readline" and "openssl" when building python.

The assumption is that a user will either be using "homebrew" or "pkgsrc" but not both.


plugins/python-build/share/python-build/pypy3.5-5.10
```diff
@@ -0,0 +1,57 @@
+case "$(pypy_architecture 2>/dev/null || true)" in
```
jayvdb: these were added as `plugins/python-build/share/python-build/pypy3.5-5.10.0`

needs a rebase.

PR #1070, opened 2018-01-09
Allow binary installs of PyPy on Xenial

BanzaiMan:
Ideally, the `require_distro` function is a little more intelligent to allow Xenial to be detected, but this should be good enough for now.

### DIFF
--- a/plugins/python-build/share/python-build/pypy2.7-5.10.0
+++ b/plugins/python-build/share/python-build/pypy2.7-5.10.0
@@ -5,9 +5,7 @@ case "$(pypy_architecture 2>/dev/null || true)" in
   fi
   ;;
 "linux64" )
-  if require_distro "Ubuntu 14.04" 1>/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


plugins/python-build/share/python-build/pypy2.7-5.10.0
```diff
@@ -7,6 +7,8 @@ case "$(pypy_architecture 2>/dev/null || true)" in
 "linux64" )
   if require_distro "Ubuntu 14.04" 1>/dev/null 2>&1; then
```
yyuu: It seems that `require_distro` can take multiple arguments to require any of given patterns.... We must be able to just append xenial after the trusty here.

https://github.com/pyenv/pyenv/blob/66412f556db1e352679b144ec17bb36bf55d3ee6/plugins/python-build/share/python-build/pypy2-5.3.1#L3

plugins/python-build/share/python-build/pypy3.5-5.10.0
```diff
@@ -7,6 +7,8 @@ case "$(pypy_architecture 2>/dev/null || true)" in
 "linux64" )
   if require_distro "Ubuntu 14.04" 1>/dev/null 2>&1; then
```
yyuu: ditto

BanzaiMan:
Ah, cool. Thanks!

Issue #174, opened 2014-05-21
pyenv shims fail with concurrent processes in a build system


rocktavious:
There appears to be a race condition if there is a pyenv installation on a build machine that runs multiple processes which create virtualenv's via the command :
pyenv virtualenv _name of env_

When there are two concurrently running processes that try to create a virtualenv for each process there is a race condition when the second process runs the command to create a new virtualenv it eventually makes a call to rehash and errors because of:
pyenv: cannot rehash: /var/lib/jenkins/.pyenv/shims/.pyenv-shim exists

I fully understand if this is a limitation of the way pyenv shims work, but i'm wondering if there is a workaround for this.  Since shim's are path based and each process can have its own environment variables, it would be nice if each process could specify a location for its set of shims so multiple processes on the same machine using pyenv won't collide when the shims are rehashed.

I would think something much like the PYENV_ROOT env var.   It could be PYENV_SHIM_ROOT env var that tells pyenv where to place the shims and if not specified it could default to the standard location.

Let me know if there is anyway i can help make this a reality.

columbusmonkey:
Where is the issue tracker for the jenkins-pyenv-plugin? (As it looks like it may remove the lock directory, even if it did not manage to create it.)

PR #1092, opened 2018-02-08
Make pyenv-rehash safer for multiple processes

clintharrison:
We encountered the same error (in the same parallelizing-build-system setting) as #174, as the hook around `pip install` occasionally ran at the same time. Since `rehash` is a relatively quick operation, and `pip install` operations can take a while, it isn't feasible to force our build system to serialize those steps.

Rather than simply emitting an error message and failing, it'd be nicer if users can opt in to using a file lock when running `pyenv-rehash`. This commit adds the `PYENV_REHASH_PARALLEL_SAFE` environment variable that can be used to accomplish this goal.

On a Linux system, this works well with `flock`. This unfortunately does not exist on macOS by default, but we can fall back to the included `shlock` and achieve reasonable success.

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -12,62 +12,47 @@ LOCK_FILE_FD=90
 # Create the shims directory if it doesn't already exist.
 mkdir -p "$SHIM_PATH"
 
-wait_for_lock() {
-  # flock(1) exists on Linux with util-linux and is the most reliable
-  if command -v flock &>/dev/null; then
-    eval "exec $LOCK_FILE_FD>\"$LOCK_FILE_PATH\""
-    flock --exclusive $LOCK_FILE_FD
-  # Fall back to shlock on BSD systems
-  elif command -v shlock &>/dev/null; then
-    # Try forever to acquire the lock
-    while ! shlock -f "$LOCK_FILE_PATH" -p $$; do
-      sleep 0.5
-    done
-  else
-    echo 'pyenv: cannot rehash: $PYENV_REHASH_PARALLEL_SAFE is set, but this system does not have flock or shlock.' >&2
+# Ensure only one instance of pyenv-rehash is running at a time.
+if command -v flock &>/dev/null; then
+  # flock(1) exists on Linux and is the most reliable
+  eval "exec $LOCK_FILE_FD>\"$LOCK_FILE_PATH\""
+  flock --exclusive $LOCK_FILE_FD
+elif command -v shlock &>/dev/null; then
+  # On macOS/BSD, use shlock when flock is not installed
+  # Try forever to acquire the lock
+  while ! shlock -f "$LOCK_FILE_PATH" -p $$; do
+    sleep 0.5
+  done
+else
+  # This is the fallback behavior when no flock/shlock exists, by
+  # setting the shell's `noclobber` option and attempting to write to
+  # the prototype shim file. If the file already exists, print a warning
+  # to stderr and exit with a non-zero status.
+  set -o noclobber
+  { echo > "$PROTOTYPE_SHIM_PATH"
+  } 2>/dev/null ||
+  { if [ -w "$SHIM_PATH" ]; then
+      echo "pyenv: cannot rehash: $PROTOTYPE_SHIM_PATH exists"
+    else
+      echo "pyenv: cannot rehash: $SHIM_PATH isn't writable"
+    fi
     exit 1
-  fi
-}
+  } >&2
+  set +o noclobber
+fi
+
+# If we were able to obtain a lock, register a trap to clean up the
+# prototype shim and release a possible lock when the process exits.
+trap release_lock EXIT
 
 release_lock() {
+  # Even if flock/shlock were used, the prototype shim should be removed.
+  rm -f "$PROTOTYPE_SHIM_PATH"
+
   if command -v flock &>/dev/null; then
     flock --unlock $LOCK_FILE_FD
   elif command -v shlock &>/dev/null; then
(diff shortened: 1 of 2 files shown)


yyuu:
There's some discussion on this on Gitter

https://gitter.im/yyuu/pyenv?at=5a8b72be53c1dbb7434ef994

libexec/pyenv-rehash
```diff
@@ -6,10 +6,44 @@ set -e
...
+  elif command -v shlock &>/dev/null; then
+    rm -f $LOCK_FILE_PATH
+  fi
+}
+
+# When the PYENV_REHASH_PARALLEL_SAFE variable is set and flock(1) or shlock(1)
+# is available, this script can acquire an exclusive file lock to ensure only
+# one process is running pyenv-rehash.
+if [ -n "$PYENV_REHASH_PARALLEL_SAFE" ]; then
+  wait_for_lock
```
yyuu: There's some code for (pseudo) multi-process locking in `pyenv-rehash`. In general, it'd be better to avoid having code for similar purpose in multiple places. 

https://github.com/pyenv/pyenv/blob/a8e207f330509b12724454b1dd38dcc31193212f/libexec/pyenv-rehash#L13-L27

Please modify the code around the `noclobber` to,

1. Use `flock` or `shlock` if it is available on the platform (otherwise, fallback to `noclobber`)
1. Wait until the release of lock instead of immediate exit as error

For now it's immediately existing as error if there's another `pyenv-rehash` process is running. However, it must be some understandable behavior to wait until the release of lock since there're some of internal calls of `pyenv-rehash` within pyenv and they're actually NOT expecting `pyenv-rehash` to fail.

libexec/pyenv-rehash
```diff
@@ -32,6 +66,9 @@ trap remove_prototype_shim EXIT
 
 remove_prototype_shim() {
   rm -f "$PROTOTYPE_SHIM_PATH"
+  if [ -n "$PYENV_REHASH_PARALLEL_SAFE" ]; then
+    release_lock
```
yyuu: The release of lock need to be done as a `EXIT` trap (e.g. `trap release_lock EXIT`) to make sure the lock is released at the end of the script.

clintharrison: This function is already registered as an exit trap, but I've renamed it so hopefully its purpose is clearer.

libexec/pyenv-rehash
clintharrison: Alright, I removed the environment variable to opt in to this behavior, and use `flock` or `shlock` when available. Unfortunately, without `flock` or `shlock`, there's no way to wait for the lock becoming available when we fall back to the `noclobber` variant. This limitation is fundamentally the reason for adding `flock` and `shlock` support.

clintharrison:
@yyuu any updates here? 🙂

clintharrison:
Ping :)

libexec/pyenv-rehash
```diff
@@ -6,32 +6,58 @@ set -e
...
+# Ensure only one instance of pyenv-rehash is running at a time.
+if command -v flock &>/dev/null; then
+  # flock(1) exists on Linux and is the most reliable
+  eval "exec $LOCK_FILE_FD>\"$LOCK_FILE_PATH\""
+  flock --exclusive $LOCK_FILE_FD
+elif command -v shlock &>/dev/null; then
+  # On macOS/BSD, use shlock when flock is not installed
+  # Try forever to acquire the lock
+  while ! shlock -f "$LOCK_FILE_PATH" -p $$; do
+    sleep 0.5
```
yyuu: POSIX's sleep(1) doesn't support subsecond. It'd be better to regard 1 second as the minimum sleep duration.

https://www.unix.com/man-page/posix/1P/sleep/

libexec/pyenv-rehash
```diff
@@ -6,32 +6,58 @@ set -e
 
 SHIM_PATH="${PYENV_ROOT}/shims"
 PROTOTYPE_SHIM_PATH="${SHIM_PATH}/.pyenv-shim"
+LOCK_FILE_PATH="${SHIM_PATH}/.pyenv.lock"
+LOCK_FILE_FD=90
```
yyuu: At least I don't think it's good idea to hardcode `fd` of lock file in the script....

libexec/pyenv-rehash
```diff
@@ -6,32 +6,58 @@ set -e
...
+
+# Ensure only one instance of pyenv-rehash is running at a time.
+if command -v flock &>/dev/null; then
+  # flock(1) exists on Linux and is the most reliable
+  eval "exec $LOCK_FILE_FD>\"$LOCK_FILE_PATH\""
+  flock --exclusive $LOCK_FILE_FD
+elif command -v shlock &>/dev/null; then
+  # On macOS/BSD, use shlock when flock is not installed
+  # Try forever to acquire the lock
+  while ! shlock -f "$LOCK_FILE_PATH" -p $$; do
```
yyuu: In my understanding, this lock relies on the file at `$LOCK_FILE_PATH` and there might be possibility of unwanted unlock by deletion of the file at the path. I think it's effectively similar as current pseudo lock mechanism based on `noclobber`.

test/rehash.bats
```diff
@@ -25,6 +25,10 @@ create_executable() {
 }
 
 @test "rehash in progress" {
+  # Skip this when flock/shlock are available
+  if [ -n $(command -v flock) ] || [ -n $(command -v shlock) ]; then
+    skip
```
yyuu: This effectively skips rehash test on Travis CI, etc. It's not good at least.

PR #1094, opened 2018-02-09
Correct CPython 3.4.8 Checksums

claco:
The current checksums for tar.xz and .tgz files are reversed. This
causes python-build to fail with a checksum mismatch error.

Reversing the checksums to allow the build to succeed.


claco:
I'm not terribly familiar with the test suite. Seems like something we could test for. Any hints on how or where in ./tests this might be implemented?

PR #1096, opened 2018-02-10
Feature Request: add a "build" hook with associated support functions

cornfeedhobo:
This PR seeks to add a new `build` hook, allowing for plugins to modify the build tree before compilation, but after extraction.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -16,7 +16,25 @@
 
 PYTHON_BUILD_VERSION="20160602"
 
+# Define `before_build` and `after_build` functions that allow
+# plugin hooks to register a string of code for execution before or
+# after the build process.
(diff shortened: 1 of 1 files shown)


cornfeedhobo:
Could someone take a look at this finally? Python installs break on various distros and this would allow for easy patching during builds. Please.

cornfeedhobo:
Is there any hope that this idea could get some attention?

native-api:
First, make sure that tests pass if you hope for any attention.

Then, for a new feature request, it's advisable to create an issue where you make a case for your feature: present the intended use cases, compare the proposed solution against other possible ones... Then link the PR to it as a proposed implementation.

It's very possible that there are other possible and perhaps better implementations, so things like this are inconvenient to discuss in a PR.

PR #1110, opened 2018-03-06
Added a check for PATH if already set

zafartahirov:
There is a minor problem when stack-running shell -- the PATH variable keeps on appending the `.../.pyenv/shims:...`. You can reproduce it by running

```
bash
bash
bash
bash
echo ${PATH}
# On my machine prints
# /home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:/home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:/home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:/home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:/home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:/home/zafar/.pyenv/shims:/home/zafar/.pyenv/bin:...
```

This PR adds a safety check if the appended path already exists in the `${PATH}`. Note that the check is if the appended path is between `(^|:)` and `($|:)` -- just in case there is a need to add sub-paths (i.e.: it is fine to add `/some/long/path` and `/some/long/` after it).

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -2,6 +2,18 @@
 set -e
 unset CDPATH
 
+# Safely adds new paths to the PATH variable.
+path_add_front() {
+  if ! echo ${PATH} | egrep -q "(^|:)$1($|:)" ; then
+    PATH=${1}:${PATH}
+  fi
+}
+path_add_back() {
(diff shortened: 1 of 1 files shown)


zafartahirov:
Still buggy :( Closing

PR #1117, opened 2018-03-13
Homebrew Set openssl PKG_CONFIG_PATH for python 3.7

jenshnielsen:
As noted in https://github.com/pyenv/pyenv/issues/993 Python 3.7 changes the way that non default location openssl is added to configure. See https://bugs.python.org/issue32598

This pr aims to make python 3.7b2 build with openssl on High Sierra again by adding openssl to PKG_CONFIG_PATH


joshfriend:
I've built 3.7.0b2 and 3.7-dev probably 5 times now and no matter what I try, I don't get the "ssl extension was not compiled" errors. However, your change doesn't break the builds and it makes sense to do this so 👍

PR #1120, opened 2018-03-15
Fix #1118: follow only dir-symlinks

### LINKED ISSUE #1118: cd: <symlink target file>: Not a directory
File `pyenv` assumes, that it can follow symlinks in `abs_dirname()`:
```bash
  abs_dirname() {
    local path="$1"

    # Use a subshell to avoid changing the current path
    (
    while [ -n "$path" ]; do
      cd "${path%/*}"
      local name="${path##*/}"
      path="$(resolve_link "$name" || true)"
    done

    pwd
    )
  }
```

However, the target file can be a symlink, and it cannot be followed.

Example from Ansible. Listing of `ansible/bin/`:
```
    ansible
    ansible-playbook -> ansible
```

Executing `~/ansible/bin/ansible-playbook` will work, but during it a `~/.pyenv/libexec/pyenv: line 44: cd: ansible: Not a directory` is emitted.

A suggested fix would be:
```bash
      path="$(test -d "$name" && resolve_link "$name" || true)"
```
... to test if the destination is a directory that can be `cd`d into.
If yes, follow that. If not, stop iterating without emitting errors.

webknjaz:
This corrects the deffect where pyenv tries to determine an absolute
path of the directory containing a file and follows symlinks
manually. Because of missing check it behaved correctly, but a
misleading error message was shown in the terminal output.
Resolves #1118.

Thanks to @HQJaTu for the initial patch.

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -35,22 +35,18 @@ else
     $READLINK "$1"
   }
 
-  resolve_link_follow() {
-    $READLINK -f "$1"
-  }
-
   abs_dirname() {
     local path="$1"
 
     # Use a subshell to avoid changing the current path
     (
-    while [ -n "$path" ]; do
+    while [[ -n "$path" && $(echo "$path" | grep /) ]]; do
       cd "${path%/*}"
       local name="${path##*/}"
       path="$(resolve_link "$name" || true)"
     done
 
-    resolve_link_follow "$(pwd)"
+    pwd
     )
   }
 fi


webknjaz:
@joshfriend @yyuu @timsavage what is the policy for bugfix releases? can we have this out asap plz?

webknjaz:
Ping @joshfriend @yyuu @timsavage

timsavage:
Please stop referencing me, I'm not part of the pyenv team.

webknjaz:
Oops, sorry. I just saw your commits in master and missed they they're coming from PR :)

libexec/pyenv
```diff
@@ -40,7 +40,7 @@ else
 
     # Use a subshell to avoid changing the current path
     (
-    while [ -n "$path" ]; do
+    while [[ -n "$path" && $(echo "$path" | grep /) ]]; do
```
yyuu: This will return "pwd" if "$path" is /. At least it must not be the correct return value of the function.

yyuu: I think some fix like following will work sufficiently. Please consider adding tests for it.

'''
while [[ "$path" == *"/"* ]]; do
  cd "${path%/*}"
  local name="${path##*/}"
  [ -n "$name" ] || break
  path="$(resolve_link "$name" || true)"
done
'''

webknjaz: Thanks! I'll check this

webknjaz: Just checked it: it doesn't work your way at all. Will need to investigate some better way.

blueyed:
I've started a while back to fix this in rbenv (upstream), but it needs updating: https://github.com/rbenv/rbenv/pull/868.

blueyed:
btw: `src/configure && make -C src` in pyenv's checkout should fix / work around this, and improve performance as a side-effect.

webknjaz:
Wow, why?

*/me needs to get back to fixing this*

PR #1121, opened 2018-03-15
Fix dead link in ToC

raimon49:
I found dead link inside ToC of this project.

* https://github.com/pyenv/pyenv#version-history
* https://github.com/pyenv/pyenv#license

Both are considered lost when merging with the [rbenv's README](https://github.com/rbenv/rbenv/blob/master/README.md).

I fixed that both were useful information and that I should leave the link.


joshfriend:
nice, thank you!

Issue #1127, opened 2018-03-21
pyenv should use version-specific `get-pip.py` when installing 2.6 and 3.2


di:
The recent release of `pip` 9.0.2 unintentionally broke support for Python 2.6, and the related update to https://bootstrap.pypa.io/get-pip.py to use this version broke attempts to `pyenv install 2.6.x` due to this incompatibility.

Instead of using https://bootstrap.pypa.io/get-pip.py, `pyenv` should be using the version-specific installers at https://bootstrap.pypa.io/2.6/get-pip.py and https://bootstrap.pypa.io/3.2/get-pip.py

These won't actually fix the issue until a `pip` 9.0.3 release is made to fix Python 2.6 compatibility, but they should continue to work into the future, especially when `pip` 10.0.0 is released, at which point https://bootstrap.pypa.io/get-pip.py will definitely no longer support Python 2.6.

Happy to make a PR for this if it's deemed valid.

joshfriend:
Now that pypa/get-pip#17 seems to be resolved I think someone could actually start working on fixing this.

di:
The 2.6 version is not actually compatible with 2.6 and won't be until a 9.0.3 release of pip is made, see https://github.com/pypa/pip/issues/5093

yyuu:
As a quick workaround , declaring `GET_PIP_URL` on calling `pyenv install` will let `python-build` to use `get-pip.py` at the given URL. As long as the URL is returning correct response, it should work.

https://github.com/pyenv/pyenv/blob/v1.2.2/plugins/python-build/bin/python-build#L1807

For intrinsic fix, for non-CPython distributions, it's actually not easy to know if the Python compatibility version is 2.6/3.2 or not, prior to the build/install. I think it's nearly safe to assume to regard most of those non-CPython distributions as kind of edge cases, though.

Anyway I think that changing `get-pip` URL for older Pythons is a kind of tentative workaround and will not live longer. I think that pinning `pip` version within the build definition of 2.6/3.2 must be the right way to deal with the situation. We're doing similar for some ancient versions like 2.5.x.

https://github.com/pyenv/pyenv/blob/v1.2.2/plugins/python-build/share/python-build/2.5.6#L4-L5

hokuto-k:
@di 
Thank you for the submitting the issue.
just a hour ago pip 9.0.3 was released.

hokuto-k:
@yyuu 
[...]

PR #1131, opened 2018-03-29
Use custom get-pip URL based on the target version

yyuu:

### Prerequisite
* [x] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from this project.
  * We are occasionally importing the changes from rbenv. In general, you can expect some changes made in rbenv will be imported to pyenv too, eventually.
  * Generaly speaking, we sometimes don't prefer to make some change in the core to keep compatibility with rbenv.
  * This change is pretty much Python specific.
  - https://github.com/pyenv/pyenv/issues/1127

### Description

Let python-build to use custom URL for `get-pip.py` some specific versions of CPython as a tentative workaround for backward compatibility issue of pip.

### Tests

Added some tests of `GET_PIP_URL` value for 2.6, 3.2, 3.3 and other versions.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -2122,7 +2122,21 @@ if [ -z "${GET_PIP_URL}" ]; then
     # Unset `PIP_VERSION` from environment before invoking `get-pip.py` to deal with "ValueError: invalid truth value" (pypa/pip#4528)
     unset PIP_VERSION
   else
-    GET_PIP_URL="https://bootstrap.pypa.io/get-pip.py"
+    # Use custom get-pip URL based on the target version (#1127)
+    case "${DEFINITION_PATH##*/}" in
+    2.6 | 2.6.* )
+      GET_PIP_URL="https://bootstrap.pypa.io/2.6/get-pip.py"
+      ;;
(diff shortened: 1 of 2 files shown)


plugins/python-build/bin/python-build
```diff
@@ -2118,11 +2118,28 @@ if [ -z "${EZ_SETUP_URL}" ]; then
 fi
 if [ -z "${GET_PIP_URL}" ]; then
   if [ -n "${PIP_VERSION}" ]; then
+    { colorize 1 "WARNING"
+      echo ": pip is going to abandon the support for the custom version installation via custom get-pip URL. The use of PIP_VERSION=${PIP_VERSION} may not work anymore."
```
yyuu: Honestly speaking, I don't have correct words for explanation of this warning.... 😥  Could someone have any idea on this?

joshfriend: How does this sound?

```sh
Setting PIP_VERSION=${PIP_VERSION} is no longer supported and may cause failures during the install process.
```

joshfriend:
This fix worked out just fine when I tested by installing 2.6.9 :+1:

PR #1137, opened 2018-04-13
Run rehash command in the background

andersonvom:
The rehash command doesn't take too long to run, but can add up when
several *env projects (rbenv, pyenv, nodenv, etc) are installed at the
same time. This makes it run in the background so that you don't have
to skip it altogether (using the `--no-rehash` flag) but also don't
have to wait a lot during shell initialization.

The `()` around the command prevent it from displaying the job number
or the "job done" message once it finishes.

### Prerequisite
* [ ] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from this project.
  * We are occasionally importing the changes from rbenv. In general, you can expect some changes made in rbenv will be imported to pyenv too, eventually.
  * Generally speaking, we sometimes don't prefer to make some change in the core to keep compatibility with rbenv.

### Description

### Tests

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -101,7 +101,7 @@ if [ -r "$completion" ]; then
 fi
 
 if [ -z "$no_rehash" ]; then
-  echo 'command pyenv rehash 2>/dev/null'
+  echo '(command pyenv rehash 2>/dev/null &)'
 fi
 
 commands=(`pyenv-commands --sh`)
--- a/test/init.bats
+++ b/test/init.bats
@@ -14,7 +14,7 @@ load test_helper
 @test "auto rehash" {
   run pyenv-init -
   assert_success
-  assert_line "command pyenv rehash 2>/dev/null"
+  assert_line "(command pyenv rehash 2>/dev/null &)"
 }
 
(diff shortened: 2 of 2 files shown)


yyuu:
I think running rehash in background will lead to non-deterministic result since there might be multiple callers of rehash in the wild. It must become the source of race conditions.

If you really want to avoid running rehash in foreground, it'd recommend you to have some single background process that runs rehash periodically, then add `--no-rehash` to avoid calling rehash from foreground process.

andersonvom:
If `rehash` can't be run multiple times simultaneously, should this be fixed in the command itself? Maybe create a lock file when it's running to prevent the situation your described.

bdellegrazie:
@andersonvom I believe there's still a race condition, see #1092

andersonvom:
If I understand the issue correctly there's no race condition. If there's already a `rehash` running, the subsequent runs will just issue an error message and quit, without causing any issues to pyenv. Running it in the foreground/background should have no effect on multiple instances of `rehash` (e.g. even if it runs in the foreground, users could spawn multiple shell sessions and still have `rehash` run at the same time.)

PR #1153, opened 2018-05-01
Remove shebang lines from scripts for performance

cmcginty:
All scripts in libexec/ (excluding pyenv) are called through pyenv,
therefore the shebang lines are not necessary. On some systems this
provides a measurable increase in performance of the shell prompt.

Related to pyenv/pyenv-virtualenv#259

On my system this removes 1050ms of runtime from `pyenv sh-activate --quiet`.

### DIFF
--- a/test/init.bats
+++ b/test/init.bats
@@ -35,7 +35,7 @@ load test_helper
   cd "$PYENV_TEST_DIR"
   cat > myscript.sh <<OUT
 #!/bin/sh
-eval "\$(pyenv-init -)"
+eval "\$(pyenv init -)"
 echo \$PYENV_SHELL
 OUT
   chmod +x myscript.sh
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 set -e
 
 if [ "$1" = "--debug" ]; then
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 set -e
 
 if [ "$1" = "--debug" ]; then
--- a/libexec/pyenv---version
+++ b/libexec/pyenv---version
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: Display the version of pyenv
 #
 # Displays the version number of this pyenv release, including the
--- a/libexec/pyenv-commands
+++ b/libexec/pyenv-commands
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: List all available pyenv commands
 # Usage: pyenv commands [--sh|--no-sh]
 
--- a/libexec/pyenv-completions
+++ b/libexec/pyenv-completions
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Usage: pyenv completions <command> [arg1 arg2...]
 
 set -e
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-#
 # Summary: Run an executable with the selected Python version
 #
 # Usage: pyenv exec <command> [arg1 arg2...]
--- a/libexec/pyenv-global
+++ b/libexec/pyenv-global
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-#
 # Summary: Set or show the global Python version
 #
 # Usage: pyenv global <version>
--- a/libexec/pyenv-help
+++ b/libexec/pyenv-help
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-#
 # Summary: Display help for a command
 #
 # Usage: pyenv help [--usage] COMMAND
--- a/libexec/pyenv-hooks
+++ b/libexec/pyenv-hooks
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: List hook scripts for a given pyenv command
 # Usage: pyenv hooks <command>
 
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: Configure the shell environment for pyenv
 # Usage: eval "$(pyenv init - [--no-rehash] [<shell>])"
 
--- a/libexec/pyenv-local
+++ b/libexec/pyenv-local
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-#
 # Summary: Set or show the local application-specific Python version
 #
 # Usage: pyenv local <version>
--- a/libexec/pyenv-prefix
+++ b/libexec/pyenv-prefix
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: Display prefix for a Python version
 # Usage: pyenv prefix [<version>]
 #
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Summary: Rehash pyenv shims (run this after installing executables)
 
 set -e
@@ -62,7 +61,6 @@ fi
 # serves as a locking mechanism.
 create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
-#!/usr/bin/env bash
 set -e
 [ -n "\$PYENV_DEBUG" ] && set -x
 
--- a/libexec/pyenv-root
+++ b/libexec/pyenv-root
@@ -1,3 +1,2 @@
-#!/usr/bin/env bash
 # Summary: Display the root directory where versions and shims are kept
 echo "$PYENV_ROOT"
--- a/libexec/pyenv-sh-rehash
+++ b/libexec/pyenv-sh-rehash
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 set -e
 [ -n "$PYENV_DEBUG" ] && set -x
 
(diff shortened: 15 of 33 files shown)


cmcginty:
Just to clarify this change essentially changes the sub-calls from `$(SCRIPT)` to `$(source SCRIPT)`. In the later case, removing the shebang line as no effect on performance.

It might be clearer to change all the internal calls to `$(source SCRIPT)` but this is a much larger change to implement for the same effect.

mhchia:
Thank you for fixing it. I met the same issue as well.
One stupid question is, I don't get why
> removing the shebang lines essentially change the bash sub-calls from $(SCRIPT) to $(source SCRIPT)

I tried remove a shebang from a script `test.sh`
```
echo $$
```
do
```
chmod +x test.sh
./test.sh
```
and get different output every time I execute it. It seems still invoking a bash process to execute it.
Am I understanding anything wrong?

test/init.bats
```diff
@@ -35,7 +35,7 @@ load test_helper
   cd "$PYENV_TEST_DIR"
   cat > myscript.sh <<OUT
 #!/bin/sh
-eval "\$(pyenv-init -)"
+eval "\$(pyenv init -)"
```
blueyed: This is not really necessary, but only changed for consistency I assume?

blueyed:
Isn't this missing the actual change to use `source` instead then, too?
Or are you saying that you get a performance boost because `/bin/sh` is used?

I've tested it using `t-source.sh`:
```sh
#!/usr/bin/bash

echo 1:$0 $$ $BASHPID

PATH=$PWD/t-source

echo 1
t.sh
echo 2
echo $(t.sh)
echo 3
source t.sh
```

And `t-source/t.sh`:
```sh
echo sub:$0 $$ $BASHPID
```

This outputs:
```
1:t-source.sh 20333 20333
1
sub:/home/user/.pyenv/t/t-source/t.sh 20334 20334
2
sub:/home/user/.pyenv/t/t-source/t.sh 20335 20335
3
sub:t-source.sh 20333 20333
```

I.e. the same PID is only used when using `source t.sh`.

I've done something in this regard also a while ago (https://github.com/pyenv/pyenv/pull/357), it was closed back then since anyenv appeared to be a replacement soon, but that is not the case yet still.

test/init.bats
cmcginty: correct

cmcginty:
> Isn't this missing the actual change to use source instead then, too?
When I was testing both methods resolved the perf issues, but seeing your test it looks like the methods are not equivalent.
> Or are you saying that you get a performance boost because /bin/sh is used?
no, for some reason remove the shebang resolve the issues. Now that I know it is related to Carbon Black, that might help explain the edge case. It's not that creating the sub-process is slow, but how Carbon Black hooks into the sub-system. 
> I.e. the same PID is only used when using source t.sh.
Yes, so this seems like the proper solution then. However it's a larger change and I wasn't sure if this PR was going to be accepted at the time and I was also trying alternatively resolve with a Carbon Black change / fix.

blueyed:
What is Carbon Black?

cmcginty:
carbonblack.com Enterprise virus/malware scanner.

AndreasBackx:
I'm not using Carbon Black and am experiencing this slowdown on my late 2013 Retina Macbook Pro but am not experiencing it on my desktop machine running Linux and an AMD Ryzen 1700. That might either be because of the more powerful CPU or something else is different. Regardless, I don't think Carbon Black has got something to do with this.

Regardless though, is it possible that this could get merged? If not, what is holding the merge back?

blueyed:
@AndreasBackx 
Feel free to use this patch locally already, of course.

But otherwise it is just not clear why that is, and it does not make sense to merge something just-because I guess?!

You could compare the output from with and without `PYENV_DEBUG=1` exported.

blueyed:
@AndreasBackx 
Have you even tried this PR/patch, or are you just seeing a slowdown in general?

AndreasBackx:
I have recorded a video of the slowdown with a comparison between:

- pyenv and pyenv-virtualenv enabled **without** this patch
- pyenv and pyenv-virtualenv enabled **with** this patch
- pyenv and pyenv-virtualenv disabled

I'm currently on the train with mobile data. I'll upload it to YouTube and post it here when I can.

I used a video instead of timing functions as I could not measure with any combination of commands that I tried. If someone can find a way to measure it, please leave a comment. I've tried timing the following commands:

- `zsh -i -c exit`
- `echo test` (this basically just runs instantly, it doesn't seem to be affected at all. Though maybe it's because Python was using sh instead of zsh. I'll have to confirm this another time though as I'm on the train atm)

The video shows the difference very clearly between enabling pyenv and disabling it. Though I don't see a clear difference with this pull request unfortunately.

blueyed:
> The video shows the difference very clearly between enabling pyenv and disabling it.

We believe you - a video would not help there/here.

> Though I don't see a clear difference with this pull request unfortunately.

So it is off topic for this PR - please create a new issue / search a more fitting one.

FWIW: it is not really necessary to use pyenv-virtualenv.
And you can also mainly use pyenv just for installation of versions - you can just use e.g. `~/.pyenv/versions/3.7.2/bin/python -m venv .venv` to create a virtualenv for a project, and then just ensure that this virtualenv gets activated, e.g. through `chpwd` hooks.
Then you can leave `pyenv init` off by default.

AndreasBackx:
> So it is off topic for this PR - please create a new issue / search a more fitting one.

I would disagree. There are no performance numbers for this PR to compare it with master. You cannot seem to reproduce it from what I can tell, so it might be best to first objectively measure whether this is an improvement for those affected rather than writing it off immediately or creating multiple issues for performance issues.

blueyed:
You said yourself:

> Though I don't see a clear difference with this pull request unfortunately.

You should at least have some clear description of your issue, and a way to reproduce/measure it.
Therefore you should take pyenv-virtualenv out of the equation in the first place.
It also heavily depends on how you setup pyenv in your shell, and if you e.g. have something slow in bash's interactive rc files, which would just make any bash script slower.

I have a useful tiny program to run some shell command via Python's timeit module, which is quite useful for profiling: https://github.com/blueyed/dotfiles/blob/master/usr/bin/timeit-shell.

As for this PR: this is only showing that your interactive bash startup is slow - and that some external programs might make it worse.

I've just confirmed though that having a shebang indeed causes some overhead, i.e. a new PID - I've expected bash to be smarter about that.

Using these files:
```
bash:#!/bin/sh
bash:echo "SLEEP"
bash:/usr/bin/sleep 1
bash:/usr/bin/bash "$@"
test.bash:#!/usr/bin/bash
test.bash:set -x
test.bash:PATH=$PWD
test.bash:t.sh
test.bash:t-with-shebang.sh
test.bash:echo "$(t-with-shebang.sh)"
t.sh:echo t.sh
t.sh:echo PID=$$ BASHPID=$BASHPID SHELL=$SHELL
t-with-shebang.sh:#!bash
t-with-shebang.sh:echo t-with-shebang.sh
t-with-shebang.sh:echo PID=$$ BASHPID=$BASHPID SHELL=$SHELL
```

`bash test.sh` shows:
```
+ PATH=/home/daniel/.pyenv/t-source
+ t.sh
t.sh
PID=14290 BASHPID=14290 SHELL=/usr/bin/zsh
+ t-with-shebang.sh
SLEEP
t-with-shebang.sh
PID=14293 BASHPID=14293 SHELL=/usr/bin/zsh
++ t-with-shebang.sh
+ echo 'SLEEP
t-with-shebang.sh
PID=14296 BASHPID=14296 SHELL=/usr/bin/zsh'
SLEEP
t-with-shebang.sh
PID=14296 BASHPID=14296 SHELL=/usr/bin/zsh
```

I.e. it calls "bash" when used in a shebang.

Therefore I think this PR makes sense after all, but should go through rbenv then also - in cas @mislav would take it.

blueyed:
Looking at strace for `./t.sh` from bash shows:
```
28610 execve("./t.sh", ["./t.sh"], 0x5585d6c1ee60 /* 81 vars */) = -1 ENOEXEC (Exec format error)
28610 openat(AT_FDCWD, "./t.sh", O_RDONLY) = 3
28610 read(3, "echo t.sh\necho PID=$$ BASHPID=$B"..., 128) = 52
28610 close(3)                          = 0
28610 pipe([3, 4])                      = 0
28610 close(3)                          = 0
28610 close(4)                          = 0
```
I.e. `execve` fails, and then it just reads the file itself.  It is also `stat`ing around for "bash", finding "/usr/bin/bash", and maybe checking if it is itself?!  - but not spawning a new process.

Using `./t-with-shebang.sh` will use `clone` (but does appear to use extra files / much additional processing/reading for me).

I have used `strace -f -o O bash`, and then `tail -f O` from another terminal why calling those scripts.

blueyed:
I've rebased this PR for now, and added a new commit to also remove it from `libexec/pyenv`.

blueyed:
For me this improves `timeit-shell ~/.pyenv/bin/pyenv init -` from 47-48ms to 40-42ms.
`pyenv versions` from ~80ms to 58ms.
`pyenv whence python` from ~100ms to ~50ms.
`pyenv exec python -V` from ~70ms to ~50ms

So it is quite good in general!

blueyed:
This appears to be quite similar to https://github.com/pyenv/pyenv/pull/357 - where I was using `source`, but has a far more easier diff/patch.

chrahunt:
LGTM. The mechanism behind this is clear - `/usr/bin/env` is no longer executed and doesn't have to search for `bash` in `PATH` for every command substitution. `bash` by default assumes a file without she-bang has shell commands and just runs it in a sub-shell [here](https://github.com/bminor/bash/blob/master/execute_cmd.c#L5765). This can save dozens or hundreds of filesystem calls.

PR #1168, opened 2018-06-04
tests: remove "supports python -S <cmd>"

blueyed:
This was imported semi-automatically in 0965577, but `ruby -S` is
different from `python -S`.

`ruby -S`: look for the script using PATH environment variable
`python -S`: don't imply 'import site' on initialization

### DIFF
--- a/test/exec.bats
+++ b/test/exec.bats
@@ -77,33 +77,3 @@ ${PYENV_ROOT}/versions/3.4/bin/python
   args
 OUT
 }
-
-@test "supports python -S <cmd>" {
-  export PYENV_VERSION="3.4"
-
-  # emulate `python -S' behavior
-  create_executable "python" <<SH
-#!$BASH
-if [[ \$1 == "-S"* ]]; then
-  found="\$(PATH="\${PYTHONPATH:-\$PATH}" which \$2)"
-  # assert that the found executable has python for shebang
-  if head -1 "\$found" | grep python >/dev/null; then
-    \$BASH "\$found"
-  else
-    echo "python: no Python script found in input (LoadError)" >&2
-    exit 1
-  fi
-else
-  echo 'python 3.4 (pyenv test)'
-fi
-SH
-
-  create_executable "fab" <<SH
-#!/usr/bin/env python
-echo hello fab
-SH
-
-  pyenv-rehash
-  run python -S fab
-  assert_success "hello fab"
-}


test/exec.bats
```diff
@@ -77,33 +77,3 @@ ${PYENV_ROOT}/versions/3.4/bin/python
   args
 OUT
 }
-
-@test "supports python -S <cmd>" {
```
yyuu: Wouldn't it be better to keep the `@test` for this test and comment out the code instead of deletion...? I'm afraid that the deletion will delete the history as well and it'll be easy to make the merge mistake again 🤔

chrahunt: There's probably a simpler way, but the issue with future merge can be avoided by doing the following:

```
git clone git@github.com:pyenv/pyenv.git
git clone git@github.com:rbenv/rbenv.git
find rbenv/test -type f | xargs sed -i -e s/ruby/python/g -e s/rbenv/pyenv/g ... --
cd pyenv
git checkout -b update-tests 0965577
cp -r ../rbenv/test .
git commit -m "update-tests"
git checkout -b really-update-tests master
git merge update-tests
```

When merging against this PR, this results in:

```
Auto-merging test/which.bats
...
Auto-merging test/exec.bats
CONFLICT (content): Merge conflict in test/exec.bats
Auto-merging test/--version.bats
CONFLICT (content): Merge conflict in test/--version.bats
Automatic merge failed; fix conflicts and then commit the result.
```

And then in `test/exec.bats` there is

```
<<<<<<< HEAD
=======

@test "supports python -S <cmd>" {
  export PYENV_VERSION="2.0"

  # emulate `python -S' behavior
  create_executable "python" <<SH
#!$BASH
if [[ \$1 == "-S"* ]]; then
  found="\$(PATH="\${RUBYPATH:-\$PATH}" which \$2)"
  # assert that the found executable has python for shebang
  if head -1 "\$found" | grep python >/dev/null; then
    \$BASH "\$found"
  else
    echo "python: no Python script found in input (LoadError)" >&2
    exit 1
  fi
else
  echo 'python 2.0 (pyenv test)'
fi
SH

  create_executable "rake" <<SH
#!/usr/bin/env python
echo hello rake
SH

  pyenv-rehash
  run python -S rake
  assert_success "hello rake"
}
>>>>>>> update-tests
```

as expected, showing that it was removed in master.

So I'm +1 on this PR.

blueyed:
@yyuu 
Are you ok with merging as-is?

The best might be to just disable the test somehow (not commenting it completely), so it would not cause merge conflicts for changes to the test itself, but I guess bats does not support it.

Issue #789, opened 2016-12-26
pyenv-exec should not change PATH variable unless CPython is used.


kota65535:
For the fix of [this issue](https://github.com/yyuu/pyenv/issues/98), `pyenv-exec` becomes to add the system python path (usually `/usr/bin`) to the head of PATH environment variable before running python. But this introduces another problem for me.

## Symptom

IntelliJ IDEA has the function to launch another terminal inside it. I realized that the PATH variable of the IDEA terminal is different from the one of the terminal that is normally launched.
It has `/usr/bin` at the head of the PATH, that means `/usr/bin` is the higher priority than `/usr/local/bin` to search executables.

The cause of this problem is as follows:

1. IntelliJ IDEA launches by python script, so it is also run through `pyenv-exec`.
2. `pyenv-exec` add `/usr/bin` to the head of PATH.
3. IntelliJ IDEA terminal inherits the PATH from IDEA itself.

## Solution

I think there are two solutions. 

1. Change the position of `/usr/bin` inserted to the tail of PATH variable.
2. Do not change PATH variable unless CPython is used.

Which is better solutions? Is there another way to resolve this?

yyuu:
I've never used IntelliJ for Python projects.... Are you running on macOS?

Honestly speaking, it is kinda misconfiguration of shell which was invoked from IntelliJ's terminal. I suspect changing the position of `/usr/bin` in `PATH` conditionally corrupts something. At present I'm not confident enough though. Anyway will try running IntelliJ when I get a chance.

kota65535:
Sorry, I'm not talking about Python project.  **IntelliJ IDEA uses a python script when launching itself**.  
If you install IntelliJ IDEA, you can see `/usr/local/bin/idea` is a python script that launches IntelliJ IDEA.  
So in the environment where pyenv installed, IntelliJ IDEA inevitably suffers the change of PATH by pyenv-exec...

yyuu:
I guess you launched IntelliJ with setting pyenv version as `system`. That must be the reason why `/usr/bin` was at the top of the `PATH`.

> 1. Change the position of /usr/bin inserted to the tail of PATH variable.
> 1. Do not change PATH variable unless CPython is used.

I think both two will reintroduce the issue of #98. If so, I think it's better to keep current code since it's for `virtualenv`. Umm 😕

yyuu:
I guess some of the following can be a workaround for the issue. Please let me know if there's some other ways.

[...]

PR #1169, opened 2018-06-04
pyenv-exec: do not use `exec -a`, do not mangle PATH for system Python

### LINKED ISSUE #98: Fails for/with `virtualenv --python="$(which python)" FOO`

blueyed:
Fixes https://github.com/pyenv/pyenv/issues/789.
Ref: #98.

### DIFF
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -41,7 +41,8 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-# CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
-# https://github.com/pyenv/pyenv/issues/98
-export PATH="${PYENV_BIN_PATH}:${PATH}"
-exec -a "$PYENV_COMMAND" "$PYENV_COMMAND_PATH" "$@"
+if [ "${PYENV_BIN_PATH#${PYENV_ROOT}}" != "${PYENV_BIN_PATH}" ]; then
+  # Only add to $PATH for non-system version.
+  export PATH="${PYENV_BIN_PATH}:${PATH}"
+fi
+exec "$PYENV_COMMAND_PATH" "$@"
--- a/test/exec.bats
+++ b/test/exec.bats
@@ -77,3 +77,39 @@ ${PYENV_ROOT}/versions/3.4/bin/python
   args
 OUT
 }
+
+@test "sys.executable with system version (#98)" {
+  system_python=$(which python)
+
+  PYENV_VERSION="custom"
+  create_executable "python" ""
+  unset PYENV_VERSION
+
+  pyenv-rehash
+  run pyenv-exec python -c 'import sys; print(sys.executable)'
+  assert_success "${system_python}"
+}
+
+@test '$PATH is not modified with system Python' {
+  # Create a wrapper executable that verifies PATH.
+  PYENV_VERSION="custom"
+  create_executable "python" '[[ "$PATH" == "${PYENV_TEST_DIR}/root/versions/custom/bin:"* ]] || { echo "unexpected:$PATH"; exit 2;}'
+  unset PYENV_VERSION
+  pyenv-rehash
+
+  # Path is not modified with system Python.
+  run pyenv-exec python -c 'import os; print(os.getenv("PATH"))'
+  assert_success "$PATH"
+
+  # Path is modified with custom Python.
+  PYENV_VERSION=custom run pyenv-exec python
+  assert_success
+
+  # Path is modified with custom:system Python.
+  PYENV_VERSION=custom:system run pyenv-exec python
+  assert_success
+
+  # Path is not modified with system:custom Python.
+  PYENV_VERSION=system:custom run pyenv-exec python -c 'import os; print(os.getenv("PATH"))'
+  assert_success "$PATH"
+}


blueyed:
The test failure is from a test that should be removed (#1168).

blueyed:
Related rbenv PR, but PATH gets mangled in rbenv for other reasons (`rbenv -S`): https://github.com/rbenv/rbenv/pull/1088.

blueyed:
See https://github.com/rbenv/rbenv/pull/1089 for the 3 separate commits/issues.
Missing here is not to change $PATH with regard to libexec and plugin bins.

Let's wait for feedback on the upstream PRs, and then we can either merge it from there, or just go forward here already (and include the missing change).

This fixes a bunch of issues, so is rather nice to have as soon as possible, of course.

blueyed:
Changed this PR back to only remove `-a` with `exec` and not changing `$PATH`.

As per https://github.com/rbenv/rbenv/pull/1089#issuecomment-394531896 changing the PATH is intentional in rbenv, and could be brought back to pyenv also, but should then also get skipped again for the `system` version.

libexec/pyenv-exec
```diff
@@ -41,7 +41,7 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-# CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
-# https://github.com/pyenv/pyenv/issues/98
-export PATH="${PYENV_BIN_PATH}:${PATH}"
-exec -a "$PYENV_COMMAND" "$PYENV_COMMAND_PATH" "$@"
+if [ "$PYENV_VERSION" != "system" ]; then
```
yyuu: Maybe I'm missing spme context, but how it should behave if there are multiple versions in PYENV_VERSION and the "system" appeared in the middle...?

blueyed: Good point.

It should only be added if the command is from a non-system version after all, i.e. with `PYENV_VERSION=system:3.5.3`, PATH should not be changed for `pip`, which is used from `system` then.
To make this work pyenv-which needs to return the used version in some new mode then maybe?

fishyai:
Hey thanks for your work here. What's the status on this? I ran into the issue again today where '/usr/bin' is prepended to my path for pyenv shims (including the system). It's preventing me from using pyenv :(

Gollum999:
+1, I'm also running into issues caused by the extra `/usr/bin`.

I have been experimenting a bit with the [pyenv-register](https://github.com/doloopwhile/pyenv-register) plugin.  For some reason it doesn't get the extra `/usr/bin` if you register a new copy of the system python, so that might be a workaround while we wait for this fix.

eregon:
A good test for this would be to assert that `$PATH` outside and inside of the system `python` (`python -c 'import os; print(os.getenv("PATH"))'`) is the same. I believe `pyenv system` (or when no python is selected) should not modify PATH, which is also the behavior of all 3 major Ruby managers.

Other cases where there are multiple pythons selected can behave differently. But at least if `pyenv` is loaded without configuration, it shouldn't prepend `/usr/bin` to PATH.

libexec/pyenv-exec
eregon: Could we go ahead with this, which I think fixes the common case of `pyenv system` unfortunately prepending `/usr/bin` to PATH and breaking other software not expecting `/usr/bin` to be so early in PATH?
If not, is there another solution to deal with this?

blueyed:
Ok, revisited it and added more tests - should be ready.

libexec/pyenv-exec
blueyed: I've changed it now to get the used version out of `pyenv-which`.

blueyed:
@eregon 
Please re-review.

blueyed:
@efiop 
> Unless I'm missing something, it looks like PATH is still modified in `pyenv` itself, where `plugin_bin` is added (this one is minor)

You mean https://github.com/pyenv/pyenv/blob/7c944542609bb02180bd1c907d8f4d69e68432f3/libexec/pyenv#L97-L100 ?  This looks like it might be used to allow for plugins to add something to the path, but am not sure.

> we will still modify PATH on non-system python

Yes, but that's in line with rbenv now, and follows its philosophy to add everything from an env at the front of PATH so that it gets preferred.

> `export PYENV_ORIGINAL_PATH="$PATH"`

This could be done - but should be asked for in rbenv then also.

I wonder though if that's really that useful, given that programs then have to check for this additionally (i.e. be aware of it in the first place).

I've seen that you've found a workaround/fix for dvc in https://github.com/iterative/dvc/pull/2525 already.

While the original issue there (https://github.com/iterative/dvc/issues/2506) is not really clear to me, I think having this PR here merged would fix it also already?

efiop:
> You mean ... ? This looks like it might be used to allow for plugins to add something to the path, but am not sure.

Yes, correct. It is not causing any problems for us so far, but still, it deserves a mention 🙂 

> This could be done - but should be asked for in rbenv then also.

Fair enough 🙂 

> I wonder though if that's really that useful, given that programs then have to check for this additionally (i.e. be aware of it in the first place).
> I've seen that you've found a workaround/fix for dvc in iterative/dvc#2525 already.

Yes, programs would have to check for that one explicitly, but at least they would have a way to get original PATH value, without pyenv additions and without having to use heuristics to get original PATH which could break at any point when pyenv decides to modify the order of PATH modification or decides to add something new.

> While the original issue there (iterative/dvc#2506) is not really clear to me, I think having this PR here merged would fix it also already?

Yes, it would for `"$pyenv_command_version" == "system"` case for sure, which is the most common one, but it would still cause problems for `!= system` case 🙂

Thank you for working on this PR! 🙂

blueyed:
> Yes, it would for `"$pyenv_command_version" == "system"` case for sure, which is the most common one, but it would still cause problems for `!= system` case 

Are you sure?  It would only add the "bin" for that Python version then, which is for example good if subprocesses then just call "pip" for example (which should use the "pip" of that version, no?).
Note that that "bin" only contains a few programs, the ones being installed for that version.

So dropping the pyenv-added PATH for non-system version is likely to even cause problems in dvc then, no?

blueyed:
@yyuu 
Please take a quick look at the changes of this PR.  I'll plan to merge this then soon.

libexec/pyenv-exec
```diff
@@ -41,7 +43,7 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-# CPython's `sys.executable` requires the `PYENV_BIN_PATH` to be at the top of the `PATH`.
-# https://github.com/pyenv/pyenv/issues/98
-export PATH="${PYENV_BIN_PATH}:${PATH}"
-exec -a "$PYENV_COMMAND" "$PYENV_COMMAND_PATH" "$@"
+if [ "$pyenv_command_version" != "system" ]; then
```
yyuu: As per my understanding, you needed to add the code around "PYENV_DISPLAY_VERSION" to didtinguish if the executing command is derived from system or not. How about checking if "PYENV_COMMAND_PATH" is starting from "PYENV_ROOT" or not...? With it, I think you don't need to modifu pyenv-which's behavior. How do you think?

efiop:
@blueyed 

> Are you sure? It would only add the "bin" for that Python version then, which is for example good if subprocesses then just call "pip" for example (which should use the "pip" of that version, no?).
Note that that "bin" only contains a few programs, the ones being installed for that version.

Hm, I think you are right, it shouldn't cause problems in `!= system` case, as it would operate within the same `PYENV_BIN_PATH`, indeed. So in that case `dvc run pip` would call `PYENV_BIN_PATH/pip` directly instead of `pip shim` first, as `PYENV_BIN_PATH` would be at the beginning of `PATH`. But that still feels a big fragile and it seems like having `_ORIG` would make it much more deterministic :slightly_smiling_face:

> So dropping the pyenv-added PATH for non-system version is likely to even cause problems in dvc then, no?

It won't cause problems though, because in `pip` case it would simply call `pip shim` and it would deal with the resolution, same way when you run `pip` from the terminal it also goes through `pip shim` instead of directly addressing `PYENV_BIN_PATH/pip`.

libexec/pyenv-exec
blueyed: Awesome, pushed a fixup.

PR #1178, opened 2018-06-28
Added CPython 3.6.6

rhymes:
CPython 3.6.6 was released on 27-06-2018: <https://www.python.org/downloads/release/python-366/>

This PR adds support for it


pradyunsg:
Hey! Sorry if this is the wrong place but I see that this isn't available when I use a Homebrew installed pyenv but 3.7.0 is (they were released the same day). I'm guessing it might be because the last tag is before this PR.

Is there anything I can do to make it possible for me to install 3.6.6 on a Homebrew installed pyenv?

jfpla:
@pradyunsg, just copy the `3.6.6` file from this MR into your pyenv install dir, which in my case is:
`/usr/local/Cellar/pyenv/1.2.5/plugins/python-build/share/python-build/`

greysteil:
@yyuu - any chance we could get a release with this?

Thanks for everything you do on pyenv - I totally rely on it!

Issue #1112, opened 2018-03-08
Pyenv shim spam breaks $PATH, scripts


deanishe:
In addition to shadowing—and therefore breaking—programs installed on the system `$PATH` with its shims, pyenv also breaks `if hash progname; then ...` tests.

If a program isn't installed system-wide, but *is* installed in a venv, the pyenv shim causes the test to erroneously pass, and of course subsequently running the program results in a "hurr durr, this command exists in these Python versions…" error message when a pyenv shim gets run instead of the real program.

The former issue (not calling existing programs) can be worked around with an extension, but AFAIK the latter can't.

Is there a solution to this problem?

pyenv is darn handy for managing venvs and Python versions. A completely borked `$PATH` is *not* a price worth paying, however.

yyuu:
Generally speaking, the usage of shims is the core design of pyenv and it'll prevent the user to the availability of some command by checking presence of the command executable. This is core design issue and there's no easy and general workaround for it. One idea would be invoking some no-op command like `--version` (e.g. `pip --version`) to verify the availability of the command.

deanishe:
Thanks for the response.

> This is core design issue and there's no easy and general workaround for it

Then what needs to change to stop it being a core design issue? Sure, it's possible to work around in most situations, but programs shouldn't be breaking `$PATH` in the first place.

(Any pyenv *does* break `$PATH`: it deliberately breaks the behaviour of they system and other programs by shadowing existing and non-existing real programs with shim scripts that throw errors.)

Maybe I've missed something due to the way I use pyenv (as a better virtualenv), but I don't really understand why pyenv even needs to drop its shims in `$PATH` in the normal case. Their default function ("I look like a real program, but I'm not! You can find one here or here") seems like a command-line version of ISPs hijacking failed DNS lookups and showing their a webpage instead of returning `NXDOMAIN`: it is helpful in certain very limited circumstances, but mostly it just breaks shit because the system no longer works the way literally everything else expects it to.

pyenv can fiddle with `$PATH` when I activate/deactivate a Python version/virtualenv. It should (have an option to) leave `$PATH` the hell alone when I'm not actively using pyenv.

bitfinity:
Yes, I'm trying to use the system python. But python is mapping to ~/.pyenv/shims/python even though I've run `pyenv deactivate`.

blueyed:
The issue is that the shims have to be in the $PATH to be found, so that it can delegate to the selected version (in the case where it is installed/available).
[...]

PR #1185, opened 2018-07-12
[WIP/RFC]: use separate shim paths per version

blueyed:
Main motivation: `which foo` should not return a pyenv shim in case `foo` is installed in some environment, but not the current.
This is a main issue with pyenv, and this fixes it.

Ref: https://github.com/pyenv/pyenv/issues/1112

Add PYENV_SHIM, and use it with remove_from_path

This is required to avoid infinite recursion with "system" version.

Could also use PYENV_SHIM_PATH, but having PYENV_SHIM is more
flexible/useful.

TODO:

- [x] make pyenv-shell and pyenv-global update $PATH.  Currently it only works for "pyenv-local" via cd hook (using zsh-autoenv).
- [ ] needs to remove any shim files in the root dir probably, otherwise it fails when there is a shim "foo" already, and rehash gets called for a version called "foo"
- [ ] pyenv-virtualenv: Also should run (the new) rehashing when activating a virtualenv directly (not using the chpwd hook).

Uses the following chpwd handler (Zsh), e.g. via
[zsh-autoenv](https://github.com/Tarrasch/zsh-autoenv):

```zsh
  _my_autoenv_pyenv_chpwd() {
    # Experimental: maintain separate shims dir for pyenv.
    local version
    if [[ -z "$PYENV_VERSION" ]]; then
      # version=$(pyenv version-name)

      # Optimized version of pyenv-version-name:
      local -a pyenv_version_file
      setopt localoptions extendedglob
      pyenv_version_file=(./(../)#.python-version(NY1:A))
      if (( $#pyenv_version_file )); then
        version=$(<$pyenv_version_file[1])
      else
        if [[ -f "$PYENV_ROOT/version" ]]; then
          version=$(<$PYENV_ROOT/version)
        else
          version=system
        fi
      fi
    else
      version=$PYENV_VERSION
    fi

    if [[ -z "$_PYENV_AUTO_VERSION" || "$_PYENV_AUTO_VERSION" != "$version" ]]; then
      _PYENV_AUTO_VERSION=$version
      out=$(PATH=$PYENV_ROOT/libexec:$PATH pyenv-sh-shell --updated-path $version | grep '^PATH')
      if [[ -z "$out" ]]; then
        echo "chpwd: pyenv-sh-shell $version failed"
      fi
      eval "$out"
    fi
  }
  add-zsh-hook chpwd _my_autoenv_pyenv_chpwd
```

Drawbacks:

- ~~calls `pyenv version-name` every time you change a directory (if
  `pyenv-shell` was not used).  This could be optimized e.g. by looking for the
  `.python-version` file only.~~

/cc @yyuu @mislav

### DIFF
--- a/libexec/pyenv-which
+++ b/libexec/pyenv-which
@@ -41,7 +41,12 @@ IFS="$OLDIFS"
 
 for version in "${versions[@]}"; do
   if [ "$version" = "system" ]; then
-    PATH="$(remove_from_path "${PYENV_ROOT}/shims")"
+    if [[ -n "$PYENV_SHIM" ]]; then
+      shim_dir=${PYENV_SHIM%/*}
+    else
+      shim_dir="${PYENV_ROOT}/shims"
+    fi
+    PATH="$(remove_from_path "$shim_dir")"
     PYENV_COMMAND_PATH="$(command -v "$PYENV_COMMAND" || true)"
   else
     PYENV_COMMAND_PATH="${PYENV_ROOT}/versions/${version}/bin/${PYENV_COMMAND}"
--- a/libexec/pyenv-which
+++ b/libexec/pyenv-which
@@ -41,13 +41,10 @@ IFS="$OLDIFS"
 
 for version in "${versions[@]}"; do
   if [ "$version" = "system" ]; then
-    if [[ -n "$PYENV_SHIM" ]]; then
-      shim_dir=${PYENV_SHIM%/*}
-    else
-      shim_dir="${PYENV_ROOT}/shims"
-    fi
-    PATH="$(remove_from_path "$shim_dir")"
-    PYENV_COMMAND_PATH="$(command -v "$PYENV_COMMAND" || true)"
+    shopt -s extglob
+    # Remove any shim dir from PATH.
+    path_without_shims="$(remove_from_path "${PYENV_ROOT}/shims*([^:])")"
+    PYENV_COMMAND_PATH="$(PATH="$path_without_shims" command -v "$PYENV_COMMAND" || true)"
   else
     PYENV_COMMAND_PATH="${PYENV_ROOT}/versions/${version}/bin/${PYENV_COMMAND}"
   fi
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git/
+.python-version
+.vscode/
+.idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Version History
 
+## 1.2.22
+
++ python-build: Add LDFLAGS for zlib on macOS >= 1100 (#1711)
++ python-build: Add the CPython 3.9.1 (#1752)
++ python-build: Change order of LDFLAGS paths (#1754)
++ python-build: Docker config for testing python-build (#1548)
++ python-build: Put prerequisite for installation before install (#1750)
++ python-build: Add GraalPython 20.3 (#1736)
++ python-build: Add CPython 3.8.7
++ python-build: Added anaconda3-2020.11 (#1774)
++ python-build: Added arm64 architecture support in python-build for macOS  (#1775)
+
 ## 1.2.21
 
 * python-build: Add CPython 3.9.0 (#1706)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM ubuntu:18.04
+
+LABEL maintainer="Chris L. Barnes <chrislloydbarnes@gmail.com>"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -y \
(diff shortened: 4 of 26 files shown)


blueyed:
Works quite well for pyenv-shell now already - but no support for fish yet.

As for pyenv-local, should we install a chpwd hook ourselves?
I know that you can have several functions in Zsh, but do not know if there is only a global one in Bash/fish.

libexec/pyenv-which
```diff
@@ -40,8 +40,10 @@ IFS="$OLDIFS"
 
 for version in "${versions[@]}"; do
   if [ "$version" = "system" ]; then
-    PATH="$(remove_from_path "${PYENV_ROOT}/shims")"
-    PYENV_COMMAND_PATH="$(command -v "$PYENV_COMMAND" || true)"
+    shopt -s extglob
+    # Remove any shim dir from PATH.
+    path_without_shims="$(remove_from_path "${PYENV_ROOT}/shims*([^:])")"
+    PYENV_COMMAND_PATH="$(PATH="$path_without_shims" command -v "$PYENV_COMMAND" || true)"
```
blueyed: This required change breaks github.com/yyuu/pyenv-which-ext, but it is also not necessary with this PR anymore.
(pyenv-which-ext fails to remove the "system" shims dir (which should be empty, but if not (due to a bug) causes an infinite loop)).

mislav:
Interesting approach! I can definitely see value in fixing `which foo` depending on Python version, but I think the extra complexity necessary to pull this off might not be worth porting to rbenv at this time. If rbenv ended up depending on a chpwd shell handler, I would rather recommend using chruby instead.

1oglop1:
Hi guys any progress on this one?

blueyed:
@1oglop1 
I am still using it myself, and invite you to try it - just check out my PR use it from there.
There are issues with other plugins at least, so if you use pyenv-virtualenv also (which is not really necessary), you should use my PR from there, too.
Also make sure to install a chpwd handler (and example for zsh/zsh-autoenv is above).
Later we would/should ship a chpwd handler probably, but I also agree with @mislav somehow that there are light solutions after all.

I am using pyenv myself mainly only for installing different Python versions, and use normal virtualenvs per project (created using `python -m venv .venv` and then have (another) chpwd handler to activate those automatically).

1oglop1:
@blueyed Thanks I will try later, I was happy having it installed via brew, but I will.

btw what's `chpwd`? I've never heard of it and I'm not using `zsh` since it's not "standard" shell installed on servers. 

But if I understand it correctly it does the same thing as `direnv`

blueyed:
You can install it in parallel using `~/.pyenv` (https://github.com/pyenv/pyenv#basic-github-checkout), but make sure to only use it from there then.  After checking it out there, add my remote and check out this branch there.
`chpwd` is a hook Zsh and Bash provide to execute a function if you change a directory, and direnv provide this via `direnv hook bash` for Bash.
You can use direnv (if you are using it already), but would need to change parts of `_my_autoenv_pyenv_chpwd`, since it uses Zsh globbing (the "Optimized version of pyenv-version-name" part).
Therefore it is not really trivial.

blueyed:
> You can use direnv 

Not really, sorry.

But you can use the method it uses to create the chpwd hook:
```
_pyenv_hook() {
  local previous_exit_status=$?;
  …
  return $previous_exit_status;
};
if ! [[ "$PROMPT_COMMAND" =~ _pyenv_hook ]]; then
  PROMPT_COMMAND="_pyenv_hook;$PROMPT_COMMAND";
fi
```

1oglop1:
@blueyed I finally got some time, to play around your branch.
It took me some time since git installation does not mention that you need to separately install `python-build` plugin to get `pyenv install` working. 
Now I can do `pyenv shell <version>` but my default path is still pointing to 
`~/.pyenv/shims` instead of `~/.pyenv/shims/<version>` + `.python-version` doesn't do anything. Is this connected to `chpwd` you mentioned or something else?
anyway spearating shims based on python version sounds like a very good idea.

ben-albrecht:
This seems to be working for me so far. I don't know if all of this is necessary, but I ended up doing the following to transition after checking this branch out in my `$PYENV_ROOT`:

1) `git clean -fdx $PYENV_ROOT/shims`
2) reinstall my python envs
3) Stop using pyenv-virtualenv (I didn't actually try it, but the PR description implies this is not yet supported)

native-api:
This PR is now so polluted with unrelated commits that I cannot see what it actually does.

In particular, I don't see the coveted "chdir hook" that is the key to its usability.

blueyed:
@native-api 
Yeah, not sure what's up there - could rebase it again (the first commit is "WIP: use separate shim paths per shell/version", 4e7d582).

> In particular, I don't see the coveted "chdir hook" that must be the key to its usability.

There's an example for one in Zsh in the description itself.

PR #1191, opened 2018-07-29
Link Python against local OpenSSL

jhunkeler:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1184

### Description
- [ ] pyenv builds/installs `openssl` into a sub-directory of a new installation prefix, but Python is not configured to use it. Without `./configure --with-openssl=${somewhere}` it will favor the system's OpenSSL. In the case of Darwin, this always fails because the version of OpenSSL is no longer supported (Python >=3.7.0).

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1504,6 +1504,9 @@ use_homebrew_openssl() {
 }
 
 build_package_mac_openssl() {
+  OPENSSL_VERSION_LTS='1.0.*'
+  OPENSSL_VERSION_STABLE='1.1.*'
+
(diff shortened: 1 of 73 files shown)


al45tair:
This problem still exists; this fix is important for users on macOS — they shouldn't have to install their own copy of OpenSSL somewhere when pyenv already downloads and builds it!

al45tair:
OK, so the patch as provided doesn't actually fix the problem, because it tries to test for the presence of the built OpenSSL before the build actually runs. This fails, so it doesn't link against the newly built OpenSSL.

Issue #886, opened 2017-04-01
3.6.1 Fails to install correctly on FreeBSD 11: invalid Python executable


xaque208:
```
$ CFLAGS="-DOPENSSL_NO_SSL2" /home/zach/.pyenv/bin/pyenv install 3.6.1
Downloading Python-3.6.1.tgz...                                       
-> https://www.python.org/ftp/python/3.6.1/Python-3.6.1.tgz
Installing Python-3.6.1...
ERROR: invalid Python executable: /home/zach/.pyenv/versions/3.6.1/bin/python3.6

The python-build could not find proper executable of Python after successful build.
Please open an issue for future improvements.
https://github.com/pyenv/pyenv/issues

BUILD FAILED (FreeBSD 11.0-RELEASE-p8 using python-build 1.0.10-2-geef042a)

Inspect or clean up the working tree at /tmp/python-build.20170401121039.4503
Results logged to /tmp/python-build.20170401121039.4503.log

Last 10 log lines:
rm -f /home/zach/.pyenv/versions/3.6.1/bin/pyvenv
(cd /home/zach/.pyenv/versions/3.6.1/bin; ln -s pyvenv-3.6 pyvenv)
if test "x" != "x" ; then  rm -f /home/zach/.pyenv/versions/3.6.1/bin/python3-32;  (cd /home/zach/.pyenv/versions/3.6.1/bin; ln -s python3.6-32 python3-32)  fi
rm -f /home/zach/.pyenv/versions/3.6.1/share/man/man1/python3.1
(cd /home/zach/.pyenv/versions/3.6.1/share/man/man1; ln -s python3.6.1 python3.1)
if test "xupgrade" != "xno"  ; then  case upgrade in  upgrade) ensurepip="--upgrade" ;;  install|*) ensurepip="" ;;  esac;   ./python -E -m ensurepip  $ensurepip --root=/ ;  fi
Collecting setuptools
Collecting pip
Installing collected packages: setuptools, pip
Successfully installed pip-9.0.1 setuptools-28.8.0
```

The CFLAGS were added after looking for other freebsd related issues on this project, and I've been building other versions of python using pyenv with that method for some time now.

blueyed:
Any update on this?

xaque208:
This remains to be an issue on 3.6.2.

xaque208:
This remains an issue on 3.6.4.
```
Creating directory /home/zach/.pyenv/versions/3.6.4/share/man
[...]

Issue #866, opened 2017-03-21
Update readme to reflect change in Homebrew


truthdoug:
The instructions  mentioned in this file were removed from the Homebrew caveats since they weren't specific to homebrew. See discussion in [this closed homebrew PR](https://github.com/Homebrew/homebrew-core/pull/11209).

PR coming.

PR #1196, opened 2018-08-06
Fix malloc bin symlinks

xaque208:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/886

### Description
- [x] See the commit message for details.

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1641,7 +1641,7 @@ build_package_symlink_version_suffix() {
   # Not create symlinks on `altinstall` (#255)
   if [[ "$PYTHON_MAKE_INSTALL_TARGET" != *"altinstall"* ]]; then
     shopt -s nullglob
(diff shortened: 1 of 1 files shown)


xaque208:
Let me know what you think.  I believe I may yet need to include another symlinks for pip, when only pip3 is created.  I can follow up later.

native-api (changes requested):
* [There are other suffixes beside `m`](https://stackoverflow.com/questions/16675865/difference-between-python3-and-python3m-executables/16677339#16677339), some have come into disuse since ([`m` is no longer used since 3.8](https://github.com/pypa/wheel/pull/303)).

native-api:
It turns out, CPython's Makefile [already links `pythonX.Y<suf>` to `pythonX.Y` since at least `2.7.4` and `3.2`](https://github.com/python/cpython/commit/771d33e1134111a041df78acf7761a9f015d4ea5). so this is not needed.

I couldn't reproduce the error in #866, and there was no debug trace given as required by the issue template -- so can't say why it happened.

PR #1216, opened 2018-09-19
Fix abs_dirname for relative symlinks in same directory

blueyed:
Ref (abs_dirname in bats): https://github.com/sstephenson/bats/pull/224
Ref: https://github.com/rbenv/rbenv/pull/868

Fixes https://github.com/pyenv/pyenv/issues/580


chrahunt:
LGTM.

PR #1217, opened 2018-09-19
Force y, Y, yes or YES to confirm installation

remilapeyre:
When installing or removing a python version `pyenv` access anything starting with a y as a confirmation.

This lead to an issue where one of our scripts returned an error message starting by Y. The proposed patch checks explicitly for one of y, Y, yes or YES in order to avoid errors.


chrahunt:
LGTM.

PR #1237, opened 2018-11-03
force unset shell pipefail option

llehtahw:
pyenv fails with following script if greadlink cannot be found in PATH

```bash
export SHELLOPTS
set -o pipefail
set -e
python -V || echo python failed # python fails
```

looking forward patches to fix

and here one proposed

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -64,6 +64,7 @@ create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
 #!/usr/bin/env bash
 set -e
+set +o pipefail
 [ -n "\$PYENV_DEBUG" ] && set -x
 
 program="\${0##*/}"


blueyed:
Related: #1273 

The following would be great, but requires a recent `env`:
```diff

diff --git c/libexec/pyenv-rehash i/libexec/pyenv-rehash
index b6a1c84..5199bb5 100755
--- c/libexec/pyenv-rehash
+++ i/libexec/pyenv-rehash
@@ -72,7 +72,7 @@ fi
 # serves as a locking mechanism.
 create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
-#!/usr/bin/env bash
+#!/usr/bin/env -S bash -p
 set -e
 [ -n "\$PYENV_DEBUG" ] && set -x
 
```

blueyed:
https://github.com/rbenv/rbenv/pull/1151

PR #1246, opened 2018-11-24
python-build: add anaconda[23]-5.3.1

binbjz:

### Prerequisite

### Description

### Tests


plugins/python-build/share/python-build/anaconda2-5.3.1
```diff
@@ -0,0 +1,19 @@
+case "$(anaconda_architecture 2>/dev/null || true)" in
+"Linux-x86" )
+  install_script "Anaconda2-5.3.1-Linux-x86" "https://repo.continuum.io/archive/Anaconda2-5.3.1-Linux-x86.sh#5685ac1d4a14c4c254cbafc612c77e77" "anaconda" verify_py27
```
yyuu: Is the checksum a SHA256...? It looks like shorter than expected 🤔

PR #1247, opened 2018-11-25
Add support for fd(1), for a 4x+ definition speedup

neersighted:
I'm not entirely sure why `find` is used instead of `ls` (changed in 4a52bfaff971b1326ab023c950f95f663f9dfc6e), as is used in upstream rbenv, but the command presents a significant bottleneck in `pyenv install -l`/`python-build --definitions`. 

If [fd](https://github.com/sharkdp/fd) is installed, we can use it to achieve a speedup of 4x, or better, in my testing:

```
Benchmark #1: ./plugins/python-build/bin/python-build --definitions
  Time (mean ± σ):     589.4 ms ±  28.2 ms    [User: 232.3 ms, System: 2252.9 ms]
  Range (min … max):   551.7 ms … 659.1 ms

Benchmark #2: /home/neersighted/.local/share/pyenv/plugins/python-build/bin/python-build --definitions
  Time (mean ± σ):      2.835 s ±  0.023 s    [User: 104.1 ms, System: 2396.6 ms]
  Range (min … max):    2.799 s …  2.871 s

Summary
  './plugins/python-build/bin/python-build --definitions' ran
    4.81 ± 0.23 times faster than '/home/neersighted/.local/share/pyenv/plugins/python-build/bin/python-build --definitions'
```

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1879,7 +1879,12 @@ usage() {
 
 list_definitions() {
   { for DEFINITION_DIR in "${PYTHON_BUILD_DEFINITIONS[@]}"; do
-      [ -d "$DEFINITION_DIR" ] && find "$DEFINITION_DIR" -maxdepth 1 -type f -print0 | xargs -0 -n 1 basename 2>/dev/null
+      [ -d "$DEFINITION_DIR" ] || continue
+      if command -v fd >/dev/null; then
+        fd . "$DEFINITION_DIR" --max-depth 1 --type f --exec basename
+      else
+        find "$DEFINITION_DIR" -maxdepth 1 -type f -print0 | xargs -0 -n 1 basename 2>/dev/null
+      fi
     done
   } | sort_versions | uniq
 }


neersighted:
Note that alternative, `ls` or `dir` could be used, if there was no real reason for switching away. `ls` is by far the most performant option:

```
Benchmark #1: find share/python-build -maxdepth 1 -type f -print0 | xargs -0 -n 1 basename
  Time (mean ± σ):      2.576 s ±  0.024 s    [User: 90.5 ms, System: 2070.3 ms]
  Range (min … max):    2.539 s …  2.619 s

Benchmark #2: find share/python-build -maxdepth 1 -type f -exec basename {} \;
  Time (mean ± σ):      2.702 s ±  0.021 s    [User: 92.0 ms, System: 2145.3 ms]
  Range (min … max):    2.672 s …  2.740 s

Benchmark #3: fd . share/python-build --max-depth 1 --type f --exec echo {/}
  Time (mean ± σ):     607.9 ms ± 128.8 ms    [User: 257.7 ms, System: 2112.5 ms]
  Range (min … max):   483.2 ms … 886.4 ms

Benchmark #4: fd . share/python-build --max-depth 1 --type f --exec basename
  Time (mean ± σ):     589.5 ms ± 104.1 ms    [User: 242.0 ms, System: 1996.9 ms]
  Range (min … max):   462.4 ms … 803.9 ms

Benchmark #5: ls share/python-build
  Time (mean ± σ):      10.1 ms ±   1.3 ms    [User: 1.2 ms, System: 7.7 ms]
  Range (min … max):     7.4 ms …  15.3 ms

Summary
  'ls share/python-build' ran
   58.43 ± 12.90 times faster than 'fd . share/python-build --max-depth 1 --type f --exec basename'
   60.26 ± 15.06 times faster than 'fd . share/python-build --max-depth 1 --type f --exec echo {/}'
  255.34 ± 33.93 times faster than 'find share/python-build -maxdepth 1 -type f -print0 | xargs -0 -n 1 basename'
  267.87 ± 35.57 times faster than 'find share/python-build -maxdepth 1 -type f -exec basename {} \;'
```

yyuu:
The only reason why we are using find(1) there is that we needed to exclude "patches" directory from the output. The easiest and sufficient workaround might be just applying "grep -vx patches" to ls(1) output.

PR #1273, opened 2019-02-01
Fix for error with `set -u` being used (via SHELLOPTS in a makefile)

dmlemos:
### Prerequisite

### Description

Given a makefile with the following:
```make
DEBUG ?= false
ifeq (${DEBUG}, true)
export SHELLOPTS += xtrace
endif

all:
  az account show
```

When I invoke make, I get this error:
```bash
$ make
++ az account show
+ export PYTHONPATH=/usr/local/Cellar/azure-cli/2.0.56/libexec/lib/python3.7/site-packages
+ PYTHONPATH=/usr/local/Cellar/azure-cli/2.0.56/libexec/lib/python3.7/site-packages
+ command -v python3.7
+ python3.7 -m azure.cli account show
+ set -e
/Users/xxx/.pyenv/shims/python3.7: line 12: PYENV_DEBUG: unbound variable
```

The script doesn't fail per see, but the python command is never executed.

The solution is to turn off `nounset` option.
Or perhaps just adding the edge case with:
```bash
if [[ "$SHELLOPTS" =~ xtrace ]]; then
  set +u
fi
```
Didn't see any harm in setting it without checking the variable.

I am sure this is an edge case, but thought to report it nonetheless. It only happens with SHELLOPTS and make.
It seems like it is not exclusive to xtrace, but something very specific with these two together, the below produces the same error (without tracing):
```make
export SHELLOPTS += emacs
```

### Tests
Not sure how to test this

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -64,6 +64,7 @@ create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
 #!/usr/bin/env bash
 set -e
+set +u
 [ -n "\$PYENV_DEBUG" ] && set -x
 
 program="\${0##*/}"


blueyed:
1. it looks like something that should go into rbenv already
2. do you get the `set -u` behavior with `/usr/bin/env bash` already?  (i.e. does `echo $FOO` triggers it there).
3. what is your normal shell?
4. what is `az`?  Is it the same within the Makefile and outside of it?  (PATH related)
5. `make` might use a different shell than your normal one, with different init scripts

In general I think it might be worth fixing, but could also look like this maybe:
```diff
diff --git a/libexec/pyenv-rehash b/libexec/pyenv-rehash
index 5c421c4..d4a65a9 100755
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -64,7 +64,7 @@ create_prototype_shim() {
   cat > "$PROTOTYPE_SHIM_PATH" <<SH
 #!/usr/bin/env bash
 set -e
-[ -n "\$PYENV_DEBUG" ] && set -x
+[ -n "\${PYENV_DEBUG:-}" ] && set -x
 
 program="\${0##*/}"
 if [[ "\$program" = "python"* ]]; then
```

As for a test, it could just `set -u` there and call a shim I guess.

blueyed:
Ok, exporting `SHELLOPTS` is the main issue here.

But a) using += here is wrong, since it is joined by colons, not spaces, and b) it is still unclear where the "nounset" (set -u) is coming from then.

For reference:
```make
SHELL := bash
export SHELLOPTS=xtrace:nounset

all:
	echo "${SHELLOPTS}"
	bash -c 'true; echo $$FOO'
```

blueyed:
https://github.com/rbenv/rbenv/pull/1151

dmlemos:
Apologies for the late reply. I was on holidays.

> 1. it looks like something that should go into rbenv already
The relationship between this project and rbenv is not clear, perhaps clarify this on the README.

> 2. do you get the `set -u` behavior with `/usr/bin/env bash` already?  (i.e. does `echo $FOO` triggers it there).
Yes.

> 3. what is your normal shell?
bash

> 4. what is `az`?  Is it the same within the Makefile and outside of it?  (PATH related)
I should have been clearer about this. It's the [azure cli](https://github.com/Azure/azure-cli) used to interact with Azure could. I just used what I was working on at the time.

> 5. `make` might use a different shell than your normal one, with different init scripts
Tried with both `sh` and `bash` and behaviour is similar, although I cannot fully reproduce it anymore.

> a) using += here is wrong, since it is joined by colons, not spaces
Correct. Thanks for pointing that out.

> b) it is still unclear where the "nounset" (set -u) is coming from then.
Nothing else was being set, just make calling a python tool directly :beaver:

I cannot seem to reproduce the issue again nonetheless.

On a related note, setting `xtrace` propagates through pyenv and conflicts with `PYENV_DEBUG`. Is this the behaviour expected? I'd like to see `pyenv` deal with the former, but only honour debug for the later.

Char-Aznable:
I have got similar unbound variable error when calling pyenv in a bash script:
```
#!/bin/bash
set -exuo pipefail
eval "$(pyenv init -)"
export -f pyenv
pyenv shell myenv
```

native-api:
https://github.com/rbenv/rbenv/pull/1151#issuecomment-833088505:

> Thank you, but I think that if `SHELLOPTS=nounset:pipefail` are exported, they will affect every single rbenv internal, not just the shim script. So it's likely that many more edits would be necessary to fix rbenv. Instead, I would advise that users not export SHELLOPTS globally.

---

I also don't see a relevance between 

> ```makefile
> export SHELLOPTS += xtrace
> ```

and the error -- because `xtrace` sets `-x`, not `-u`.

native-api:
Let's take a step back, create an issue and discuss how we should approach this.
There are multiple use cases here (e.g. regular use vs `eval $(pyenv init -)`) and multiple possible solutions, we need to get an overview of them all and weigh their pros and cons.

PR #1274, opened 2019-02-05
Support XCode 10 on Mac OS

### LINKED ISSUE #1219: Install failed, "zlib not available" on macOS Mojave
### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): macOS Mojave 10.14
- [x] OS architecture (e.g. amd64):amd64
- [x] pyenv version: 1.2.7
- [x] Python version: 3.7.0 & 2.7.15
- [x] C Compiler information (e.g. gcc 7.3): Apple LLVM version 10.0.0 (clang-1000.11.45.2)
- [ ] Please attach verbose build log as gist

see https://github.com/Homebrew/homebrew-core/issues/29176#issuecomment-398656987

Temporary workaround:

```
CFLAGS="-I$(xcrun --show-sdk-path)/usr/include" pyenv install ...
```

andrew-humu:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1219

### Description

Fixes #1219

On MacOS 10.9 and later, uses `xcrun --show-sdk-path` to find SDK path, instead of relying on Mac OS headers being available at the system level.  This is required because of a change in XCode 10's default behaviour.

https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes#3035624

Inspired by these comments:

https://github.com/pyenv/pyenv/issues/1219#issuecomment-429331397

https://github.com/pyenv/pyenv/issues/1219#issuecomment-451643266

https://github.com/pyenv/pyenv/issues/1219#issuecomment-452315815

and this blog post:

https://medium.com/@pimterry/setting-up-pyenv-on-os-x-with-homebrew-56c7541fd331

### DIFF
--- a/src/shobj-conf
+++ b/src/shobj-conf
@@ -199,6 +199,10 @@ darwin*)
 			;;
 		 *) 	# Mac OS X 10.9 (Mavericks) and later
 			SHOBJ_ARCHFLAGS=
+
+			# Xcode 10 and later no longer install mac OS SDK headers in the base system
+			SHOBJ_CFLAGS="${SHOBJ_CFLAGS} -I${xcrun --show-sdk-path}/usr/include"
+
 			# for 32 and 64bit universal library
 			#SHOBJ_ARCHFLAGS='-arch i386 -arch x86_64'
 			#SHOBJ_CFLAGS=${SHOBJ_CFLAGS}' -arch i386 -arch x86_64'
@@ -209,7 +213,7 @@ darwin*)
 		;;
 	esac
 
-	SHLIB_LIBS='-lncurses'	# see if -lcurses works on MacOS X 10.1 
+	SHLIB_LIBS='-lncurses'	# see if -lcurses works on MacOS X 10.1
 	;;
 
 openbsd*|netbsd*|mirbsd*)
@@ -349,7 +353,7 @@ hpux9*)
 #
 #	SHLIB_XLDFLAGS='+b $(libdir)'
 #	SHLIB_LIBSUFF='sl'
-#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'	
+#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'
 
 	;;
 
@@ -381,7 +385,7 @@ hpux10*)
 #
 #	SHLIB_XLDFLAGS='+b $(libdir)'
 #	SHLIB_LIBSUFF='sl'
-#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'	
+#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'
 
 	;;
 
@@ -412,7 +416,7 @@ hpux11*)
 #
 #	SHLIB_XLDFLAGS='+b $(libdir)'
 #	SHLIB_LIBSUFF='sl'
-#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'	
+#	SHLIB_LIBVERSION='$(SHLIB_LIBSUFF).$(SHLIB_MAJOR)'
 
 	;;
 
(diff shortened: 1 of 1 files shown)


dakl:
What's the status on this?

chrahunt:
@dakl this needs to be reviewed and tested. I don't personally have access to a mac, so I cannot test or do any of the sanity checks I would normally for a code review. Here are some details that might help things along:

## Review

Anyone can do a code review. If you have access to a mac that is better since you can quickly address small questions yourself.

Things to consider for any reviewer (in order):

1. What is the proposed change trying to fix? Is the approach a good one? - I'm not familiar with the file that was edited but it looks like it is only used for the build of the executable that may be used to speed up path resolution. It doesn't look like this is related to the Python build.
2. Is there anything basically wrong, e.g. syntax or typos? - One note, `shobj-conf` is a shell script but the fix here looks like it is trying to invoke a subcommand using `${}`. Not sure what shell that is compatible with but it isn't bash. Maybe `$()` was meant instead?
3. Is the fix compatible with all relevant macOS and XCode versions? What versions was it tested on?
4. What are the failure scenarios and how are they handled? e.g. `xcrun` not available, `--show-sdk-path` invalid option. Would there at least be something in the logs so we'd be able to debug it if someone has an issue?

## Testing

Since this repo does not currently have a CI setup that tests builds, we rely on people on various platforms and configurations to report whether the fix worked for them. A :+1: is nice, but a comment stating that you were able to pull this PR locally and successfully compile (including your macOS and XCode version) is better.

andrew-humu:
Thanks for commenting, @chrahunt!

I'll just respond to a couple points with clarification:

2. If I'm being honest, `${}` is just syntax I copied from elsewhere in this file.  From what I can tell it is standard bash syntax for variable expansion (different from `$()`, which evaluates a command in a subshell).  It seems likely to me that if theis were problematic, that would have surfaced before now. 
 https://stackoverflow.com/questions/27472540/difference-between-and-in-bash

---

4. I think this is a good question to raise.  I'm not a MacOS guru, and I'm not sure what exactly the conditions are for `xcrun` to be available.  Howeever, I figured it was better to propose a solution and start the conversation than leave things in their current, broken state.

joshfriend:
I really want to get this fixed, so I will try to find some time to test it since I am a full time macOS user.

I just need to figure out how to undo the installation of the sdk to the legacy location that I did previously:

```
sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /
```

jeffb4:
On 10.14.1 this patch still exhibits bad `pyenv install` behavior
`CFLAGS="-I$(xcrun --show-sdk-path)/usr/include" pyenv install -v 3.6.7` works fine, whereas `pyenv install` on this branch still fails with zlib error

I do not have the global sdk installation Josh mentions above, and I'm happy to test any iterations you might have to fix this

andrew-humu:
Thanks @jeffb4!  I will try to find some time!

src/shobj-conf
```diff
@@ -199,6 +199,10 @@ darwin*)
 			;;
 		 *) 	# Mac OS X 10.9 (Mavericks) and later
 			SHOBJ_ARCHFLAGS=
+
+			# Xcode 10 and later no longer install mac OS SDK headers in the base system
```
ushuz:
```suggestion
			# Xcode 10 and later no longer install macOS SDK headers in the base system
```

cclauss:
Can you please add the change from #1333 and make sure the tests pass?  Thanks!

joshfriend:
Firstly: really sorry to let this sit so long, haven't done much python at work recently, which means my related side work also takes a hit.

Secondly: The approach in this PR doesn't work because if you trace the path of the scripts calling `shobj-conf` (which is what's modified here) upward, you get one path:

```
test/run
    -> src/configure
        -> shobj-conf
```

So basically at no point during `pyenv install` is the file modified here called.

I'm working on this right now actually, have it working, but it needs polishing (helpful errors if there is no `xcrun`, potentially falling back to homebrew zlib, idk)

joshfriend:
Closing in favor of #1339

Thank you for your valiant effort, @andrew-humu 😄

Issue #272, opened 2014-11-16
PyPI package (install with pip install pyenv)


obestwalter:
First things first: **big thumbs up for pyenv :)** for me it takes away the cumbersomeness of switching between Py2 and Py3 and different projects in combination with pyenv-virtualenvwrapper.

I am aware that this is technically not a python Package but I still think it belongs on PyPi and should be pip installable. I am not deeply into Python Packaging myself, as most of the stuff  I do is SAAS, but I would be interested if this is allowed. I am sure it is technically possible and not even very hard to do. Did you look into that already?

The sad state atm is that there is [pyenv package](https://pypi.python.org/pypi/pyenv/0.0.1) by @kennethreitz on PyPi with similar motivation, but it is clearly abandoned as it is pointing to a non existing Github repo. There should be at least a pointer to this project here instead.

obestwalter:
@kennethreitz freed the name on PyPi (thanks Kenneth :) ): https://sourceforge.net/p/pypi/support-requests/441/

For now I registered my fork of the installer there, basically just to secure the name and clarify the intention till it is decided how to move forward with this.

@yyuu I would like to maintain the Python specific wrapper part of the installer. I will think about how to deal best with keeping the two in sync (easiest would be to fetch the newest installer script directly on each install, but that would cause problems for people which use devpi, so the installer bin should somehow be part of the package). 

Another idea would be to merge the wrapper code directly into the main project and maintain it together. I actually like that idea. As it really belongs together. There are just two possible ways to use the installer then. And the code of the shell script and the setup.py stuff do not collide anyway.

Anyway - I will whip my fork into shape over the next few days and go from there.

yyuu:
Great effort :+1:

I think it is better to merge the changes in [obestwalter/pyenv-installer](https://github.com/obestwalter/pyenv-installer) into [yyuu/pyenv-installer](https://github.com/yyuu/pyenv-installer). Plus, I'd like to add you as a collaborator of yyuu/pyenv-installer. How do you think of it?

obestwalter:
Great! I will get the fork into a shape that is not cpmletely embarassing and then I will send you a pull request. Thanks for taking me on board :)

obestwalter:
ok - here it is: https://github.com/yyuu/pyenv-installer/pull/16

@yyuu I released it already from my fork - please let me know if you see any problems. 

[...]

PR #1276, opened 2019-02-11
supporting pip install pyenv

kirankotari:
#272 Change in pyenv structure to support pypi

command `pip install pyenv --target $HOME/.pyenv`
How to get updates: add the installed path to eash_install.pth file to recognise pyenv by pip


kirankotari:
@yyuu update the paths in test files. 

Expected in following location: **pyenv.bats**

```bats
@test "default PYENV_ROOT" {
  PYENV_ROOT="" HOME=/home/mislav run pyenv root
  assert_success
  assert_output "/home/mislav/.pyenv"
}
```

joshfriend:
why was it necessary to relocate **nearly 600** files?

yyuu:
We have once tried to support PyPI as https://github.com/pyenv/pyenv-installer/pull/16 (thanks to @obestwalter ). However, we ended up to abandon the approach since it just added extra complexity and dependencies more than benefit. I don't mean supporting PyPI is worthless, but I would never prefer this to be merged into the pyenv's main repo.

kirankotari:
I too agree of not getting merged and changing nearly 600 files is not good. The new structure is like python package which is recognised by pypi and pip install. 

when I am porting pyenv for windows, I have successfully able to do `pip install pyenv-win` thought the similar approach might help pyenv.

kirankotari:
Checking for alternative solutions without changing the structure.

PR #1284, opened 2019-02-17
Add micropython 1.10

scop:


plugins/python-build/share/python-build/micropython-1.10
```diff
@@ -0,0 +1,2 @@
+#require_gcc
+install_package micropython-1.10 "https://micropython.org/resources/source/micropython-1.10.tar.xz#8488ec243217c4578f1cf2ed20e749344b01510a52d443da95304949cf7abfdb" micropython
```
chrahunt: Please wrap this in a check for `has_tar_xz_support` and provide a tar.gz fallback. You can see an exaple in [3.7.2](https://github.com/pyenv/pyenv/blob/ddb335c1979554196ccd18e7b2c26cc747cd4a89/plugins/python-build/share/python-build/3.7.2#L4).

chrahunt: Please wrap this in a check for `has_tar_xz_support` and provide a tar.gz fallback. You can see an exaple in [3.7.2](https://github.com/pyenv/pyenv/blob/ddb335c1979554196ccd18e7b2c26cc747cd4a89/plugins/python-build/share/python-build/3.7.2#L4).

scop: Done.

chrahunt:
Thanks!

PR #1285, opened 2019-02-17
Use xz compressed micropython 1.9.* tarballs

scop:


chrahunt:
For the time being I think we should prefer gz over xz, where available. The "portability" of gz over xz means one less thing that could cause build issues, and users upgrading a large number of machines may always use [`PYTHON_BUILD_MIRROR_URL`](https://github.com/pyenv/pyenv/blob/d08c9cfb362c5a7e18a92acd2253a16935ad9a99/plugins/python-build/README.md#special-environment-variables) to point to a local resource if bandwidth is a concern.

plugins/python-build/share/python-build/micropython-1.9.3
```diff
@@ -1,2 +1,2 @@
 #require_gcc
-install_package micropython-1.9.3 "https://github.com/micropython/micropython/releases/download/v1.9.3/micropython-1.9.3.tar.gz#f94c0b4834edca625b86c0846c89e609ee8fbdbe98e7f719d5108d74b6b4945e" micropython
+install_package micropython-1.9.3 "https://micropython.org/resources/source/micropython-1.9.3.tar.xz#3aa37065f5ea8df372d36253d0ae6333c68572e577df8558d4b7b93e070c624d" micropython
```
chrahunt: Please wrap this in a check for `has_tar_xz_support`. You can see an exaple in [3.7.2](https://github.com/pyenv/pyenv/blob/ddb335c1979554196ccd18e7b2c26cc747cd4a89/plugins/python-build/share/python-build/3.7.2#L4).

chrahunt: Please wrap this in a check for `has_tar_xz_support`. You can see an exaple in [3.7.2](https://github.com/pyenv/pyenv/blob/ddb335c1979554196ccd18e7b2c26cc747cd4a89/plugins/python-build/share/python-build/3.7.2#L4).

plugins/python-build/share/python-build/micropython-1.9.4
```diff
@@ -1,2 +1,2 @@
 #require_gcc
-install_package micropython-1.9.4 "https://github.com/micropython/micropython/releases/download/v1.9.4/micropython-1.9.4.tar.gz#0db042011bffcbd65362b67eb3cca87eaefa9f2a55b747fa75e922c706b8ce1a" micropython
+install_package micropython-1.9.4 "https://micropython.org/resources/source/micropython-1.9.4.tar.xz#1d358d99e908a94fd6dd0d6b448ee56c5830b609cf27a42e0db3bee40cc52a9e" micropython
```
chrahunt: Same with this one.

scop:
Now tests for tar.xz support.

chrahunt:
Thank you!

PR #1289, opened 2019-02-18
Install python-gdb.py.

### LINKED ISSUE #1190: Install python-gdb.py
I just had occasion to learn about and use `python-gdb.py` yesterday and it seems like it would be a good idea for pyenv to install it rather than requiring a builder to keep the source tree around.  Unless i'm missing something about the script requiring the whole source tree.  It gets generated directly in the root of the build right next to the binaries.

Hopefully I'll get a chance soon to look at adding this myself but I like to announce first anyways so people can weigh in and tell me preemptively if I'm on the wrong track.

chrahunt:
The script in `Tools/gdb/libpython.py` has helper functions for
debugging Python processes under gdb. When installed to `{exe}-gdb.py`
it will automatically be loaded by gdb assuming it is in the safe path.

gdb resolves all symlinks, so having the single script is sufficient
regardless of how Python is invoked.

Since it is usually only included as part of a dbg/dev package, this
file is intentionally excluded from install by the CPython build
scripts. Like the packaging with EPEL/Debian, we opt to manually copy
it after the build/install.

To ensure this file gets picked up it is enough for users to put this in
their `~/.gdbinit`:

```
add-auto-load-safe-path ~/.pyenv
```

Fixes #1190.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1813,6 +1813,16 @@ build_package_verify_py38() {
   build_package_verify_py37 "$1" "${2:-3.8}"
 }
 
+# Copy Tools/gdb/libpython.py to pythonX.Y-gdb.py (#1190)
+build_package_copy_python_gdb() {
+  if [ -e "$BUILD_PATH/$1/Tools/gdb/libpython.py" ]; then
+    local version_re='-([0-9]\.[0-9]+)'
+    [[ "$1" =~ $version_re ]]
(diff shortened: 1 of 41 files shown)


chrahunt:
@altendky, do you mind giving this a try?

altendky:
@chrahunt, thanks for working this up.  I normally try to be really responsive but I'm running behind.  Maybe tonight depending how other things go.  Cheers.

altendky:
A quick check worked for me both directly and via a venv.  Thanks!  Luckily I don't need this often but...  when I do it'll be nice to have it handy.  Cheers.

PR #1291, opened 2019-02-18
Add CI builds that actually build Python.

chrahunt:
For now this just builds the latest CPython, but it can be
extended to include other versions.

Skip with `[skip build]` in commit message.

This covers (Ubuntu Xenial) Linux and macOS.

Currently there is an issue in Travis where multiple `osx_image` entries results in multiple linux builds being created. I've raised a support ticket, but it shouldn't cause problems unless we add many more build tests (limit is 200).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,78 @@
-sudo: false
-install: git clone --depth 1 https://github.com/sstephenson/bats.git
-script: make test
+# Two types of tests:
+#
+# 1. shell tests run using bats
+# 2. build tests also run using bats
+#
+# Shell tests are run unconditionally on Linux (see jobs.include).
+#
+# Build tests can be skipped by starting the commit message with
+# '[skip build]'.
+#
+# Build tests are run on platforms generated from all combinations
+# of keys below.
 language: c
+
+os:
+- linux
+- osx
+
+dist:
+- xenial
+
+# https://docs.travis-ci.com/user/reference/osx/#macos-version
+osx_image:
+- xcode9.4
+- xcode10
+
 env:
-  - PYENV_NATIVE_EXT=
-  - PYENV_NATIVE_EXT=1
+- PYTHON_BUILD_VERSION=3.8-dev
+- PYTHON_BUILD_VERSION=3.7.2
+
+before_install:
+- date +%Y-%m-%dT%H:%M:%S
+
+install: git clone --depth 1 https://github.com/bats-core/bats-core.git bats
+
+# Default for auto-generated jobs.
+script: make test-build
+
+after_script: |
+  cat "$(ls -tr ${TMPDIR:-/tmp}/python-build.*.log | tail -1)"
+
+jobs:
+  include:
+  # Shell-based tests should execute every time.
+  - stage: test shell
+    script: make test
+    env: PYENV_NATIVE_EXT=1
+    after_script: []
+    os: linux
+  - stage: test shell
+    script: make test
+    env: PYENV_NATIVE_EXT=
+    after_script: []
+    os: linux
+
+  exclude:
+  # For each osx_image but one there should be an entry in the exclude
+  # list, to prevent duplicate Linux builds.
+  - os: linux
+    osx_image: xcode9.4
+
+  allow_failures:
+  - env: PYTHON_BUILD_VERSION=3.8-dev
+
+stages:
+- test shell
+- name: test
+  if: NOT (commit_message =~ /^\[skip build\]/)
+
+# Default 
+
 notifications:
   email:
     on_success: never
+
 deploy:
   provider: releases
   api_key:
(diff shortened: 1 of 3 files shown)


chrahunt:
We should fold in some of the items from #617 later. For now this should be good enough to point to as a working example of builds for issues with lots of build problems.

plugins/python-build/test/build/test.bats
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bats
+
+load ../test_helper
+export PATH="$BATS_TEST_DIRNAME/../../bin:$PATH"
+
+export PYTHON_BUILD_VERSION="${PYTHON_BUILD_VERSION:-3.8-dev}"
```
yyuu: "3.8-dev" 🤔 

This would mean pyenv CI will start failing if something bad has been pushed to CPython HEAD. Wouldn't it be sufficient to test some newer released version of CPython, e.g. 3.7.2....?

.travis.yml
```diff
@@ -1,13 +1,69 @@
...
+- xcode10
+
 env:
-  - PYENV_NATIVE_EXT=
-  - PYENV_NATIVE_EXT=1
+- PYTHON_BUILD_VERSION=3.8-dev
+- PYTHON_BUILD_VERSION=3.7.2
+
+before_install:
+- date +%Y-%m-%dT%H:%M:%S
```
yyuu: What is this for...?

.travis.yml
```diff
@@ -1,13 +1,69 @@
...
+
+install: git clone --depth 1 https://github.com/bats-core/bats-core.git bats
+
+# Default for auto-generated jobs.
+script: make test-build
+
+after_script: |
+  cat "$(ls -tr ${TMPDIR:-/tmp}/python-build.*.log | tail -1)"
```
yyuu: If what you are trying to do here is to print out build log, I think passing "--verbose" to python-build would be more _reliable_ than this hack 🤔

Makefile
```diff
@@ -8,5 +8,8 @@ test: bats
 	PATH="./bats/bin:$$PATH" test/run
 	cd plugins/python-build && $(PWD)/bats/bin/bats $${CI:+--tap} test
 
+test-build: bats
+	cd plugins/python-build && $(PWD)/bats/bin/bats $${CI:+--tap} test/build
+
 bats:
-	git clone --depth 1 https://github.com/sstephenson/bats.git
+	git clone --depth 1 https://github.com/bats-core/bats-core.git bats
```
yyuu: 👍

.travis.yml
chrahunt: When running tests I noticed a few times that the workers would hang for a bit after the "Worker Information" and "Build System Information" printed. `before_install` executes before any of our tests or setup so it could be used in conjunction with the test start time (visible by hovering over the duration in the far right column of any build [here](https://travis-ci.com/chrahunt/pyenv/builds/101372597)) to rule out our tests as being the reason for some builds taking a while.

plugins/python-build/test/build/test.bats
chrahunt: Right now in the `.travis.yml` it is set to test 3.7.2 and 3.8-dev. That's a good point about spurious failures - we could configure [`allow_failures`](https://docs.travis-ci.com/user/customizing-the-build/#matching-jobs-with-allow_failures) to allow the 3.8-dev ones to fail without causing the whole build to appear to have failed, but still test it so we can see if there is actually something wrong.

.travis.yml
chrahunt: I would have liked to specify the log file explicitly via e.g. environment variable instead of getting the latest `python-build.*.log` but we don't support that in `python-build` (yet).

No matter what we do it should be in the `after_script` step so it is in its own section separate from the test results/output (see [here](https://travis-ci.com/chrahunt/pyenv/jobs/178677022)).

Do you think it would be better to run with `--verbose` and echo `$output` to a specific file that we cat here instead of this?

plugins/python-build/test/build/test.bats
yyuu: Gotcha. At least I am okay to run test against `3.8-dev`, and, with it we might be able to know potential issues with new release proactively.

.travis.yml
yyuu: Ah, I guess I wasn't understanding what's going on there 😅 Since `python-build` is invoked via `bats`, to see the output `$output` has to be printed.... Okay, now everything is clear (to me) and I'm totally okay with current configurations. Thanks for explanation!

plugins/python-build/test/build/test.bats
chrahunt: We now exclude failures that occur under 3.8-dev. Travis support helped also helped identify a way to exclude the duplicate Linux jobs. We will need to make sure to keep `osx_image` and `jobs.exclude` in-sync until it is fixed, but it's still better than having to explicitly mention all combinations, which will have more later.

Issue #839, opened 2017-02-22
How to install Python 3.6.0 with OpenSSL 1.1 on MacOS?


dimaqq:
I can `brew install openssl@1.1` and get OpenSSL 1.1.0e (atm.)

I'd like to `pyenv install 3.6.0` with that OpenSSL, but it seems that pyenv overrides my choices of CPFLAGS, LDFLAGS, etc.

Is there a way to force its hand?

cculianu:
Uh. I want to do the same. I can't get pyenv to link against openssl@1.1 no matter how hard I try.

What gives?  Openssl1.1 offers features like hashlib.scrypt which are needed by me.

ThomasWaldmann:
Would a bounty help here?

ThomasWaldmann:
Also strange: even if one has installed `readline` and `openssl` (1.0.2) before installing pyenv and building pythons with it, pyenv seems to download these again.

I would like to have pyenv use the existing packages, so I can influence (e.g. via PKG_CONFIG_PATH or compiler/linker flags env variable) which openssl version pyenv uses for building the pythons (because it should be the same version as I use to build borgbackup with).

cculianu:
I'm in agreement here.

I may just put up a PR to handle these cases -- pyenv is pretty nicely architected so people not familiar with it can modify it. That's a good thing!

I really really need this feature too -- the features you describe and the ability to specify openssl 1.1, etc.

:/

ThomasWaldmann:
my comment above was about travis-ci's macOS testing.

strange, in my vagrant-based testing, I see:
```
    darwin64: python-build: use openssl from homebrew
    darwin64: python-build: use readline from homebrew
    darwin64: Downloading Python-3.7.0.tar.xz...
    darwin64: -> https://www.python.org/ftp/python/3.7.0/Python-3.7.0.tar.xz
    darwin64: Installing Python-3.7.0...
    darwin64: python-build: use readline from homebrew
    darwin64: Installed Python-3.7.0 to /Users/vagrant/.pyenv/versions/3.7.0
```
So, pyenv does not download readline and openssl, but picks it up from homebrew, as desired.

cculianu:
Yeah but it picks up 1.0.2 even when you have 1.1.0 installed :'(

ThomasWaldmann:
yeah, can confirm that (even though I pointed PKG_INFO_PATH at openssl@1.1).

cculianu:
[...]

PR #1302, opened 2019-03-16
macOS: prefer homebrew openssl 1.1.x over 1.0.x, see #839

ThomasWaldmann:
previously, it was not possible to compile a python with homebrew openssl 1.1 because the code always queried "openssl" and ignored "openssl@1.1".

now, if 1.1 is installed, it is used to compile python and only if it is not, 1.0 is checked and used.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1492,15 +1492,20 @@ has_broken_mac_openssl() {
 }
 
 use_homebrew_openssl() {
-  local ssldir="$(brew --prefix openssl 2>/dev/null || true)"
-  if [ -d "$ssldir" ]; then
-    echo "python-build: use openssl from homebrew"
-    export PKG_CONFIG_PATH="$ssldir/lib/pkgconfig/:${PKG_CONFIG_PATH}"
-    export CPPFLAGS="-I$ssldir/include ${CPPFLAGS}"
-    export LDFLAGS="-L$ssldir/lib ${LDFLAGS}"
+  local ssl11dir="$(brew --prefix openssl@1.1 2>/dev/null || true)"
+  local ssl10dir="$(brew --prefix openssl 2>/dev/null || true)"
+  if [ -d "$ssl11dir" ]; then
+    echo "python-build: use openssl 1.1 from homebrew"
+    local ssldir=$ssl11dir
+  elif [ -d "$ssl10dir" ]; then
+    echo "python-build: use openssl 1.0 from homebrew"
+    local ssldir=$ssl10dir
   else
     return 1
   fi
+  export PKG_CONFIG_PATH="$ssldir/lib/pkgconfig/:${PKG_CONFIG_PATH}"
+  export CPPFLAGS="-I$ssldir/include ${CPPFLAGS}"
+  export LDFLAGS="-L$ssldir/lib ${LDFLAGS}"
 }
 
 build_package_mac_openssl() {


ThomasWaldmann:
I am aware that this PR is far from perfect, but it could be temporarily used until someone (not me, I don't like shell scripting) writes even better code:

E.g. it could be checked whether `PKG_CONFIG_PATH` (and the other 2 env vars) already has an entry pointing at an openssl location and if so, do not modify them. Then users could choose which openssl version they want by setting these env vars before invoking `pyenv install`. Would also need a piece of docs.

ThomasWaldmann:
BTW, would be cool to have this available soon, I need python with openssl 1.1 for borgbackup testing.

cculianu:
It's late here in Europe but I am super excited you did this! I'll take it for a spin tomorrow and let you know what I think and how it works for me.  THANK YOU for this. This opens up a lot of possibilities as OpenSSL 1.1 has scrypt and other stuff that python makes use of.

yyuu:
Thanks for making changes :+1:

btw, I've been thinking what should happen if,

1. OpenSSL releases newer than 1.1 have been released
2. Homebrew's `openssl` has been upgraded to install some version newer than 1.1

With current code, it will always prefer 1.1 over the _default_ `openssl` of Homebrew. I don't think it's the right behavior, however, in fact, I don't have other better workaround for now. So far I've been waiting for Homebrew to upgrade their default OpenSSL version, but it may gonna require extra time to finish the upgrade. It might be time for us to consider adding this kind of _temporary_ workaround.... 🤔

cculianu:
@yyuu I really appreciate your taking the time to think this through and perhaps being amenable to this.  OpenSSL 1.1 is really so much better than 1.0.2 in terms of feature coverage.. would be great to have it.  Thanks.

ThomasWaldmann:
Guess if this needs changing, it can be just changed again in future. Considering how long it took from 1.0 to 1.1 and how long it takes software and distributions to adapt to changes, I guess this will not frequently change and not quickly need to be changed.

ThomasWaldmann:
Another consideration: is it going to blow up something if pyenv produces a python compiled with / linked to openssl 1.1 and some python-based software compiles / links some C extension with 1.0?

cculianu:
@ThomasWaldmann I'm hoping not.  On Linux at least all the pythons you get on all the distros are compiled & linked against OpenSSL 1.1.  Even a lot of the Pythons you see in the Windows world are linked against this.  pyenv on MacOS has been the odd-man-out with its OpenSSL 1.0 as far as I can tell.

Now -- whether someone built some programs on top of python assuming 1.0.. no idea. Someone more familiar with this packaging zoo might know better.. my gut tells me "it'll be fine!".  :)

ThomasWaldmann:
any news here?

cculianu:
Yeah I really need this  It works perfectly.

Any chance we can get this merged very soon? I really really really need this to be standard for my open source project -- people use pyenv to build our app on MacOS.

cculianu:
openssl 1.0 sucks. 1.1 is the wave of the FUTURE!

cculianu:
@yyuu I beg you. :)

plugins/python-build/bin/python-build
```diff
@@ -1492,15 +1492,20 @@ has_broken_mac_openssl() {
 }
 
 use_homebrew_openssl() {
-  local ssldir="$(brew --prefix openssl 2>/dev/null || true)"
-  if [ -d "$ssldir" ]; then
-    echo "python-build: use openssl from homebrew"
-    export PKG_CONFIG_PATH="$ssldir/lib/pkgconfig/:${PKG_CONFIG_PATH}"
-    export CPPFLAGS="-I$ssldir/include ${CPPFLAGS}"
-    export LDFLAGS="-L$ssldir/lib ${LDFLAGS}"
+  local ssl11dir="$(brew --prefix openssl@1.1 2>/dev/null || true)"
```
yyuu: This lookup behaviour might be conflict with `openssl` lookup in `compute_sha2()`. But it would work sufficiently as long as there is some `openssl` available in `PATH` 🤔 

https://github.com/pyenv/pyenv/blob/v1.2.9/plugins/python-build/bin/python-build#L253

cculianu: I love you. ❤️

ThomasWaldmann: I just noticed the compute_sha2 works a bit differently than the compute_md5 right below it (and, assuming that it works, the latter does not need to use brew to locate openssl).

ThomasWaldmann: I also wondered why both compute_sha2 and compute_md5 try sha256sum and md5sum commands last (after trying openssl).

ThomasWaldmann: @yyuu ^^^

joshfriend:
@ThomasWaldmann thanks so much for this PR ❤️

PR #1307, opened 2019-03-22
Update documentation

alanyee:

### Prerequisite

### Description

I update the documentation to be more direct to the users' needs, especially new users.

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,21 +107,21 @@ operating system will do the following:
 When you execute a shim, pyenv determines which Python version to use by
 reading it from the following sources, in this order:
 
-1. The `PYENV_VERSION` environment variable (if specified). You can use
-   the [`pyenv shell`](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell) command to set this environment
-   variable in your current shell session.
+1. The `PYENV_VERSION` environment variable (if specified). You can use the
+   [`pyenv shell <version>`](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell)
+   command to set this environment variable in your current shell session.
 
 2. The application-specific `.python-version` file in the current
    directory (if present). You can modify the current directory's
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,21 +107,21 @@ operating system will do the following:
 When you execute a shim, pyenv determines which Python version to use by
 reading it from the following sources, in this order:
 
-1. The `PYENV_VERSION` environment variable (if specified). You can use
-   the [`pyenv shell`](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell) command to set this environment
-   variable in your current shell session.
+1. The `PYENV_VERSION` environment variable (if specified). You can use the
+   [`pyenv shell <version>`](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell)
```
native-api, line 111: The suggested formulation gives a wrong idea that the command can only be used in this way. (While there are actually two more.)

When referring to a command as a whole, they don't replicate its invocation syntax -- because that'd be unreadable and duplication.

This could be good if the text was formulated so that the code is presented not as representation of the entire command but as a _one possible invocation of it_ ("with `pyenv shell <version>`" instead of "with the `pyenv shell <version>` command"). But since the version can be selected with this command in other ways, too, claiming that this can only be done with this invocation is again misinformation.

(same goes for the other edits below)

libexec/pyenv-help
```diff
@@ -154,7 +154,7 @@ if [ -z "$1" ] || [ "$1" == "pyenv" ]; then
   print_summaries commands local global shell install uninstall rehash version versions which whence
   echo
   echo "See \`pyenv help <command>' for information on a specific command."
-  echo "For full documentation, see: https://github.com/pyenv/pyenv#readme"
+  echo "For full documentation, see: https://github.com/pyenv/pyenv/blob/master/COMMANDS.md"
```
native-api, line 157: This is the output of `pyenv help` -- i.e. the general help. So it should be referring to general documentation. Link to `COMMANDS.md` could probably be given, but separately.

native-api (changes requested):

Issue #1311, opened 2019-03-27
python-build does not properly find ssl library :: pyenv doctor fails caused python-build fail


AlyssonOliveira:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): GNU/Linux Debian, Independent
- [x] OS architecture (e.g. amd64): Independent
- [x] pyenv version: 1.2.7-59-gdae256f
- [x] Python version: 3.7.x
- [x] C Compiler information (e.g. gcc 7.3): gcc 8.3 and all
- [x] Please attach verbose build log as gist

```
pyenv doctor 
Cloning /home/username/.pyenv/plugins/pyenv-doctor/bin/.....
Installing python-pyenv-doctor...

BUILD FAILED (Debian testing using python-build 1.2.7-59-gdae256f)

Inspect or clean up the working tree at /tmp/python-build.20190326210727.25763
Results logged to /tmp/python-build.20190326210727.25763.log

Last 10 log lines:
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for rl_gnu_readline_p in -lreadline... yes
checking for readline/readline.h... no
checking for SSL_library_init in -lssl... no
configure: error: OpenSSL is not installed.
Problem(s) detected while checking system.

[...]

Issue #1268, opened 2019-01-14
[MacOS] It is currently impossible to use pyenv to install/build python that links against OpenSSL 1.1


cculianu:
### Description
- [x] Platform information: MacOSX 10.13 High Sierra
- [x] OS architecture: amd64
- [x] pyenv version: 1.2.9-2-g6309aaf
- [x] Python version: Any/All
- [x] C Compiler information: clang-1000.11.45.5 
- [ ] Please attach verbose build log as gist

---

Hi all,

I'm on OSX with Brew.

It's currently impossible to get pyenv to build & install a version of Python linked against OpenSSL 1.1.  Passing the arg "--with-openssl=" to Python via PYTHON_CONFIGURE_OPTS or CONFIGURE_OPTS doesn't have the intended effect as pyenv also inserts its own path to openssl (1.0). Any combination of this plus LDFLAGS + CFLAGS also doesn't take.  They show up on the command-line but the pyenv-specified flags are the ones that end up taking (the pyenv flags point to old openssl 1.0).

If I unistall openssl 1.0 and just leave 1.1 installed, pyenv insists on downloading OpenSSL 1.0 and compiling it privately.

The only way to do it is to "fool" pyenv and the python build process by uninstalling openssl1.0 and symlinking /usr/local/opt/openssl to /usr/local/opt/openssl@1.1. (While this "works", this is a hack and a nightmare for those of use that work on open source projects where our users expect to be able to reproducibly and easily deterministically build an app.)

The situation is pretty frustrating because OpenSSL 1.1 offers newer features that end up showing up in the Python SSL API's.  Here is 1 exampe: the `scrypt` function which is not available unless Python is linked against OpenSSL 1.1: https://docs.python.org/3/library/hashlib.html#hashlib.scrypt

**Suggestions**: Allow users to override the openssl options and/or have pyenv understand that OpenSSL1.1 is a thing and allow users to select it at build/install-time.

Not having access to OpenSSL 1.1 from within Python is very limiting!

Thanks in advance.

cculianu:
Update: One of the places where a patch could be added would be here:

https://github.com/pyenv/pyenv/blob/6309aaf295b05fece4abb52ba1000f98e40390e7/plugins/python-build/bin/python-build#L1494

Either it could:

1. Auto-detect openssl1.1 if it exists and prefer that or
2. Allow users to tell the build script to look for openssl1.1 with say some option like --use-openssl1.1 or something.

[...]

PR #1312, opened 2019-03-27
Update generated configuration files before run ./configure

AlyssonOliveira:

### Prerequisite
**Not applicable.**

**Not applicable.**

  - [1311](https://github.com/pyenv/pyenv/issues/1311)
  - [1184](https://github.com/pyenv/pyenv/issues/1184)
  - [1268](https://github.com/pyenv/pyenv/issues/1268)
  - [993](https://github.com/pyenv/pyenv/issues/993)

### Description
`python-build` plugin fails to find some dependency packages because it uses the default `configure` script,  so update the `configure` script and other generated scripts on target system to solve this issue by running autoreconf intead autoconf

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -791,7 +791,7 @@ build_package_standard() {
 }
 
 build_package_autoconf() {
-  { autoconf
+  { autoreconf
   } >&4 2>&1
 }
 


blueyed:
Does it fix/close all the referenced issues?

native-api (approved):
LGTM -- according to https://stackoverflow.com/questions/27285052/difference-between-autoconf-and-autoreconf, `autoreconf` is a higher-level tool that's preferred to `autoconf`` in general use.

However, `build_package_autoconf` is currently never used in any of installation scripts -- so this won't change anything.

PR #1326, opened 2019-04-07
openssl related improvements

ThomasWaldmann:
take this rather as an idea than a ready-to-use PR.

it's just 2 things i noticed when looking at that code that seemed strange:
- why is it invoking openssl before even trying sha256sum or md5sum? this commit is just moving the openssl based elif-block to the end.
- why invoke the brew openssl even for trivial stuff like `openssl dgst -sha256` (i found the historic reason, see commit comment - the question now is whether that is still needed). the code after this commit is now closer to the md5 openssl based code in the other function.

not tested on macOS yet.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -253,8 +253,7 @@ compute_sha2() {
     output="$(sha256sum -b)" || return 1
     echo "${output%% *}"
   elif type openssl &>/dev/null; then
-    local openssl="$(command -v "$(brew --prefix openssl 2>/dev/null || true)"/bin/openssl openssl | head -1)"
-    output="$("$openssl" dgst -sha256 2>/dev/null)" || return 1
+    output="$(openssl dgst -sha256)" || return 1
     echo "${output##* }"
   else
     return 1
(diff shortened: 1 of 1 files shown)


yyuu:
I'd _guess_, the reason why it is invoking `openssl dgst` prior to `sha256sum` is that macOS doesn't have the latter by default. The answer to your second question is, it is implemented to avoid use macOS's default `/usr/bin/openssl`, I guess. Anyway it's passing 2 `openssl` to `command -v`, it'll fallback to _ordinary_ `openssl` if `brew` is not available. Both changes are pretty much macOS stuffs, but I don't want to make changes to them to keep differences between https://github.com/rbenv/rbenv as smaller as possible

ThomasWaldmann:
guess this code is used for all supported OSes, not just macOS. on linux, md5sum and sha256sum are quite frequently there and checking them first would avoid any openssl / brew troubles (and the fact that it might invoke some brew command on non macOS, not sure what that could do :).

i know macOS default openssl is old crap, but for very basic stuff like md5 and sha256 it might be good enough.

understandable if you don't want to depart from rbenv code.

if anybody thinks this might be interesting for rbenv upstream, feel free to submit it there (i won't).

PR #1327, opened 2019-04-08
Import ruby-build 20190401

yyuu:

### Prerequisite
  * This is core feature and cannot be implemented as some hook
  * This is import from the upstream
  - There's no existing issue for this

### Description
    * I've imported recent changes from https://github.com/rbenv/ruby-build

### Tests
    * I have already tweaked/fixed broken tests after the import


yyuu:
Hmm, somehow travis-ci isn't working as I was expecting. I'm going to merge this once after I fixed CI issue....

Makefile
```diff
@@ -8,8 +8,18 @@ test: bats
 	PATH="./bats/bin:$$PATH" test/run
 	cd plugins/python-build && $(PWD)/bats/bin/bats $${CI:+--tap} test
 
-test-build: bats
-	cd plugins/python-build && $(PWD)/bats/bin/bats $${CI:+--tap} test/build
+PYTHON_BUILD_ROOT := $(CURDIR)/plugins/python-build
```
chrahunt: Doing this in make instead of bats seems a bit harder to extend in the future. Imagine all the open issues we have where the failed build is only evident after testing the compiled result.

.travis.yml
```diff
@@ -36,9 +36,6 @@ install: git clone --depth 1 https://github.com/bats-core/bats-core.git bats
 # Default for auto-generated jobs.
 script: make test-build
 
-after_script: |
```
chrahunt: Without this and bats now every CI run has a lot of output printed instead of being in a collapsible section in the Travis CI interface - compare [this](https://travis-ci.org/pyenv/pyenv/jobs/520866465) (post-change) and [this](https://travis-ci.org/pyenv/pyenv/jobs/517222222) (pre-change). In the latter case some 9000 lines are hidden by default, which is nice.

plugins/python-build/bin/python-build
```diff
@@ -925,6 +936,13 @@ fix_jruby_shebangs() {
   done
 }
 
+build_package_truffleruby() {
```
chrahunt: Can be removed?

plugins/python-build/bin/python-build
```diff
@@ -1561,15 +1579,38 @@ build_package_mac_openssl() {
...
-  "$RUBY_BIN" -e 'begin
-    require "openssl"
-  rescue LoadError
-    $stderr.puts "The Ruby openssl extension was not compiled. Missing the OpenSSL lib?"
-    $stderr.puts "Configure options used:"
-    require "rbconfig"; require "shellwords"
-    RbConfig::CONFIG.fetch("configure_args").shellsplit.each { |arg| $stderr.puts "  #{arg}" }
-    exit 1
-  end' >&4 2>&1
+  "$RUBY_BIN" -e '
```
chrahunt: We should do the Python equivalent `import _ssl`, but it can be a separate issue.

plugins/python-build/bin/python-build
yyuu: At least I'm pretty sure it's some unused code and we can safely remove it. I ended up to import them as well to use it as placeholder when diffing between ruby-build's impl...

plugins/python-build/bin/python-build
yyuu: JFYI, in fact, this is unused code from python-build as we're using `verify_python_module` instead.

https://github.com/pyenv/pyenv/blob/v1.2.11/plugins/python-build/bin/python-build#L1846

.travis.yml
yyuu: Ah, to be honest, I didn't think much about Travis CI's interface 😞  At least there would be something we can/should improve.

In fact, at that time what I wanted to do was to avoid travis's build timeout after non-output for 10 minutes. There could be even more better workaround for the purpose at least.

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

PR #1344, opened 2019-05-22
Remove extra space from pyenv init command

kevin-andrews:
### Description
There's an extra space before eval that is retained when copied and pasted into an editor. While not a common use case, it's hard to spot and sourcing your profile will result in the error `command not found: eval` so it's worth fixing.


blueyed:
Isn't there still an extra space there?!

In general it would be better to use rbenv's (upstream) README (not caused by you though).

native-api:
No longer relevant after https://github.com/pyenv/pyenv/pull/1920

Issue #1347, opened 2019-05-25
is putting "eval "$(pyenv init -)"" in ~/.bash_profile really a good suggestion?


AmaiKinono:
It's suggested in README.md to put `eval "$(pyenv init -)"` (and environment variables) in `~/.bash_profile`, which may be not reasonable.

I am a (manjaro) Linux user. As far as I am concerned, `~/.bash_profile` is not sourced when logging in under X (which means, under a graphical environment). Here's a little experiment:

- put `export USING_TTY=1` in `~/.bash_profile`, reboot the computer.
- open a terminal emulator, run `$ echo $USING_TTY`, nothing is printed.
- switch to a text tty, run `$ echo $USING_TTY`, "1" is printed.

Most desktop environments sources `~/.profile` when logging in, so that should be the right place for Linux users to put environment variables and `eval $(pyenv init -)` in. When using text tty, only `~/.bash_profile` is sourced, but it's a common practice to source `~/.profile` and `~/.bashrc` inside `~/.bash_profile`, so this would make pyenv works both on desktop environment and text tty.

Please tell me if I am wrong.

Btw, here's a little problem. I myself put these in `~/.profile`, and most of the functionalities work well under a desktop environment. `$ which python` gives me `~/.pyenv/shims/python`, which should be a sign that I'm doing it right, but `$ pyenv shell` gives `pyenv: shell integration not enabled. Run 'pyenv init' for instructions.`. Does anybody know why this happens?

*Update:* I found I have to put `export -f pyenv` after `eval "$(pyenv init -)"` to export the function for sub-shells to use. This fixes the above problem.

georgalis:
Good observation about all the different ways to bootstrap your environment! I've supported many shells (bash, bourne, zsh, ksh, ash, etc) on different OS (Linux, BSD, Darwin) at different sites. I've also created a tool set of command line functions and aliases. Getting your shell all setup with tuned and bug free preferences can take forever and a week and it's the last thing you want to fix when you've got other work to do. When users ask for help they generally don't want you refactoring their multiple shell conf files either.

Here's the standard I use when I'm setting up a system or my own accounts. All of my functions and aliases go into .profile (including all the stuff that has been added to user's bashrc, bash_profile, etc). In .profile, I use bourne compatible scripting to determine the OS, shell, and installed programs, with conditionals to set only appropriate environment, being careful to branch early for non interactive shells (so ssh works as expected)
`/usr/bin/tty -s || return # the following for interactive sessions only`

[...]

PR #1359, opened 2019-06-21
Update shell configuration info in README

AmaiKinono:
- `~/.bash_profile` may not be sourced in desktop environment. When this happenes, `~/.profile` can be used instead.
- The `pyenv` function should be exported if `eval $(pyenv init -)` is put into `~/.profile` or `~/.bash_profile`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,12 +190,16 @@ easy to fork and contribute any changes back upstream.
 
 2. **Define environment variable `PYENV_ROOT`** to point to the path where
    pyenv repo is cloned and add `$PYENV_ROOT/bin` to your `$PATH` for access
-   to the `pyenv` command-line utility.
-
+   to the `pyenv` command-line utility. For example, if you would like to put
+   environment variables in `~/.bash_profile`:
     ```sh
     $ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
     $ echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
     ```
+    
+    - **Linux note**: Your desktop environment may not source `~/.bash_profile`.
+      If this happens, you may want to use `~/.profile` instead, since it's usually
+      sourced by the display manager.
     - **Zsh note**: Modify your `~/.zshenv` file instead of `~/.bash_profile`.
     - **Ubuntu and Fedora note**: Modify your `~/.bashrc` file instead of `~/.bash_profile`.
     - **Proxy note**: If you use a proxy, export `http_proxy` and `HTTPS_PROXY` too.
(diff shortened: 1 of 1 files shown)


georgalis:
Thanks! I would be more specific about when to use `export -f pyenv` because readers of the doc won't know whether "may need" applies to them or when it might apply to them in the future. Explaining when export must be used, will remove confusion.

AmaiKinono:
My thought is when put `eval $(pyenv init -)` in `~/.profile` or `~/.bash_profile`, `export -f pyenv` is needed. But I didn't say that because I am not sure if this is true for most environments, so I simply say "if `pyenv shell` or `pyenv rehash` doesn't work, you should do this." I'd like to hear your thought about this.

georgalis:
My thought is shell configuration schemes are a mess and that has a lot to do with why users so often have broken environments or are confused about their setup. While the invocation section of shell man pages seem carefully written enough, there is no explanation of the requirements driving the complexity.

> If bash is invoked with the name sh, it tries to mimic the startup behavior of historical versions of sh as closely as possible, while conforming  to  the  POSIX standard as well.  When invoked as an interactive login shell, or a non-interactive shell with the --login option, it first attempts to read and execute commands from /etc/profile and ~/.profile, in that order.  The --noprofile option may be used to inhibit this behavior.  When invoked as an  interactive  shell  with  the name  sh,  bash  looks  for the variable ENV, expands its value if it is defined, and uses the expanded value as the name of a file to read and execute.  Since a shell invoked as sh does not attempt to read and execute commands from any other startup files, the --rcfile option  has  no  effect.   A  non-interactive  shell invoked with the name sh does not attempt to read any other startup files.  When invoked as sh, bash enters posix mode after the startup files are read.

The only case for different shell configurations I know of is interactive or non-interactive; so, I source .profile from .bashrc, then in .profile I use a conditional return if non-interactive, `test -t 0 || return 0` before setting interactive variables, aliases and functions. This way I have one file for all my shell configurations, and I can use it on all the different systems I use.

I use additional branches within that file to detect OS, installed software and the specific shell, so aliases, functions and environmental configurations only get set when appropriate.

I source ~/.profile from all of my shell rc files (zsh, ksh, bash, etc) to keep all of my configuration in one file. That file checks if it is sourced from a non-interactive shell and returns without error if so, which keeps scp (and anything else using the shell programmatically) from breaking.

The thing to keep in mind is shell environment setup is not by design, it's by evolution. There probably isn't a good answer for everyone today. I like the idea of exporting functions from profile, it works when you use run control (rc) files for run control and profile for interactive function setup, as best I can tell. For me, I'll source my .profile from run control (rc files) and do my best to test the environment before setting anything that isn't universal (eg, only set zsh features in zsh, etc).

As for what is best in pyenv docs, not sure, but I would avoid anything like 'you may need' in documentation or instructions about what you actually need. Better to state the circumstances you can think of, eg

- Put `eval $(pyenv init -)` where you define interactive functions, such as `.profile` or `.bash_profile` 
- You will need to `export -f pyenv` to access the function from a sub-shell, such as a desktop environment.

Nobody wants to be definitive about how anybody in the world should always setup their system, and there is usually more than one solution. But if you explain how to deal with the specific problem cases you can think of, that's good documentation and the elucidation sometimes identifies a better coding solution too.

BTW - my favorite shell is ksh but the good version is only available on NetBSD, so I usually use bash on Mac and Linux.

AmaiKinono:
It's updated.And thank @georgalis for sharing your setup! It makes a lot of sense,, and maybe I should do it in the same way.

blueyed:
Would be great to get this via rbenv's (upstream) README: https://github.com/rbenv/rbenv/blob/master/README.md

There are changes in this regard IIRC, so it would be good to not derive further from it, but get it straight via/from there.

README.md
```diff
@@ -190,30 +190,42 @@ easy to fork and contribute any changes back upstream.
...
    Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
-   configuration file since it manipulates `PATH` during the initialization.
+   configuration file since it manipulates `PATH` during the initialization. For example:
     ```sh
     $ echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
     ```
+    
+    If you put this in the run control file (like `~/.bashrc`), this will work fine; otherwise (like if you
+    use `~/.bash_profile` or `~/.profile` instead, and it is not sourced in `~/.bashrc`), you need to put
+    `export -f pyenv` after this to access the function from a sub-shell.
```
nchammas: Should/could `export -f pyenv` instead somehow be part of `pyenv init -`? That would avoid adding any extra steps for users to do.

(FWIW, I found this tip via #1347. I was having the same issue with sub-shells and your suggestion there fixed things for me. Thanks!)

native-api:
Superseded by https://github.com/pyenv/pyenv/pull/1920

Issue #1360, opened 2019-06-25
Discussion: Determinism of *conda builds


clbarnes:
Context: I'm writing a script to automate populating miniconda versions in pyenv-build (which are currently several years behind). It likely won't be hard to generalise it to anaconda, too.

It would be nice if calling `pyenv install miniconda-{version}` produced the same result every time. Pyenv obviously can't guarantee that the upstream artifacts won't change, but there are two situations where pyenv could, but does not, try to mitigate potential issues:

1. Fix versions (e.g. `4.3.30.1`, which is in the script for `4.3.30`). These tend to be OS/arch specific.
2. miniconda-latest

Miniconda-latest is a problem mainly because the resulting version ends up named `miniconda-latest` on the user's machine. Without the context of when that was installed, the user doesn't actually know what version it is. It would add complexity for the maintainers, but it may be more intuitive for the users if `pyenv install miniconda-latest` did not just grab the "Miniconda-latest" artifact, but was treated specially, like the 3.x-latest aliases. Installing miniconda-latest could find the latest available release; activating miniconda-latest could find the latest installed release.

IMHO, the current miniconda-latest adds more potential for confusion than it does convenience. Even if we don't choose to do *-latest handling ourselves, I would consider a policy which recommends they not be added to the pyenv-build list.

Fix versions are a bit more tricky. Obviously it's best to encourage users to use the patched version, but for every obsolete version of python, there's some person or company intentionally trying to use it. Determinism would be great, but introducing a bunch of OS/arch-specific fix versions is likely to be more confusing than it's worth.

There isn't necessarily a solution; just a strategy to be discussed. What do you think?

MrPowers:
@clbarnes - thanks for the great contributions to this library and sorry in advance for the somewhat off topic comment.

Looks like [the current miniconda code is written in Python](https://github.com/pyenv/pyenv/blob/df9fa1dc30b6448ef8605e2c2d4dfc2a94d6a35d/plugins/python-build/scripts/add_miniconda.py).  It's the only Python file in this project.

From the README: "pyenv was made from pure shell scripts. There is no bootstrap problem of Python."

I'm a bit surprised to see conda stuff in this repo because I thought the sole purpose of this project was to switch between Python versions.  

Let me know your thoughts.  I am a Python noob, so feel free to educate me on why it's important to have Python code in this repo :)

clbarnes:
[...]

PR #1361, opened 2019-06-25
Update miniconda builds

clbarnes:
We're pretty behind on miniconda versions. This includes all new major.minor.patch versions since 4.3.30, as well as a script to automatically populate new miniconda versions in the future.

- Ignores releases below 4.3.30
- Ignores "-latest" releases
- Ignores sub-patch releases if that major.minor.patch already exists
  - But otherwise, takes the latest sub-patch release for given OS/arch
- Assumes all miniconda3 releases > 4.3.30 default to python 3.6 (correct at time of writing)

### Description

- miniconda[23] versions 4.3.31 through 4.6.14.

### Tests

I am not sure how best to test adding new builds; is there anything more ergonomic than just trying to build each one, one at a time?

See also #1360 for a discussion on miniconda-latest.

### DIFF
--- a/plugins/python-build/scripts/add_miniconda.py
+++ b/plugins/python-build/scripts/add_miniconda.py
@@ -1,0 +1,262 @@
+#!/usr/bin/env python3.7
+"""Script to add non-"latest" miniconda releases.
+
+- Ignores releases below 4.3.30.
+- Ignores sub-patch releases if that major.minor.patch already exists
+  - But otherwise, takes the latest sub-patch release for given OS/arch
+- Assumes all miniconda3 releases default to python 3.6 (correct at time of writing)
+
+Use -d for dry run.
+"""
+import textwrap
+from argparse import ArgumentParser
+from collections import defaultdict
+from enum import Enum
+from functools import total_ordering
+from pathlib import Path
+from typing import NamedTuple, List, Optional, DefaultDict, Dict
+
+import requests_html
+
+CONDA_REPO = "https://repo.anaconda.com"
+MINICONDA_REPO = CONDA_REPO + "/miniconda"
+# ANACONDA_REPO = CONDA_REPO + "/archive"
+
+install_script_fmt = """
+case "$(anaconda_architecture 2>/dev/null || true)" in
+{install_lines}
+* )
+  {{ echo
+    colorize 1 "ERROR"
+    echo ": The binary distribution of Miniconda is not available for $(anaconda_architecture 2>/dev/null || true)."
+    echo
+  }} >&2
+  exit 1
+  ;;
+esac
+""".lstrip()
+
+install_line_fmt = """
+"{os}-{arch}" )
+  install_script "Miniconda{suffix}-{version_str}-{os}-{arch}" "{repo}/Miniconda{suffix}-{version_str}-{os}-{arch}.sh#{md5}" "miniconda" verify_{py_version}
+  ;;
+""".strip()
+
+here = Path(__file__).resolve()
+out_dir: Path = here.parent.parent / "share" / "python-build"
+
+
+class StrEnum(str, Enum):
+    """Enum subclass which members are also instances of str
+    and directly comparable to strings. str type is forced at declaration.
+
+    Adapted from https://github.com/kissgyorgy/enum34-custom/blob/dbc89596761c970398701d26c6a5bbcfcf70f548/enum_custom.py#L100
+    (MIT license)
+    """
+    def __new__(cls, *args):
+        for arg in args:
+            if not isinstance(arg, str):
+                raise TypeError('Not text %s:' % arg)
+
+        return super(StrEnum, cls).__new__(cls, *args)
+
+    def __str__(self):
+        return str(self.value)
+
+
+class SupportedOS(StrEnum):
+    LINUX = "Linux"
+    MACOSX = "MacOSX"
+
+
+class SupportedArch(StrEnum):
+    PPC64LE = "ppc64le"
+    X86_64 = "x86_64"
+    X86 = "x86"
+
+
+class Suffix(StrEnum):
+    TWO = "2"
+    THREE = "3"
+
+
+class PyVersion(StrEnum):
+    PY27 = "py27"
+    PY36 = "py36"
+    PY37 = "py37"
+    PY38 = "py38"
+    PY39 = "py39"
+
+    def version(self):
+        first, *others = self.value[2:]
+        return f"{first}.{''.join(others)}"
+
+    def version_info(self):
+        return tuple(int(n) for n in self.version().split('.'))
+
+
+@total_ordering
+class VersionStr(str):
+    def info(self):
+        return tuple(int(n) for n in self.split('.'))
+
+    def __eq__(self, other):
+        return str(self) == str(other)
+
+    def __lt__(self, other):
+        if isinstance(other, VersionStr):
+            return self.info() < other.info()
+        raise ValueError("VersionStr can only be compared to other VersionStr")
+
+    @classmethod
+    def from_info(cls, version_info):
+        return VersionStr('.'.join(str(n) for n in version_info))
+
+    def __hash__(self):
+        return hash(str(self))
+
+
+class MinicondaVersion(NamedTuple):
+    suffix: Suffix
+    version_str: VersionStr
+
+    @classmethod
+    def from_str(cls, s):
+        miniconda_n, ver = s.split('-')
+        return MinicondaVersion(Suffix(miniconda_n[-1]), VersionStr(ver))
+
+    def to_filename(self):
+        return f"miniconda{self.suffix}-{self.version_str}"
+
+    def default_py_version(self):
+        if self.suffix == Suffix.TWO:
+            return PyVersion.PY27
+        else:
+            # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-python.html
+            return PyVersion.PY36
+
+    def with_version_triple(self):
+        return MinicondaVersion(self.suffix, VersionStr.from_info(self.version_str.info()[:3]))
+
+
+class MinicondaSpec(NamedTuple):
+    version: MinicondaVersion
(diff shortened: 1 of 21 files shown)


mariusvniekerk:
Anything remaining on this @yyuu before we can merge?

plugins/python-build/scripts/add_miniconda.py
```diff
@@ -0,0 +1,261 @@
...
+
+class MinicondaSpec(NamedTuple):
+    version: MinicondaVersion
+    os: SupportedOS
+    arch: SupportedArch
+    md5: str
+    py_version: Optional[PyVersion] = None
+
+    @classmethod
+    def from_filestem(cls, stem, md5, py_version=None):
```
yyuu: Please consider using sha256sum in lieu of md5sum. The current code actually can validate MD5 checksum, however, it's just left for backward compatibility and no new code should use the legacy hashing algo.

plugins/python-build/scripts/add_miniconda.py
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/env/python3.7
```
yyuu: It might be little funny if a Python script packaged with pyenv is configured to skip using pyenv 😛

clbarnes: Oops, that shouldn't have the slash before python.

I don't think this necessarily skips using pyenv though - wouldn't `/usr/bin/env python3.7` use whichever python3.7 is first in the path, including whatever pyenv shims in?

plugins/python-build/scripts/add_miniconda.py
clbarnes: I used md5 because that's the checksum they have on the installer archive. If there's an equally accessible sha256 checksum I'll look into it.

clbarnes:
I still haven't tested most of the builds this adds - is there a good strategy for that?

Any thoughts on the sub-patch builds?

plugins/python-build/scripts/add_miniconda.py
yyuu: It actually depends on how `$PATH` is configured for the caller's shell, however, having `/usr/bin/env python3.7` in shebang would work sufficiently for the purpose.

clbarnes:
- rebased on master
- fixed shebang of script
- re-ran script to add miniconda[23]-4.7.10

mariusvniekerk:
@yyuu What more is needed here?

clbarnes:
I need to update it to use python 3.7 for recent miniconda versions, and test all of them. Is there a good strategy for testing large numbers of builds?

clbarnes:
I have rebased this on master and fixed the py37 issue.

The review comment from @yyuu can't be addressed, as the usage of md5sum is anaconda's choice, not mine.

All this needs now is testing, but I can't find any information on how to test installed python versions without my own (mainline) pyenv installation interfering; see #1445 .

`bats` testing gets the same two failures on master and on this branch:

```
 ✗ sys.executable with system version (#98)                                                                                                     
   (from function `flunk' in file test/test_helper.bash, line 42,                                                                               
    from function `assert_success' in file test/test_helper.bash, line 47,                                                                      
    in test file test/exec.bats, line 90)                                                                                                       
     `assert_success "${system_python}"' failed                                                                                                 
   command failed with exit status 1                                                                                                            
 ✗ /tmp/pyenv.NZj/root/shims:/home/barnesc/code/pyenv/test/libexec:/home/barnesc/code/pyenv/test/../libexec:/tmp/pyenv.NZj/bin:/usr/bin:/bin:/us
r/sbin:/sbin:/usr/local/bin is not modified with system Python                                                                                  
   (from function `flunk' in file test/test_helper.bash, line 42,                                                                               
    from function `assert_success' in file test/test_helper.bash, line 47,                                                                      
    in test file test/exec.bats, line 102)                                                                                                      
     `assert_success "$PATH"' failed                                                                                                            
   command failed with exit status 1   
```

clbarnes:
Rebased on master again. Still waiting for documentation or advice on how to test new versions.

clbarnes:
Rebasing this on PR #1548 locally allowed me to test a handful of these very easily and in complete isolation.

plugins/python-build/scripts/add_miniconda.py
```diff
@@ -0,0 +1,265 @@
...
+        spec.version.version_str.info(),
+        spec.version.suffix.value,
+        spec.os.value,
+        spec.arch.value,
+    )
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.add_argument("-d", "--dry_run", action="store_true")
```
yyuu: trivial - I would _feel_ that `--dry-run` is more intuitive than underscore name 🤔

plugins/python-build/scripts/add_miniconda.py
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env python3.7
```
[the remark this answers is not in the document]
yyuu: It would be nice if there's some usage example of this script. In fact, just running the script without giving any arguments will let the script to update the python-build's build definitions for miniconda based on the contents of Miniconda's download page, however, it's not very obvious without looking into the script itself. Since this is just some sort of utility script for maintainers, it doesn't need to be some individual documentation file, however, it's nice if there's some practical usage example is in somewhere (e.g. show message on `--help`, etc.)

plugins/python-build/scripts/add_miniconda.py
yyuu: I still prefer to use SHA256 in general as MD5 is not considered as safe since awhile ago. However, I can understand that it's not possible to generate SHA256 without downloading remote contents in case the official doesn't provide it. It might be nice if generated file will have some shell script comment like `# FIXME: please replace this md5sum by sha256sum` maybe...? 🤔

plugins/python-build/scripts/add_miniconda.py
```diff
@@ -0,0 +1,265 @@
...
+    return (
+        spec.version.version_str.info(),
+        spec.version.suffix.value,
+        spec.os.value,
+        spec.arch.value,
+    )
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
```
yyuu: trivial - it's nice if argparse can take `--verbose` or `--debug` to enable debug logging.

plugins/python-build/scripts/add_miniconda.py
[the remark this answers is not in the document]
clbarnes: Yep - I think at time of writing 3.8 wasn't out, and 3.7 was pretty new so I wanted to be explicit about it. Now I'm happy to just specify 3 in the shebang and document that it's 3.7+.

Usage example is a good idea. I'll write a module docstring and have that reported by the help message.

plugins/python-build/scripts/add_miniconda.py
clbarnes: Yes, I dimly recall some time in the past argparse not being able to handle dashes in the middle of an option name but that doesn't seem to be the case now.

plugins/python-build/scripts/add_miniconda.py
clbarnes: Sure, I can add more verbose logging.

clbarnes:
I think I'll add a readme with a usage example etc.

muendelezaji:
It appears there might be a problem in the merged code - there are duplicate architecture entries in the [miniconda2-4.7.12](https://github.com/pyenv/pyenv/pull/1361/files/81e797be12e86e70bf5e332b3ac15512d65ad950#diff-e51d5e7af8a59099804d79c11541a472) and [miniconda3-4.7.12](https://github.com/clbarnes/pyenv/blob/81e797be12e86e70bf5e332b3ac15512d65ad950/plugins/python-build/share/python-build/miniconda3-4.7.12) python-build files. This manifests as the 4.7.12 shadowing the 4.7.12.1 version i.e. unable to install 4.7.12.1.

Is that intended, or should there be a separate 4.7.12.1 python-build file?

clbarnes:
As I understand it, pyenv does not and has never handled distinguishing subpatch versions for the \*condas. As such, this script picks the latest available subpatch version when any given patch version is first picked up. It doesn't overwrite old versions when new subpatch versions are introduced, partially for reasons specified in #1360, partially to prevent it being destructive. Something is very wrong with anaconda's numbering scheme if anybody's workflow would be broken by having a newer subpatch version than they expected. 

I'm open to supporting subpatch versions and it would probably be trivial to update the script to do so, but this is a bigger new feature than I wanted to introduce with this PR. Other alternatives for handling subpatch versions are welcome in #1360.

muendelezaji:
> As such, this script picks the latest available subpatch version when any given patch version is first picked up.

Just to clarify, I also expected this behaviour on a fresh install when not overwriting a previous miniconda. But the order in the shell script means that the case statement returns the first match (i.e. 4.7.12) and never gets to the latest subpatch version.

For example, from [miniconda3-4.7.12](https://github.com/clbarnes/pyenv/blob/81e797be12e86e70bf5e332b3ac15512d65ad950/plugins/python-build/share/python-build/miniconda3-4.7.12)
```shell
# ...
# matches this first
"Linux-x86_64" )
  install_script "Miniconda3-4.7.12-Linux-x86_64" "https://repo.anaconda.com/miniconda/Miniconda3-4.7.12-Linux-x86_64.sh#0dba759b8ecfc8948f626fa18785e3d8" "miniconda" verify_py37
  ;;

# ...
# never gets here
"Linux-x86_64" )
  install_script "Miniconda3-4.7.12.1-Linux-x86_64" "https://repo.anaconda.com/miniconda/Miniconda3-4.7.12.1-Linux-x86_64.sh#81c773ff87af5cfac79ab862942ab6b3" "miniconda" verify_py37
  ;;

# ...
```

So the question was, is this deliberately written that way, should they be swapped around, or any other approach (removing subpatch completely)?

clbarnes:
Yes, you're right. That wasn't intentional. I'll see what I can do.

Issue #1364, opened 2019-06-29
Setting an alias for .python-version 


aiguofer:
One of my favorite features of pyenv is the ability to configure multiple versions and/or virtualenvs in my `.python-version` either locally or global. This is really useful for a variety of scenarios.

The only problem is that this doesn't play nice with tools that display your current virtualenv or pyenv version. For example, powerlevel9k and [this module](https://github.com/aiguofer/pyenv.el) I tweaked for Emacs will concat all the lines with ":". This looks horrible when you have multiple versions set up.

It would be awesome if each `.python-version` could have its own alias (and maybe use a `global` alias for the global version). Maybe the first line could be `<name>:alias` and that could be ignored for determining which versions to use but tools could use it to determine the name to display.

Thoughts?

aiguofer:
I took a stab at what this could look like. Looking around, it seems that `pyenv-version-name` was the ideal candidate for this, however it seems that most everything uses this to determine the actual versions. I decided to instead create a separate command that actually shows the display name. This could in turn be used for tools that want to display the version using `pyenv version-display-name`.

Obviously, there should be other commands for setting these aliases, but I thought I'd just present this for now.

aiguofer:
I ended up creating a plugin instead so I'll go ahead and close this. 

See https://github.com/aiguofer/pyenv-version-alias

Any feedback would be great!

PR #1365, opened 2019-06-29
Version alias

aiguofer:
This PR addresses #1364. The goal is to be able to display an alias for the local or global python version so that tools that display your pyenv version can use that instead.

I could try to turn this into a plugin if it doesn't fit into the ideas for core functionality. Please let me know what you think, and help me turn it into a module if need be.


joshfriend:
closing since you've made this into a plugin and closed your original issue.

PR #1374, opened 2019-07-18
macOS installation steps: define PYENV_ROOT variable.

jkukul:
During homebrew installation, the `PYENV_ROOT` variable is neither set at any point, nor any instructions are given to set it manually.

Without `PYENV_ROOT` environment variable, e.g. `pipenv` cannot properly detect that `pyenv` is installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,14 @@ package manager for macOS.
 
 To upgrade pyenv in the future, use `upgrade` instead of `install`.
 
-Then follow the rest of the post-installation steps under [Basic GitHub Checkout](https://github.com/pyenv/pyenv#basic-github-checkout) above, starting with #3 ("Add `pyenv init` to your shell to enable shims and autocompletion").
+Then:
+1. Define environment variable `PYENV_ROOT`:
+
+```
+$ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
+```
+
+2. Follow the rest of the post-installation steps under [Basic GitHub Checkout](https://github.com/pyenv/pyenv#basic-github-checkout) above, starting with #3 ("Add `pyenv init` to your shell to enable shims and autocompletion").
 
 ### Advanced Configuration
 


blueyed:
Should be checked with the upstream readme probably: https://github.com/rbenv/rbenv/blob/master/README.md

If it's not in there, it should be added there probably first.

Merging the README would be good in general I think.

jkukul:
I'm closing this one, because after another look I realised that current instructions are sufficient. 

This part of the instruction:

> Then follow the rest of the post-installation steps under [Basic GitHub Checkout](https://github.com/pyenv/pyenv#basic-github-checkout) above, starting with #3 ("Add `pyenv init` to your shell to enable shims and autocompletion").

effectively makes users to add `pyenv init -` to their shell and that sets the `PYENV_SHELL` env variable. This is enough to make `pipenv` properly detect `pyenv` installation, `PYENV_ROOT` isn't actually needed.

parkerbxyz:
👋 @jkukul I also ran into the same problem you did when I first installed `pyenv` and `Pipenv` and would have opened my own pull request with the same proposed changes had you not already done so. 

I have the following in my `.zshrc` file:

```sh
# pyenv (https://github.com/pyenv/pyenv)
if command -v pyenv 1>/dev/null 2>&1; then
  # define environment variable
  # enable shims and autocompletion
  eval "$(pyenv init -)"
fi
```

I tried removing `export PYENV_ROOT="$HOME/.pyenv"` and reloading my shell to see what would happen.

Here's what happened when I tried to create a virtual environment with `Pipenv` after reloading:

```sh
Using /shims/python (None) to create virtualenv…
⠙ Creating virtual environment...The path /shims/python (from --python=/shims/python) does not exist
```

When I added `export PYENV_ROOT="$HOME/.pyenv"` back to my `.zshrc` file, reloaded my shell, and tried to create a virtual environment with `Pipenv` again, it worked: 

```sh
Using /Users/username/.pyenv/shims/python (3.7.4) to create virtualenv…
⠧ Creating virtual environment...Using base prefix '/Users/username/.pyenv/versions/3.7.4'
New python executable in /Users/username/.local/share/virtualenvs/python-Ms9yNXIx-/Users/username/.pyenv/shims/python/bin/python
Installing setuptools, pip, wheel...
done.
Running virtualenv with interpreter /Users/username/.pyenv/shims/python

✔ Successfully created virtual environment! 
```

This leads me to believe that `PYENV_ROOT` may still be necessary (at least in my case) and it may still be worth updating the instructions to help others others in the same boat.

Do you see something wrong with the way I've set things up or the way I'm testing this? I know I may be misunderstanding something here and I'm open to feedback.

PR #1380, opened 2019-08-11
fast installation steps: define PYENV_PYTHON_MIRROR_URL variable.

Puins:
In China,source url download speed is slow, add third-party open source url.

Add PYENV_PYTHON_MIRROR_URL variable ,Improving download speed.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -480,7 +480,13 @@ reuse_existing_tarball() {
 }
 
 download_tarball() {
-  local package_url="$1"
+  #replaces python sources url
+  if [ -n "$PYENV_PYTHON_MIRROR_URL" ]; then
+      local package_url="$(echo "$1" | sed -r "s#(.*//www.python.org/ftp/python)*#$PYENV_PYTHON_MIRROR_URL#")"
+  else
+      local package_url="$1"
+  fi
+
   [ -n "$package_url" ] || return 1
 
   local package_filename="$2"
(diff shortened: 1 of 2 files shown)


hongyi-zhao:
Doesn't take effect, see following:

```
$ env PYENV_PYTHON_MIRROR_URL=https://cdn.npm.taobao.org/dist/python pyenv install -v 2.7.7
/tmp/python-build.20200326150107.13965 ~/.repo/github.com/pyenv/pyenv.git
Downloading Python-2.7.7.tar.xz...
-> https://www.python.org/ftp/python/2.7.7/Python-2.7.7.tar.xz
^Cerror: failed to download Python-2.7.7.tar.xz

```

Mark24Code:
good job

plugins/python-build/bin/python-build
```diff
@@ -480,7 +480,13 @@ reuse_existing_tarball() {
 }
 
 download_tarball() {
-  local package_url="$1"
+  #replaces python sources url
+  if [ -n "$PYENV_PYTHON_MIRROR_URL" ]; then
+      local package_url="$(echo "$1" | sed -r "s#(.*//www.python.org/ftp/python)*#$PYENV_PYTHON_MIRROR_URL#")"
```
joshfriend, line 485: This only works for CPython versions.

pyenv can install numerous other types including Jython, PyPy, Stackless, miniconda, etc. and this solution will fail silently if you try to install one of those versions from a mirror

Perhaps a better solution would be one where the entire URL and base path of the python source/binary is replace

joshfriend:
I would rather the original mirror mechanism be fixed as opposed to adding in an entirely new one

PR #1386, opened 2019-08-29
Add troubleshooting section; add Q/A about Python dependencies

britho:
### Description

When using pyenv on crostini, I noticed that the installation of at least Python 3.7.4 failed due to the missing core libraries.  I thought it might be a good opportunity to capture the libraries needed in order for Python to be used at least in the README.  

### Disclaimer

I'm new to using pyenv so I need to investigate how exactly it works, but if it doesn't install these libraries if they are missing then I think there is room for a feature.

Let me know what you think, and thank you for providing such a wonderful tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ This project was forked from [rbenv](https://github.com/rbenv/rbenv) and
     * [Homebrew on macOS](#homebrew-on-macos)
     * [Advanced Configuration](#advanced-configuration)
     * [Uninstalling Python Versions](#uninstalling-python-versions)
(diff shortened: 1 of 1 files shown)


joshfriend:
All of this seems to be covered already in the [common build problems wiki](https://github.com/pyenv/pyenv/wiki/Common-build-problems)

This wiki page is linked in the README, the new issue template, and also every time a python build fails.

britho:
Ah thanks.  I missed that, my fault.  I'll close this out.

PR #1390, opened 2019-08-31
Add CPython 3.8.0b4

### LINKED ISSUE #1389: Include beta versions
Two days ago CPython 3.8.0b4 was released:

https://www.python.org/downloads/release/python-380b4/

However, pyenv will only allow `3.8-dev`, which has two disadvantages:

* It requires cloning the full CPython repository, which is much slower than downloading the source tarball
* At this moment it's two commits ahead of 3.8.0b4, and will be more separated in the future

There was a commit claiming to add CPython 3.8.0b1 (https://github.com/pyenv/pyenv/commit/2350e57b00f0003e156f65b0555d823c96c069c9) but in fact it was adding the git install from the `3.8` branch.

Juanlu001:
Fix #1389.

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1389

### Description

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/3.8.0b4
+++ b/plugins/python-build/share/python-build/3.8.0b4
@@ -1,0 +1,10 @@
+#require_gcc
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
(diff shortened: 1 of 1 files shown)


Juanlu001:
Thanks!

PR #1396, opened 2019-09-18
Fix path to Unix port in micropython build

cdwilson:
This fixes a path in #1394 (sorry, I had pushed the wrong commit earlier...)

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1010,7 +1010,7 @@ build_package_micropython() {
   fi
   { cd mpy-cross
     "$MAKE" $MAKE_OPTS
-    cd ports/unix
+    cd ../ports/unix
     "$MAKE" $MAKE_OPTS axtls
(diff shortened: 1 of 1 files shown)


plugins/python-build/bin/python-build
```diff
@@ -1010,7 +1010,7 @@ build_package_micropython() {
   fi
   { cd mpy-cross
     "$MAKE" $MAKE_OPTS
```
yyuu: wouldn't it be simpler if you do as "$MAKE" -C mpy-cross $MAKE_OPTS ...?

cdwilson: > wouldn't it be simpler if you do as `"$MAKE" -C mpy-cross $MAKE_OPTS`

yes, you are totally right, feel free to make the change as you suggested

yyuu: Let's defer the change for now until something else comes up 😉

Issue #1375, opened 2019-07-18
pyenv install doesn't work with homebrew installed tcl-tk 


grahamannett:
When installing any version of python with pyenv, I can't get tkinter to work.  Python install fine but I believe that the files related to tcl-tk are causing issues and python-build either needs to skip the files related in 

For instance with python installed from pyenv in any manner of way:
```
env CFLAGS="-I/usr/local/opt/tcl-tk/include" CPPFLAGS="-I/usr/local/opt/tcl-tk/include" LDFLAGS="-L/usr/local/opt/tcl-tk/lib"  pyenv install --verbose 3.7.4
```

```
env CPPFLAGS="-I/usr/local/opt/tcl-tk/include" LDFLAGS="-L/usr/local/opt/tcl-tk/lib"  PYTHON_CONFIGURE_OPTS="--enable-framework" pyenv install --verbose 3.7.4
```

```
env PYTHON_CONFIGURE_OPTS="--enable-framework" pyenv install --verbose 3.7.4
```
etc. 

and then trying to use tkinter from that python:
```
python -m tkinter -c 'tkinter._test()'                                                                                  
Traceback (most recent call last):
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/tkinter/__main__.py", line 7, in <module>
    main()
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/tkinter/__init__.py", line 3988, in _test
    root = Tk()
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/tkinter/__init__.py", line 2025, in __init__
    self._loadtk()
  File "/Users/graham/.pyenv/versions/3.7.4/lib/python3.7/tkinter/__init__.py", line 2040, in _loadtk
    % (_tkinter.TK_VERSION, tk_version))
RuntimeError: tk.h version (8.6) doesn't match libtk.a version (8.5)
```

Looking at the log does it matter that the tcl-tk stuff that is included with macosx is included before the brew ones?
```
building '_tkinter' extension
[...]

Issue #1125, opened 2018-03-21
Cannot pass multipart options to configure


ezwelty:
I am trying to override the paths to the Tcl/Tk libraries by setting `PYTHON_CONFIGURE_OPTS`. When building manually, I can do this:

```bash
./configure \
  --with-tcltk-includes="-I/usr/local/opt/tcl-tk/include" \
  --with-tcltk-libs="-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6"
```

But with pyenv, no matter what I try, I get:

```bash
$ env PYTHON_CONFIGURE_OPTS=SOMETHING pyenv install 2.7.14 --verbose
...
configure: error: unrecognized option: `-ltcl8.6'
```

Lines of interest from the stack trace suggest a parsing bug somewhere:

```
+ eval 'PYTHON_CONFIGURE_OPTS_ARRAY=( "${value[@]}" )'
++ PYTHON_CONFIGURE_OPTS_ARRAY=("${value[@]}")
+ [[ '' == *\-\-\e\n\a\b\l\e\-\s\h\a\r\e\d* ]]
+ [[ --with-tcltk-includes='-I/usr/local/opt/tcl-tk/include' --with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6' == *\-\-\e\n\a\b\l\e\-\s\h\a\r\e\d* ]]
+ [[ '' == *\-\-\s\h\a\r\e\d* ]]
+ [[ --with-tcltk-includes='-I/usr/local/opt/tcl-tk/include' --with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6' == *\-\-\e\n\a\b\l\e\-\f\r\a\m\e\w\o\r\k* ]]
+ [[ --with-tcltk-includes='-I/usr/local/opt/tcl-tk/include' --with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6' == *\-\-\e\n\a\b\l\e\-\u\n\i\v\e\r\s\a\l\s\d\k* ]]
+ [[ --with-tcltk-includes='-I/usr/local/opt/tcl-tk/include' --with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6' != *\-\-\e\n\a\b\l\e\-\u\n\i\c\o\d\e\=* ]]
...
[...]

PR #1397, opened 2019-09-18
working with tcl-tk in PYTHON_CONFIGURE_OPTS

grahamannett:
Hey this should fix a few issues such as #1375 and #1125 .  I am not sure if there is a better way to go about doing this but this seems to work while not making issues for other variables passed to pyenv install.

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1125
  - https://github.com/pyenv/pyenv/issues/1375

### Description
- This fix would allow tcl-tk stuff passed in a string. For instance 
PYTHON_CONFIGURE_OPTS="--with-tcltk-includes='-I/usr/local/opt/tcl-tk/include' --with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6'" pyenv install 3.7.3
order doesnt matter: 
PYTHON_CONFIGURE_OPTS="--with-tcltk-libs='-L/usr/local/opt/tcl-tk/lib -ltcl8.6 -ltk8.6' --with-tcltk-includes='-I/usr/local/opt/tcl-tk/include'" pyenv install 3.7.3

### Tests

I'm not sure how to test this without manually building and running something like 
`python -m tkinter -c 'tkinter._test()'` and verifying the box pops up.  It is possible it installs the python version fine and the tkinter stuff doesn't work.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -2222,6 +2222,19 @@ if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
   fi
 fi
 
+# regex_to_match="(--with-tcltk-libs='([^']+)')"
+if [[ "$PYTHON_CONFIGURE_OPTS" =~ (--with-tcltk-libs=\'([^\']+)\') ]]; then
+  tcltk_match="${BASH_REMATCH[1]}"
+  tcltk_match_quoted="${tcltk_match//--with-tcltk-libs=/}"
+  # remove it from PYTHON_CONFIGURE_OPTS since it will mess up compile
+  PYTHON_CONFIGURE_OPTS="${PYTHON_CONFIGURE_OPTS//$tcltk_match/}"
+
+  # having issues passing the single quoted part, couldnt pass as single var and still work
+  package_option python configure "--with-tcltk-libs='${tcltk_match_quoted}'"
+  unset tcltk_match
+  unset tcltk_match_quoted
+fi
+
 # Unset `PIP_REQUIRE_VENV` during build (#216)
 unset PIP_REQUIRE_VENV
 unset PIP_REQUIRE_VIRTUALENV


plugins/python-build/bin/python-build
```diff
@@ -2222,6 +2222,19 @@ if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
   fi
 fi
 
+# regex_to_match="(--with-tcltk-libs='([^']+)')"
+if [[ "$PYTHON_CONFIGURE_OPTS" =~ (--with-tcltk-libs=\'([^\']+)\') ]]; then
+  tcltk_match=${BASH_REMATCH[1]}
```
yyuu: Please add surrounding double quotes.

plugins/python-build/bin/python-build
```diff
@@ -2222,6 +2222,19 @@ if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
...
+  tcltk_match_quoted=${tcltk_match//--with-tcltk-libs=/}
```
yyuu: ditto

plugins/python-build/bin/python-build
```diff
@@ -2222,6 +2222,19 @@ if [[ "$PYTHON_CONFIGURE_OPTS" != *"--enable-unicode="* ]]; then
...
+  # remove it from PYTHON_CONFIGURE_OPTS since it will mess up compile
+  PYTHON_CONFIGURE_OPTS=${PYTHON_CONFIGURE_OPTS//$tcltk_match/}
```
yyuu: ditto

plugins/python-build/bin/python-build
grahamannett: im not sure I understand, double quotes around the regex? like `"(--with-tcltk-libs=\'([^\']+)\')"` ?   That wont work.  Ideally I guess I can make the regex a var that it matches so then you dont have the \ to escape the single quote.  Or are you referring to something else

grahamannett: Oh, Im guessing you are meaning around `${BASH_REMATCH[1]}` and similar? will do

PR #1409, opened 2019-10-08
allow tcl-tk as argument or try with homebrew by default

grahamannett:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1125
For anyone trying to use tcltk from homebrew before (or if you had tcltk installed from homebrew tkinter would fail I believe)
 #1125 
### Description

I previously added the regex match for this, but having to pass the config opts everytime while other packages use homebrew by default made this seem worthwhile.  If there is anything that would make this more streamlined, or a better place to put it in python-build then let me know.  

This should hopefully allow more people to use pyenv and things like matplotlib or whatever that may previously had issues.

### Tests

### DIFF
--- a/libexec/pyenv-help
+++ b/libexec/pyenv-help
@@ -27,6 +27,8 @@ command_path() {
 }
 
 extract_initial_comment_block() {
+  LC_CTYPE=C
+  LANG=C
   sed -ne "
     /^#/ !{
       q


grahamannett:
can this get looked at @yyuu or someone else?  itd be helpful for installing 3.8.0 and those of use who use install pyenv --HEAD from brew

grahamannett:
Can this be merged so users of asdf can use python-build and not run into problems?

grahamannett:
@joshfriend or @yyuu any interest in merging this?

joshfriend:
can you fix the tests?

grahamannett:
@joshfriend im not sure i understand travis-ci enough to know why it is failing.  locally my tests run exactly the same.  maybe ill try again and see if i can add a test for the gui portion as it seems plausible and im worried i will create issues for linux users

Issue #1400, opened 2019-09-22
Documentation: Zsh should modify zshrc instead of zshenv 


ryanermita:
# Description
- macOS Mojave
- pyenv version: 1.2.13
- Python version: 2.7.16 (system default)
- z shell version: zsh 5.3 (x86_64-apple-darwin18.0)
- homebrew version: 2.1.11

In the [installation guide](https://github.com/pyenv/pyenv#installation), for the step 3 it says that:
```
Zsh note: Modify your ~/.zshenv file instead of ~/.bash_profile.
```

this step makes `pyenv` unable to use shims properly and I encounter this issue #1238 . this is solved (https://github.com/pyenv/pyenv/issues/1238#issuecomment-533856021) by moving this code snippet below from `.zshenv` to `.zshrc`
```
if command -v pyenv 1>/dev/null 2>&1; then
  eval "$(pyenv init -)"
fi
```
Is this a documentation issue or related specifically to zsh version?

mazzma12:
and what about `~/.zprofile` (It is sourced at the login session) ? I have it this way and no problems at the moment.

hsnusonic:
* zsh version: 5.7.1
* Homebrew version: 2.1.11

I used `brew install pyenv` and followed the installation guide. I also found that `.zshenv` not worked for me, so I use `.zprofile` instead.

ryanermita:
@mazzma12 @hsnusonic I include the zsh and homebrew version in the issue description, this might be helpful for this issue.

joshfriend:
`.zshrc` is only sourced for interactive shells, whereas `.zshenv` is always sourced. Moving the init could cause you other problems.

ryanermita:
ohh. thanks for the info @joshfriend

kekhuay:
i can confirm that `.zshenv` doesn't work on Arch Linux and OSX but `.zshrc` and `.zprofile` does.

tekumara:
I've also encountered this problem on macOS Mojave with `brew install pyenv`.

brew installs pyenv into `/usr/local/bin`. However when `command -v pyenv` runs inside `.zshenv`, the path does not yet contain `/usr/local/bin` and so pyenv isn't initialised.

`/etc/zprofile` adds `/usr/local/bin` to the path but runs ***after*** `.zshenv`.

The execution order is:
```
.zshenv → [.zprofile if login] → [.zshrc if interactive] → [.zlogin if login] → [.zlogout sometimes]
```

[...]

PR #1447, opened 2019-11-02
Update README.md to use .zshrc

gsheni:
### Description
- With the latest macOS Catalina, I found that setting PYENV and PATHs in `.zshenv` did not properly load the python shim. Moving the commands to `.zshrc` fixed this. 
- Furthermore, I would suggest changing all `.zshenv` references to `.zshrc`
> macOS Terminal considers every new shell to be a login shell and an interactive shell. So, in Terminal a new zsh will potentially run all configuration files. For simplicity’s sake, you should use just one file. The common choice is .zshrc.


joshfriend:
#1400

PR #1449, opened 2019-11-06
Fix Readme fo zsh

michalkoza:

### Prerequisite

### Description
pyenv resides in /usr/local/bin. This path is not yet in the PATH when .zshenv is being sourced. It is added to path during .zprofile sourcing so .zprofile seems a better place thatn .zshenv

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ easy to fork and contribute any changes back upstream.
     $ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
     $ echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
     ```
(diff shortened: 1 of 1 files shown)


joshfriend:
Please see #1400.

I'm not sure why so many people are suggesting this change recently

michalkoza:
Well there is an issue here obviously so no wonder people suggest fixes.
You say that it must be .zshenv because only this is guaranteed to be sourced.
But you seem to ignore our reasoning: that during .zshenv pyenv is not in the PATH as /usr/local/bin  is not in the PATH yet so it cannot work at all the way it is now.
Our suggestions may not be perfect but at least then it works for "normal" sessions.

leochoo:
@joshfriend  #1400  does not solve the problem. Please don't ignore other comments and acknowledge that `zshenv` is not working.

PR #1450, opened 2019-11-06
Correct URLs for Stackless builds and add Stackless 2.7.16

nathan-v:
* Stackless.com seems to no longer have SSL support so Stackless URLs to that site updated to HTTP
* Added Stackless 2.7.16


joshfriend:
> Stackless.com seems to no longer have SSL support

The stackless.com site redirects to their github wiki page, so probably switch the download URLs to ones from the [releases page](https://github.com/stackless-dev/stackless/releases)

nathan-v:
https://github.com/stackless-dev/stackless/wiki/Download

The Wiki points to the files on the site still but HTTPS via Github instead makes sense. I'll update the 
PR.

Issue #1454, opened 2019-11-11
`pyenv help realpath.dylib' results in "sed: RE error: illegal byte sequence"


ericvw:
This recently came to my attention since the [v1.2.15](https://github.com/pyenv/pyenv/blob/master/CHANGELOG.md#1215) release which enabled showing text for all `pyenv` commands.

Due to `realpath.dylib` being a compiled binary, the normal shell routines for extracting the help text won't work.

### Details
- Platform information (e.g. Ubuntu Linux 16.04): macOS 10.14.6
- OS architecture (e.g. amd64): amd64
- pyenv version: 1.2.15

Relevant DEBUG output:
```shell
> env PYENV_DEBUG=1 pyenv help realpath.dylib
[...]
+ [pyenv:118] case "$command" in
++ [pyenv:131] command -v pyenv-help
+ [pyenv:131] command_path=/usr/local/Cellar/pyenv/1.2.15/libexec/pyenv-help
+ [pyenv:132] '[' -z /usr/local/Cellar/pyenv/1.2.15/libexec/pyenv-help ']'
+ [pyenv:140] shift 1
+ [pyenv:141] '[' realpath.dylib = --help ']'
+ [pyenv:148] exec /usr/local/Cellar/pyenv/1.2.15/libexec/pyenv-help realpath.dylib
+ [pyenv-help:19] '[' realpath.dylib = --complete ']'
+ [pyenv-help:143] unset usage
+ [pyenv-help:144] '[' realpath.dylib = --usage ']'
+ [pyenv-help:149] '[' -z realpath.dylib ']'
+ [pyenv-help:149] '[' realpath.dylib == pyenv ']'
+ [pyenv-help:159] command=realpath.dylib
++ [pyenv-help:160] command_path realpath.dylib
++ [pyenv-help:25] local command=realpath.dylib
++ [pyenv-help:26] command -v pyenv-realpath.dylib
+ [pyenv-help:160] '[' -n /usr/local/Cellar/pyenv/1.2.15/libexec/pyenv-realpath.dylib ']'
+ [pyenv-help:161] '[' -n '' ']'
+ [pyenv-help:164] print_help realpath.dylib
+ [pyenv-help:114] local command=realpath.dylib
+ [pyenv-help:115] local summary usage help
++ [pyenv-help:116] documentation_for realpath.dylib
++ [pyenv-help:90] local filename
+++ [pyenv-help:91] command_path realpath.dylib
[...]

PR #1456, opened 2019-11-12
Filter out 'realpath' bash builtin extension from 'commands'

ericvw:
Due to fcf539ec enabling the help message being dynamic and discovering
everything from `pyenv-commands`, a side effect was the `realpath*`
`bash` builtin extension being listed.  This change elides the shared
library from being listed since it is not a command one can execute via
`pyenv <command>`.

Closes: #1454

### DIFF
--- a/libexec/pyenv-commands
+++ b/libexec/pyenv-commands
@@ -40,4 +40,4 @@ shopt -s nullglob
       fi
     done
   done
-} | sort | uniq
+} | grep -v '^realpath' | sort | uniq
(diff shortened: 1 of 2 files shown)


joshfriend:
Thanks for the PR, is your solution here addressing the same problem as #1455?

If so, I'd lean towards closing this in favor of #1455 because a blacklist would need periodic updating (and it seems like the other approach filters the commands list more directly based on the root cause of the issue.)

ericvw:
> Thanks for the PR, is your solution here addressing the same problem as #1455?

Yes, it looks like it.  PR #1455 looks like a better general approach to solving this problem.

Closing this in favor of #1455.

PR #1458, opened 2019-11-14
Nicer zsh completion

mkoskar:
This is completion based on ``zshcompsys`` rather than older ``zshcompctl``.

### DIFF
--- a/completions/pyenv.zsh
+++ b/completions/pyenv.zsh
@@ -1,18 +1,12 @@
-if [[ ! -o interactive ]]; then
-    return
-fi
+#compdef pyenv
 
-compctl -K _pyenv pyenv
+# Install as _pyenv into your fpath e.g., /usr/share/zsh/site-functions/ or:
+#
+#   $ . <gitdir>/completions/pyenv.zsh
+#   $ compdef _pyenv pyenv
 
 _pyenv() {
-  local words completions
-  read -cA words
-
-  if [ "${#words}" -eq 2 ]; then
-    completions="$(pyenv commands)"
-  else
-    completions="$(pyenv completions ${words[2,-2]})"
-  fi
-
-  reply=(${(ps:\n:)completions})
+    _arguments \
+        ': :("${(f)$(pyenv commands)}")' \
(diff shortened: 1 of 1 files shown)


completions/pyenv.zsh
```diff
@@ -1,18 +1,12 @@
-if [[ ! -o interactive ]]; then
```
mayanksuman: Why this is removed?

The completion should not be loaded if zsh is not interactive. Please revert this change.

mkoskar: The whole point is that you just drop the file in your ``fpath`` and you're done. That's how majority of ``zsh`` completions are done and it's also suitable for packagers. Alternative is to ``compdef`` it yourself if you prefer and that you can do from your ``.zshrc`` after "interactive check".

mayanksuman: Yeah! It is easier only when a majority of users get pyenv by packaging, but most of them get it by git clone. The `pyenv-installer` also uses `git clone`.

The completion file in question is dynamically loaded when user run `eval $(pyenv init -)` in `~/.zshrc`. However, if want to add it to `fpath`, it is not as easy as sourcing. Please refer to [this answer]( https://unix.stackexchange.com/a/213658).

mayanksuman: Further, I guess the output of `pyenv init -` need to change for the changes in completion file.

mkoskar: Ah yeah, I see ... I haven't noticed it in my installation. Indeed running from git working copy include ``source .../pyenv.zsh``. TBH I don't run ``eval $(pyenv init -)`` as I don't need all the bell and whistles. This was inspired by Archlinux' ``pyenv`` package which includes ``pyenv.zsh`` in ``/usr/share/zsh/site-functions`` and obviously doesn't work, even doing ``eval $(pyenv init -)`` wouldn't help.

Not sure where to get from here, as for me I'm happy to keep this in my ``~/.zfunctions/_pyenv`` but tweaking ``libexec/pyenv-init`` a bit might solve it for both usecases.

adigitoleo:
Either this issue or #1644 could be closed to avoid confusing potential contributors.

PR #1464, opened 2019-11-19
Enable users to discover how to update the python-build plugin

jsoref:

### Prerequisite
Considered. Declined.
rbenv/ruby-build#1380

### Description
- [x] I was using a `brew install pyenv` environment and was instructed to install `python 3.8`. I ran `pyenv install -l` and all I saw was `3.8-dev`. I checked with `brew` and confirmed I was running the latest version of `pyenv`. We checked and the person who instructed me to install it was using an older version of `pyenv`. This was incredibly confusing. I checked the `pyenv install --help` and there was nothing obvious that indicated what I was doing wrong.

At some point, we decided that I should just install `3.8-dev`, and I managed to mistype the name (as `python3.8-dev` iirc) and then I saw the command that explained how to update, and that's when we understood what was missing.

This PR attempts to address that missing link by making the standard path for normal users easier.

### Tests

### DIFF
--- a/plugins/python-build/bin/pyenv-install
+++ b/plugins/python-build/bin/pyenv-install
@@ -24,6 +24,10 @@
 # python-build, including a list of environment variables for adjusting
 # compilation, see: https://github.com/pyenv/pyenv#readme
 #
+# If the version you are looking for is missing, you probably need to update
+# your pyenv python-build plugin. For information on how to do so, run:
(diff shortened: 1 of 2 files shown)


clbarnes:
Why not just tell people to `pyenv update`?

native-api:
Python-build is now bundled with Pyenv so this is no longer relevant.

PR #1502, opened 2019-12-26
Add PyPy 7.3.0

lazka:
The official pypy builds should now be portable on linux.
Also add aarch64 builds.


plugins/python-build/bin/python-build
```diff
@@ -1030,6 +1030,7 @@ pypy_architecture() {
     "ppc64" ) echo "linux-ppc64" ;;
     "ppc64le" ) echo "linux-ppc64le" ;;
     "x86_64" ) echo "linux64" ;;
+    "aarch64" ) echo "linux-aarch64" ;;
```
yyuu: 😄

PR #1509, opened 2020-01-03
Replace `type cmd` with `cmd --help`

jbylund:
### Description
This addresses a case I find myself in where aria2c is "present" on a system, but broken due to system libraries being in a bad state.  It replaces the `type cmd` check to see if an executable is present with a `cmd --help` to see if the executable can be executed, and exit non-zero.

### Tests
No new tests added (should still prefer aria2c/curl/wget as before)

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -339,7 +339,7 @@ http() {
 detect_http_client() {
   local client
   for client in aria2c curl wget; do
-    if type "$client" &>/dev/null; then
+    if "$client" --help &>/dev/null; then
       echo "$client"
       return
     fi
--- a/plugins/python-build/test/fetch.bats
+++ b/plugins/python-build/test/fetch.bats
@@ -21,6 +21,7 @@ setup() {
 @test "using aria2c if available" {
   export PYTHON_BUILD_ARIA2_OPTS=
   export -n PYTHON_BUILD_HTTP_CLIENT
+  stub aria2c "--help : true"
   stub aria2c "--allow-overwrite=true --no-conf=true -o * http://example.com/* : cp $FIXTURE_ROOT/\${5##*/} \$4"
 
   install_fixture definitions/without-checksum


native-api (commented):
I don't think that this is something that we should be recovering from...
If "system libraries are in a bad state", anything else can break at any moment just as well.

Even the suggested change is a not a reliable test. E.g. the program can display help okay -- but crash once it tries to do something more sophisticated that involves more libraries.

jbylund:
Right now the pattern could I guess be summarized as:

```python
best_candidate = None
for best_candidate in candidates_by_preference:
    if candidate_exists(candidate):
        break
else:
    raise SomeError("no downloaders found")
download_using(best_candidate)
```
or something that smells like that.  But there's a chance that a candidate exists but is semi or non-functional.  What about

```python
for candidate in candidates_by_preference:
	try:
		download_using(candidate)
		break
	except:
		pass # <= couldn't download successfully, either not found or not functional
else:
	raise SomeError("no functional downloaders found")
```

seems to me that this would test exactly what we care about (can you successfully download using the given tool), rather than a proxy which is not perfectly correlated?

jbylund:
Also, whoops, forgot this wasn't a python repo itself.

native-api:
> But there's a chance that a candidate exists but is semi or non-functional.

That's what I'm talking about. We do not, and should not be bothering about this. If a user has a broken system, it's their problem, not ours. We aren't wasting our time and effort working around problems resulting from others' gross negligience. In fact, we can't do that anyway even if we wanted to, as per above.

jbylund:
What about trying to fetch using different tools instead of testing for tool presence then failing if a present tool fails the download?

native-api:
A tool can fail the download for a zillion reasons that don't warrant switching to another tool

native-api:
@jbylund Why don't you just fix the actual problem rather than waste your time shopping for workarounds? If some of your systems have a broken `aria2c` -- just delete it from them! Since it can't be used anyway, there's no point in keeping it, right?

If you can't do that for whatever reason, change the configuration for yourself so that that it won't be found.

jbylund:
This is a shared system where I don't have access to change the installed python (or the same aria2c).  pyenv _is_ the workaround to not being able to get system admins to keep things up to date/working.  I did make a fix like the one in this pr so that I now have a working python in my home dir.

native-api:
> If you can't do that for whatever reason, change the configuration for yourself so that it won't be found.

What I mean, you can make a workaround that works independently from Pyenv.

E.g. if `aria2c` is in `/usr/bin`:

* make a directory inside your profile
* hardlink or symlink everything from `/usr/bin` there _except_ the broken programs that you want to hide.
* replace `/usr/bin`'s `PATH` entry with this dir.

Another way is to make a script with the name "`aria2c`" that would run something else (that works) and place it earlier on the `PATH`.

jbylund:
It's clear to me that there are other ways I could have installed python in my users home dir.
It seems to me that this change would make pyenv work in some situations where it currently does not.
I think your worry is there are situations in which pyenv currently works where it would not with this change?

Issue #969, opened 2017-08-18
Do not change PATH unnecessarily / multiple times


silenc3r:
```eval "$(pyenv init -)"``` prepends PATH variable with ```${PYENV_ROOT}/shims``` every time it is executed.

If you have ```eval "$(pyenv init -)"``` in your ```~/.bashrc``` and run tmux then you have ```${PYENV_ROOT}/shims``` appear twice in your PATH.  

It's a minor nitpick but it's easily solvable with adding something like:

    if [[ ! "$PATH" == "*${PYENV_ROOT}/shims*" ]]; then
        ...code modifying PATH...
    fi

to scripts modifying PATH

blueyed:
PR welcome.

But in general it is recommended to call it in your shell's profile, and not rc file.

blueyed:
btw: as for Zsh: $PATH/$path is typically setup to have unique entries.
There it is a matter of not prepending it before other entries then.

blueyed:
The PR should probably get done against rbenv then also, so that it can be merged from there.

silenc3r:
Tmux for some reason spawns login shell by default so putting pyenv init in shell's profile doesn't change anything in that case.

blueyed:
True.

I am using this in my tmux.conf: `set -g default-command "exec $SHELL"`

darkfishy:
**Assumption**: Using BASH for $SHELL

I would suggest against calling `$(pyenv init -)` in either *.bash_profile* or *.profile* because, by design, BASH doesn't call *.\*profile* unless it is a login shell. However, if chained properly, `.bashrc` can get called for both login and interactive shells. See below for examples.

    ~/ $ bash
    Activating .bashrc
    ~/ $ bash -l
    Activating .bash_profile
    Activating .profile
    Activating .bashrc
    ~/ $

@silenc3r: Does your way work without uglifying `$PATH`?

jack-pace:
For what it's worth, if you are using `pyenv-virtualenv` you don't have a choice here. You need to run `eval $(pyenv init -)` in your `.bashrc`, else the `pyenv` _function_ is not defined in the tmux shell and you cannot run `pyenv activate ...`.

> Failed to activate virtualenv.
> 
> Perhaps pyenv-virtualenv has not been loaded into your shell properly.
> Please restart current shell and try again.

native-api:
Duplicate of https://github.com/pyenv/pyenv/issues/1649

PR #1513, opened 2020-01-06
stop re export shims if exists

mozartilize:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/969

### Description

If there is a better solution like use macro for reusable code plz guide me :)

Just make it works cause it's really annoying when I have to restart the shell with `exec zsh` and the PATH get populated

### Tests


mozartilize:
maybe the test "can add shims to PATH more than once" should be removed

Issue #1445, opened 2019-11-01
Docs: how to test added python versions


clbarnes:
A common and indeed neverending dev task is to add new versions of python. The best way to test these is unclear, especially when they're added in an automated way or in bulk. e.g. #1361 has been sitting stale for some time.

What is the best way to test one, or more, added python versions, without interference with an existing pyenv installation (presumably most pyenv devs will also be pyenv users). Presumably people have been testing python versions since pyenv was created - how do you do it, and can that information be added to the README for new devs to use?

clbarnes:
@yyuu any thoughts on this? Would be really beneficial to get PR #1361 merged and lower the barrier to entry for all contributors.

native-api:
The way to test a Python installation is to install it and run `python -m test`.

The question is, which installations to test, and in what environments?

Plus, a test failure may be due to some kind of unfixed bug or incompatibility (especially for prebuilt versions that were built for another environment), through no fault of Pyenv.

So, I suggest to (which is how current scripts are written):

* write installation scripts with a specific range of supported environments in mind and manually test in those as per above
* base them on existing scripts since those gradually accumulate proven best practices as problems are discovered and fixed
* if it's a prebuilt binary, insert a guard like `require_distro()` to warn users if they install it in an environment that it wasn't designed for

native-api:
Could you perhaps make a PR that would add this information to the README (a subsection in https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md and a pointer to that README from the Developement section in the main README, I suppose)?

This is probably what you expected to see when you were looking for documentation.

PR #1547, opened 2020-03-05
Vagrant config for testing python-build

clbarnes:
### Prerequisite

  - https://github.com/pyenv/pyenv/issues/1445

### Description

For some time, there hasn't been a clear path for testing python-build. I suspect that most pyenv developers are pyenv users as well: there has been no documented way of keeping the development version of pyenv away from the version of pyenv supporting their python work.

The deafening silence in #1445 and #1361 suggests it's an issue which currently doesn't have an easy answer. Now it does.

With vagrant and virtualbox installed, a single command will spin up a clean ubuntu 18.04 virtual machine with all of the build dependencies, and the user's local pyenv repo mounted at `~/.pyenv` and installed in the usual way. Then they can install python versions to their heart's content, and can trash the VM whenever they want, with zero impact on their host machine and its pyenv installation. It's completely optional and doesn't impact the development of anyone not using this workflow (I don't know how those people are testing their builds, and apparently they don't want to tell me). Changes made inside the VM are persisted on the host, and vice versa. Users just have to be careful not to make commits inside the VM if they are relying on the host's global git config.

### Tests

### Alternatives

Could use docker instead of vagrant, and provide a docker image which has the build requirements installed already. Rebuilding a clean container would probably be faster under docker, and I think docker has fewer restrictions on e.g. virtualisation being enabled in the BIOS. However, docker as a tool is MUCH less ergonomic than vagrant (unless you also use docker-compose), and you run into issues with persistence and file ownership when trying to use docker for development rather than deployment.

### Possible extensions

An additional script for attempting to build lots of python versions in a batch, possibly in a multi-threaded way, and report successes/ failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,6 +391,21 @@ tracker](https://github.com/pyenv/pyenv/issues).
   [pyenv-virtualenv]: https://github.com/pyenv/pyenv-virtualenv#readme
   [hooks]: https://github.com/pyenv/pyenv/wiki/Authoring-plugins#pyenv-hooks
 
+### Testing new python versions
+
+This repository contains a [vagrant](https://www.vagrantup.com/) configuration,
+to spin up and provision a Ubuntu 18.04 virtual machine with your local copy of the repo
+mounted at `~/.pyenv`.
+
+With vagrant and [virtualbox](https://www.virtualbox.org/) installed, simply
+
+```sh
+vagrant up
+vagrant ssh
+```
+
+and you will be dropped into a bash shell where you can test your pyenv changes.
+
 ### Version History
 
 See [CHANGELOG.md](CHANGELOG.md).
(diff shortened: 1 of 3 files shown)


clbarnes:
More on docker vs vagrant

If all you want to do is check that the build completes, docker is the better option. It's much lighter weight, and the spinning up and trashing processes are much, much faster, and can be done in a single line (`docker run -t my_container pyenv install 3.8.1`). It's a little less intuitive in that you have to rebuild the image for every change you make to pyenv. Multiplexing installations can be done from the host, each in their own container, which is a big plus.

Vagrant is slightly more difficult to set up, slightly more verbose to actually do the testing, certainly less zeitgeisty. However, it makes it much easier to actually test the python versions you're building.

Having now set up both, I'm leaning towards docker. Interested in which is more likely to be merged.

clbarnes:
Opened #1548 ; feel free to pick one

clbarnes:
Decided docker is better.

PR #1548, opened 2020-03-05
Docker config for testing python-build

clbarnes:
An alternative to #1547 (see there for more discussion).

Pick this one if you want a faster spin up -> build python -> destroy environment cycle. Pick vagrant if you want to be able to make changes to your build script, have them persisted outside of the container, and re-test in the same container.

### DIFF
--- a/plugins/python-build/README.md
+++ b/plugins/python-build/README.md
@@ -220,3 +220,51 @@ Please see the [pyenv wiki](https://github.com/pyenv/pyenv/wiki) for solutions t
 If you can't find an answer on the wiki, open an issue on the [issue
 tracker](https://github.com/pyenv/pyenv/issues). Be sure to include
 the full build log for build failures.
+
+## Contributing
+
+### Testing new python versions
+
+If you are contributing a new python version for python-build, 
+you can test the build in a [docker](https://www.docker.com/) container based on Ubuntu 18.04.
+
+With docker installed:
+
+```sh
+docker build -t my_container .
+docker run my_container pyenv install <my_version>
+```
+
+To enter a shell which will allow you to build and then test a python version,
+replace the second line with
+
+```sh
+docker run -it my_container
+```
+
+The container will need to be rebuilt whenever you change the repo,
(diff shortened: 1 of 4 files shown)


clbarnes:
### Possible extension

Have travis check whether a commit has more versions than master, and if so, try to build every one of them. It may take a while but it will prove that new additions work.

clbarnes:
On the testing point, you can test all new or changed versions compared to master with

```sh
docker build -t test_container

git diff --name-only master \
  | grep '^plugins/python-build/share/python-build/' \
  | awk -F '/' '{print $NF}'
  | xargs -I _ docker run test_container pyenv install _
```

plugins/python-build/README.md
```diff
@@ -220,3 +220,51 @@ Please see the [pyenv wiki](https://github.com/pyenv/pyenv/wiki) for solutions t
...
+
+With docker installed:
+
+```sh
+docker build -t my_container .
+docker run my_container pyenv install <my_version>
+```
+
+To enter a shell which will allow you to build and then test a python version,
+replace the second line with
```
cclauss, line 239:
```suggestion
### Testing new Python versions

If you are contributing a new Python version for python-build, 
you can test the build in a [Docker](https://www.docker.com/) container based on Ubuntu 18.04.

With Docker installed:

```sh
docker build -t my_container .
docker run my_container pyenv install <my_version>
```

To enter a shell which will allow you to build and then test a python version,
replace the second line with
```

clbarnes:
Would be great to get a review on this! @yyuu ? I found it invaluable for testing new miniconda builds when I added a bunch a while back. The above bash script could potentially go it into CI for new build scripts.

clbarnes:
I'd love to get this reviewed - @joshfriend @chrahunt , maybe? Not really sure how to move this forward.

When I was working on bulk-contributing new versions previously, nobody was able to answer how to test that they work "en masse". I *still* don't know how people are testing newly submitted versions. This represents a drastic improvement whatever that testing workflow is, and doesn't seem to add much maintenance overhead.

PR #1549, opened 2020-03-05
Customise `.python-version` filename for local project versions

nolar:
### Description

[22K+ projects](https://github.com/search?l=&q=filename%3A.python-version&type=Code) define their own `.python-version` files in their source code, version-controlled.

This file supposed to be an environment configuration, which is individual to every developer and their habits/preferences. A strict Python version to be used in a project can be opinionated. For actual code restrictions, `python_requires="==3.8.1"` of `setup.py` should be used.

This also breaks workflows when plugins like `pyenv-virtualenv` are used, with their own named versions instead of numeric Pythons. Having a fixed numeric Python version in a project assumes that all dependencies will be installed into that binary Python build instead of virtualenvs — for all projects that rely on this Python version.

With this PR, the developers will have an option to ignore the recommended Python versions in `.python-version` file of every project, and configure their own dot-file name to use by setting `PYENV_VERSION_FILENAME` on their workstations.

_The variable name looks not so good, can be discussed._

### Prerequisite

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,6 +369,7 @@ name | default | description
 `PYENV_DEBUG` | | Outputs debug information.<br>Also as: `pyenv --debug <subcommand>`
 `PYENV_HOOK_PATH` | [_see wiki_][hooks] | Colon-separated list of paths searched for pyenv hooks.
 `PYENV_DIR` | `$PWD` | Directory to start searching for `.python-version` files.
+`PYENV_VERSION_FILENAME`  | `.python-version` | Customize filename for local project versions.
 `PYTHON_BUILD_ARIA2_OPTS` | | Used to pass additional parameters to [`aria2`](https://aria2.github.io/).<br>If the `aria2c` binary is available on PATH, pyenv uses `aria2c` instead of `curl` or `wget` to download the Python Source code. If you have an unstable internet connection, you can use this variable to instruct `aria2` to accelerate the download.<br>In most cases, you will only need to use `-x 10 -k 1M` as value to `PYTHON_BUILD_ARIA2_OPTS` environment variable
 
 
--- a/libexec/pyenv-local
+++ b/libexec/pyenv-local
@@ -32,9 +32,9 @@ fi
 versions=("$@")
 
 if [ "$versions" = "--unset" ]; then
-  rm -f .python-version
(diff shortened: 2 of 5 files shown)


clbarnes:
My (admittedly somewhat unfeeling) personal opinion:

22,000 projects are wrong, and when they figure that out because their development environments start showing problems, they'll delete the file, stick it in the gitignore, and everyone will be better off for it with next to no effort on their part. I suspect that you'd find thousands of projects with vim swap files and git merge temporary files committed too, but it's not up to the vim and git developers to overcomplicate their product because people aren't using the *simple* version of it correctly.

nolar:
@clbarnes Do they also put local paths like `/Users/nolar/src/…` into git-committed `.env`?

Precise patch-versions of Python are too specific to a local environment and a configuration of a workstation: some versions can be installed, some are not. Do I really need _all_ Python versions if I have many repos locally, or `3.8.1` is enough for all `>=3.8,<3.9` projects?

Specific binary builds like `3.8.1` as de-facto virtualenvs are problematic too, as long as they will be shared by 2+ projects that decided to pin that exact version for no reason. Can you find an ironic equivalent of such a dependency mess created by `.env` files without any way to opt out of this madness?

clbarnes:
I completely agree with all of your points about why projects shouldn't commit `.python-version` to source control. IMO, the solution is to not do that, and for developers who understand what the file is for to raise PRs with well-reasoned and objectively correct descriptions.

native-api:
> You already can opt out by setting the `PYENV_VERSION` envvar which overrides any version derived from CWD.

If you wish a more permanent approach of being a rebel and using the codebase contrary to its maintainer's wishes (while still using the same version control) -- you go the route of https://softwareengineering.stackexchange.com/questions/214614/how-to-maintain-a-forked-git-repo-with-feature-branches-that-upstream-wont-pull like with any other changes that upstream won't accept.

native-api:
Ther's yet another downside to an envvar. It will affect _all_ projects, not just the one that you have a problem with.

native-api:
Closing since you can already do what you want -- and with an envvar, too.

If you want to automatically set it whenever you're in the specific subtree, you can use https://github.com/cxreg/smartcd for that.

PR #1551, opened 2020-03-09
Remove *-latest from python-build

clbarnes:
Reasoning in the attempted discussion issue #1360 . Paraphrased:

It would be nice if calling pyenv install miniconda-{version} produced the same result every time. Pyenv obviously can't guarantee that the upstream artifacts won't change, but one situation where it could attempt to mitigate non-determinism is `miniconda-latest`.

Miniconda-latest is a problem mainly because the resulting version ends up named miniconda-latest on the user's machine. Without the context of when that was installed, the user doesn't actually know what version it is.


clbarnes:
Any particular justification for the close? If you could also stick it in the linked discussion issue for posterity, that would be great.

anton-petrov:
@clbarnes Each user decides for himself the issue of the consistency of his working environment. How about, for example, python 3.10-dev (building from github sources)? We need to remove them too?

clbarnes:
> How about, for example, python 3.10-dev (building from github sources)? We need to remove them too?

I mean, maybe?

>  In any way, this a discussable question - to delete, or not to delete...

Totally agreed! That's why I raised the linked discussion issue. I didn't raise this PR for another 9 months. Exactly as you say, it's discussable, which is why I requested feedback when it was closed without comment.

PR #1570, opened 2020-03-24
Fixes 'set by' info not showing for `pyenv versions --skip-aliases` and new symlink formate in `pyenv versions`

### LINKED ISSUE #1563: Change the ``pyenv versions'' display format for symlink of the env name.
Hi,

Currently, the pyenv versions command will display as follows:

```
werner@ubuntu-01:~/Desktop$ pyenv versions | grep hpc
  3.8.1/envs/hpc
  hpc
```

Here, we have a env named as 3.8.1/envs/hpc and symlinked it as hpc.  So, I think the following display format maybe more intuitive:
  
    hpc --> 3.8.1/envs/hpc 
Regards

### LINKED ISSUE #1569: 'set by' info not showing for `pyenv versions --skip-aliases` if PYENV_VERSION is a virtualenv
If PYENV_VERSION is not a virtualenv then `pyenv versions --skip-aliases` works fine.  
But if PYENV_VERSION is a virtualenv then `pyenv versions --skip-aliases` not showing 'set by' info, since `python versions` shows 'set by' info next to aliases.
##### Sample
```
nkpro@SR:~$ pyenv shell venv37
(venv37) nkpro@SR:~$ pyenv versions
  system
  3.6.10
  3.7.7
  3.7.7/envs/venv37
* venv37 (set by PYENV_VERSION environment variable)
(venv37) nkpro@SR:~$ pyenv versions --skip-aliases
  system
  3.6.10
  3.7.7
  3.7.7/envs/venv37
```

nkpro2000sr:
closes  #1563, closes #1569  
Envs and Path is merged to symlink format ( Env -> OrginalPath ).  
```
nkpro@SR:~$ pyenv versions
* system  (set by PYENV_VERSION environment variable)
  3.6.10 
  3.7.7 
  venv37 -> 3.7.7/envs/venv37
nkpro@SR:~$ pyenv shell venv37
(venv37) nkpro@SR:~$ pyenv versions
  system 
  3.6.10 
  3.7.7 
* venv37 -> 3.7.7/envs/venv37 (set by PYENV_VERSION environment variable)
(venv37) nkpro@SR:~$ pyenv versions --skip-aliases
  system 
  3.6.10 
  3.7.7 
* 3.7.7/envs/venv37 (set by PYENV_VERSION environment variable)
```
> output of `pyenv versions --bare` and `pyenv versions --bare --skip-aliases` are same as before.

### DIFF
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -76,9 +76,11 @@ else
   IFS=: current_versions=($(pyenv-version-name || true))
   IFS="$OLDIFS"
   include_system="1"
+  print_symlink="1"
 fi
 
 num_versions=0
+declare -A has_aliases
 
 exists() {
   local car="$1"
@@ -93,10 +95,10 @@ exists() {
 }
 
 print_version() {
-  if exists "$1" "${current_versions[@]}"; then
-    echo "${hit_prefix}$1 (set by $(pyenv-version-origin))"
+  if exists "$1" "${current_versions[@]}" || exists "$3" "${current_versions[@]}"; then
+    echo "${hit_prefix}$1${print_symlink+"${@:2}"} (set by $(pyenv-version-origin))"
   else
-    echo "${miss_prefix}$1"
+    echo "${miss_prefix}$1${print_symlink+"${@:2}"}"
   fi
   num_versions=$((num_versions + 1))
 }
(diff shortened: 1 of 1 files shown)


nkpro2000sr:
Yet to change tests.

hongyi-zhao:
This fixes still has the following problems:  When using the tab-completion for pyenv shell/local/global, the entries in the completion list still include all of them.  I think maybe filter out the symlink's source.

Regares

hongyi-zhao:
@nkpro2000sr I tried this commit:https://github.com/pyenv/pyenv/pull/1570/commits/77efd5ebc22fbb1d16583ddf7006e6563eec56e4

It cannot fix the problem of tab-completion discussed above:

```
$ pyenv local <tab>
3.8.1/envs/hpc
[...]
hpc
[...]
```

Regards

nkpro2000sr:
> This fixes still has the following problems: When using the tab-completion for pyenv shell/local/global, the entries in the completion list still include all of them. I think maybe filter out the symlink's source.
> 
> Regares

but tab-completion has to show all the possibilities.  
Though `pyenv global 3.7` and `pyenv global 3.7.7/envs/3.7` are the same,  
still we can do both.
maybe setting an environmental variable for that will be nice :+1: .

what do you think?

nkpro2000sr:
[this commit](https://github.com/pyenv/pyenv/pull/1570/commits/77efd5ebc22fbb1d16583ddf7006e6563eec56e4) is for an bug that `-->` is include is tab-completion suggestion.

hongyi-zhao:
> > This fixes still has the following problems: When using the tab-completion for pyenv shell/local/global, the entries in the completion list still include all of them. I think maybe filter out the symlink's source.
> > Regares
> 
> but tab-completion has to show all the possibilities.
> Though `pyenv global 3.7` and `pyenv global 3.7.7/envs/3.7` are the same,
> still we can do both.
> maybe setting an environmental variable for that will be nice +1 .

I thinks it will be more usable to let the tab-completion displays in the same manner as the ''pyenv versions'' do.  But as you told above, the tab-completion has to show all the possibilities, it seems my suggestion is impossible with the tab-completion method.

Regards 




> 
> what do you think?

hongyi-zhao:
> [this commit](https://github.com/pyenv/pyenv/pull/1570/commits/77efd5ebc22fbb1d16583ddf7006e6563eec56e4) is for an bug that `-->` is include is tab-completion suggestion.

I've checked out this.
Thanks.

nkpro2000sr:
> Another suggestion:
> Or let the tab-completion display the symlinked env-name as following:
> `symlink(-->symlink-src)`
> 
> Regards

I think the display_command and the actual_command must be the same.  
if we display 3.7(-->3.7.7/envs/3.7), then tab-completion will fill the same (eg: `pyenv global 3.7(-->3.7.7/envs/3.7)`) which leads to syntax error.

Maybe there will be any other way to make it possible :smiley_cat: .  
If you know how to do this. I am glad to implement it :+1: .

hongyi-zhao:
@nkpro2000sr I also have no idea to do that.  But fortunately, the currently fix by you already seems pretty good.

neersighted:
I personally am opposed to these changes as they break the current API surface. I for one script `pyenv` using the `--skip-aliases` option, and its removal would be quite disruptive to me, and presumably many other users.

hongyi-zhao:
@neersighted Just do little adjustments for your case, it's easy, from my point of view.

neersighted:
What is disruptive about removing an option/changing default behavior without any sort of deprecation period is fragmentation. My scripts would have to check for both options over a period of several months at least -- and I can imagine the half-life of some Docker images is even longer. As such I really dislike breaking the existing API.

nkpro2000sr:
@neersighted Mostly API calls use '--bare' and i didn't changed it except skipping aliases. which is no need since real path is showing for '--bare' (i think).

nkpro2000sr:
New approach without breaking the existing API :smile_cat:

hongyi-zhao:
@nkpro2000sr It seems this method is buggy and problematic for the case I posted here:

https://github.com/pyenv/pyenv/issues/1592

Anyway, it seems the currently used method in pyenv is more robust.

Regards

nkpro2000sr:
> @nkpro2000sr It seems this method is buggy and problematic for the case I posted here:
> 
> #1592
> 
> Anyway, it seems the currently used method in pyenv is more robust.
> 
> Regards

I read your issue, but I don't think the error is because of this method.  
Could you try again with the original [pyenv-versions](https://github.com/pyenv/pyenv/blob/master/libexec/pyenv-versions).  
> replace the current pyenv-versions in .pyenv/libexec with the original one.

hongyi-zhao:
@nkpro2000sr It seems the original pyenv-versions hasn't this problem

Regards

PR #1576, opened 2020-04-02
Add sed pipeline to strip colored 'ls' command output

juniac:
Add sed pipeline to strip colored 'ls' command output on macos color enabled 'ls' (CLICOLOR, CLICOLOR_FORCE)
and use original 'ls' command


alivx (changes requested):
Both lines give the same output, I guess there is no need to add extra complexity.

jaceppo (approved):

jaceppo (approved):

PR #1586, opened 2020-04-12
pypy 7.3.1 support

pthariensflame:
Adds support for PyPy 7.3.1, in both source and binary forms, and in both 2.7 and 3.6 variations.  This was adapted directly from the 7.3.0 support scripts.


gaborbernat:
Any updates on this?

PR #1593, opened 2020-04-20
Add CPython v2.7.18.

felicianotech:
Adding the Python v2.7.18 release. Bats tests still pass, tested the install in an Ubuntu 19.10 VM.


joshfriend:
Thanks! ❤️

PR #1596, opened 2020-04-26
Remove Ruby references from python-build

singingwolfboy:


joshfriend:
@yyuu @blueyed does this change make it more difficult to merge from rbenv?

yyuu:
I think they are imported from https://github.com/rbenv/ruby-build, and, we have left them as is since they are dead code path from python-build’s point of view. These dead lines maybe help generating meaningful diff when importing changes from ruby-build, but maybe not. I an okay to remove them all.

PR #1598, opened 2020-04-28
Add ZPlug installation instructions

andypalmer:
### Description
This PR updates the documentation to include installation via ZPlug plugin.

The plugin at https://gitlab.com/RiverGlide/zsh-pyenv.git is pretty straightforward, and would be a good candidate for bringing under the pyenv organisation.
It uses git submodules to checkout pyenv, and the plugin file itself adds pyenv and pyenv-install to the path as well as calling `eval $(pyenv init -)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,19 @@ If you're on Windows, consider using @kirankotari's [`pyenv-win`](https://github
 Visit my other project:
 https://github.com/pyenv/pyenv-installer
 
+### via ZPlug plugin manager for Zsh
+
+Add the following line to your `.zshrc`:
+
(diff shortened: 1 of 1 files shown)


joshfriend:
Not going to merge this because the plugin is not maintained or used by any of the pyenv maintainers. I also see that it installs pyenv as a git submodule and is therefore locked to a specific version of the repo (which we cant change)

I've created a new wiki page for the wild west of alternate install methods, and copied your suggestion into it: https://github.com/pyenv/pyenv/wiki/Alternate-Installation-Methods

PR #1599, opened 2020-04-28
improve the flow of macOS installation instructions

dmzobel:
# Description
This update removes several clicks from installation workflow of new MacOS users. I've seen users encounter problems because they did not fully complete the install steps; this update endeavors to help the user do the right thing.


joshfriend:
Good call on placing that at the front of the instructions

Issue #1610, opened 2020-05-14
Request: add Python 3.8.3


bersbersbers:
See https://github.com/python/cpython/releases/tag/v3.8.3

joshfriend:
#1612

PR #1612, opened 2020-05-14
Add CPython 3.8.3

gotche:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1610

### Description

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/3.8.3
+++ b/plugins/python-build/share/python-build/3.8.3
@@ -1,0 +1,10 @@
+#require_gcc
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-1.1.0j" "https://www.openssl.org/source/old/1.1.0/openssl-1.1.0j.tar.gz#31bec6c203ce1a8e93d5994f4ed304c63ccf07676118b6634edded12ad1b3246" mac_openssl --if has_broken_mac_openssl
+install_package "readline-8.0" "https://ftpmirror.gnu.org/readline/readline-8.0.tar.gz#e339f51971478d369f8a053a330a190781acb9864cf4c541060f12078948e461" mac_readline --if has_broken_mac_readline
+if has_tar_xz_support; then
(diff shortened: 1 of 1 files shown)


m-aciek:
@gotche Thank you!

@joshfriend Are you planning a new release in a near future?

gotche:
Hey @m-aciek , you don't really need to wait for a release. If you install pyenv using the installer, and get all the plugins, you can use 

```bash
pyenv update
```

to pull the latest goodies

m-aciek:
@gotche I have homebrew's pyenv :(

But thank you for the message.

robertleeblairjr:
Same. Sounds like a nice suggestion, if it's possible, to add this function `pyenv update` for use with the Homebrew install.

gotche:
@m-aciek @robertleeblairjr I have never used homebrew (or even a Mac, so please bear that in mind), what is preventing you from installing pyenv directly from source?

joshfriend:
Homebrew version of pyenv is provided as a convenience for our users. It's a bit annoying to have to tag releases and update the formula every time a new patch version of python comes out so I hope you can understand why it lags behind sometimes!

I think there is a way to install via homebrew where it just clones the pyenv repo direclty instead of using the zip of the tag? "Head only formula" or something is what it's called I think.

I always recommend the github checkout method for installation

heb-mm:
I'm on Brew's pyenv. How can I get update so that I can use 3.8.3?

MrPaulAR:
Remove the brew version and then install from source.

davidszotten:
> Homebrew version of pyenv is provided as a convenience for our users. It's a bit annoying to have to tag releases and update the formula every time a new patch version of python comes out so I hope you can understand why it lags behind sometimes!

This is completely understandable!

The readme currently suggests installing pyenv using homebrew for os x. should we change that recommendation? or recommend installing `--head` (does master tend to be stable enough for that)

longer term, would it be feasible to split the installation recipes from the bundle so they can change without releasing a new version?

joshfriend:
I've frowned upon homebrew installation for pyenv ever since homebrew maintainers forced everyone to remove the caveats section from their formulae which is how those users were previously reminded to add `pyenv-init` to their shells.

> does master tend to be stable enough for that

Probably, releases just happen whenever things people wanted get merged and one of us gets spare time.

> would it be feasible to split the installation recipes from the bundle

nodenv took that route, and as a user I was always frustrated that after installing nodenv, I would not be able to install any versions of node. If the versions list was auto installed with pyenv, it would probably be fine though. I really honestly like that idea.

PR #1644, opened 2020-06-29
make zsh completion works no need source in zshrc.

roachsinai:
Just put it on $fpath (Archlinux did it in default for pyenv).

By the way, this commit make pyenv zsh completion support [fzf-tab](https://github.com/Aloxaf/fzf-tab) @Aloxaf .

image

### Description

This PR seems only involved to zsh completion.

### DIFF
--- a/completions/pyenv.zsh
+++ b/completions/pyenv.zsh
@@ -1,13 +1,28 @@
 #compdef pyenv
+if [[ ! -o interactive ]]; then
+    return
+fi
 
-_pyenv() {
-  local -a comples
-  if [ "${#words}" -eq 2 ]; then
-    comples=($(pyenv commands))
-  else
-    comples=($(pyenv completions ${words[2,-2]}))
-  fi
-  _describe -t comples 'comples' comples
-}
+local state line
+typeset -A opt_args
 
-_pyenv
+_arguments -C \
+    {--help,-h}'[Show help]' \
+    {--version,-v}'[Show pyenv version]' \
+    '(-): :->command' \
+    '*:: :->option-or-argument'
+
+case "$state" in
+    (command)
+        local -a commands
+        commands=(${(f)"$(pyenv commands)"})
+        _describe -t commands 'command' commands
+        ;;
+    (option-or-argument)
+        local -a args
+        args=(${(f)"$(pyenv completions ${line[1]})"})
+        _describe -t args 'arg' args
+        ;;
+esac
+
+return 0
--- a/completions/pyenv.zsh
+++ b/completions/pyenv.zsh
@@ -1,18 +1,13 @@
-if [[ ! -o interactive ]]; then
-    return
-fi
-
-compctl -K _pyenv pyenv
+#compdef pyenv
 
 _pyenv() {
-  local words completions
-  read -cA words
-
+  local -a comples
   if [ "${#words}" -eq 2 ]; then
-    completions="$(pyenv commands)"
+    comples=($(pyenv commands))
   else
-    completions="$(pyenv completions ${words[2,-2]})"
+    comples=($(pyenv completions ${words[2,-2]}))
   fi
-
-  reply=(${(ps:\n:)completions})
+  _describe -t comples 'comples' comples
 }
+
+_pyenv


adigitoleo:
According to the discussion in #1458 the removal of the interactive check apparently breaks completion for people who use `git clone` or `pyenv-installer` instead of proper packages...

The following alternative is a bit more modern and works fine even after running `eval "$pyenv init -)"` from `.zshrc`

```zsh
#compdef pyenv
if [[ ! -o interactive ]]; then
    return
fi

local state line
typeset -A opt_args

_arguments -C \
    {--help,-h}'[Show help]' \
    {--version,-v}'[Show pyenv version]' \
    '(-): :->command' \
    '*:: :->option-or-argument'

case "$state" in
    (command)
        local -a commands
        commands=(${(f)"$(pyenv commands)"})
        _describe -t commands 'command' commands
        ;;
    (option-or-argument)
        local -a args
        args=(${(f)"$(pyenv completions ${line[1]})"})
        _describe -t args 'arg' args
        ;;
esac

return
```

Also, it would be good to remove `--version` from the `pyenv commands` output so that it's not suggested twice:

```
❯ pyenv commands
commands
completions
exec
global
help
hooks
init
install
local
prefix
rehash
root
shell
shims
uninstall
--version
version
version-file
version-file-read
version-file-write
version-name
version-origin
versions
whence
which
```

native-api:
@adigitoleo You can suggest your version as a PR to either of the PRs. To do that, specify the PR branch as "base branch" when creating the PR.

adigitoleo:
@native-api Thanks, let me know if I did something wrong (still learning github).

native-api (commented):
@adigitoleo Okay, I've merged your PR.
Will this work for a clean Pyenv install according to the instructions in the README?

I have little experience with Zsh and am currently under impression that `#compdef` needs to be used completely differently from traditional completion scripts: it should not (maybe even _cannot)_ be sourced directly but should rather be placed into some special directory specified in Zsh config or something.

If that's true, this PR is going to break the completion as it's now.

adigitoleo:
@native-api Ah, you're right, good catch. This won't work for `pyenv-init` (not sure why I thought it would). We can pick one of:
- special case for zsh in `pyenv-init`
- let it be, and packagers will need to patch the completion file

It's a bit awkward, but I think the second option is probably best after all.

native-api:
I didn't quite understand what you mean:

>     * let it be, and packagers will need to patch the completion file

Do "packagers" include _us?_ Keep in mind that Pyenv is designed to be usable from a Git checkout.

I reckon if this new completion mechanism is incompatible with our needs, this just means we shouldn't be using it, right?

native-api:
> Do "packagers" include _us?_

Even if they don't, it's not a good thing to make life harder for packagers without a good reason. (And "just to use a new fancy tech that's not really any different from the old one" hardly qualifies.)

adigitoleo:
> Do "packagers" include _us?_

No, I meant to say distribution/OS packagers, sorry for the confusion.
 
> Even if they don't, it's not a good thing to make life harder for packagers without a good reason. (And "just to use a new fancy tech that's not really any different from the old one" hardly qualifies.)

Unfortunately, the current completion script does make life harder for _distro_ packagers. For example, the current Arch linux package installs the completion scripts into standard system directories, expecting them to be picked up by the respective shell's completion system. It seems that this isn't the case for the provided zsh completion script, which is what motivated me to suggest the version from my earlier comment.

However, since my version breaks for the recommended installation method, I am happy for this to be closed.

native-api:
Whatever the proposed solution is, it cannot break the use cases that we ourselves recommend.

If the `compsys` way has some additional benefits, the proposition should support both.

ckp95:
Is there a solution for this? I really want to have the fzf-tab completion work with pyenv.

native-api:
As specified earlier, we need to support dynamic load of completions. If there's a way to plug a configuration file into `compsys` dynamically, we can very well use that.

native-api:
The best way would be to support both static and dynamic load of completions.

onaforeignshore:
I wanted to use the modern version to have selectable completion of the options and available versions, so I researched this and came up with the code below. The installation of this version is explained in the steps below.

I have diverted from your completion in the following way:

- only use pyenv to determine virtualenvs, available versions and installed versions (this is done once per session)
- added the descriptions from the command line help for each option
- stopped autocompletion when no more logical options or flags were required

This requires the `_pyenv` file to be on the `fpath` and to be invoked using `autoload` (see below)

I am happy with the result and am sharing it here for others to use, even if it will not be added to the project.

```bash
#compdef pyenv
# ------------------------------------------------------------------------------
#
# Description
# -----------
#  Completion script for the pyenv command
#  (https://github.com/pyenv/pyenv)
#
# Authors
# -------
#
#  * Christo Kotzé <onaforeignshore@gmail.com>
#
# ------------------------------------------------------------------------------

# pyenv currently uses compctl completion.
# To use this file instead, you need to do the following:
# 1. Download this file and place it in a folder
# 2. Make sure that this folder is available on the fpath
# > fpath=(/new/folder $fpath)
# 3. Remove the old completion system
# > unfunction _pyenv
# 4. Load this completion file
# > autoload -U _pyenv
#
# Note:
# You can add steps 2-4 in your .zshrc file to automate this.

typeset -A opt_args
local state cmd
local curcontext="$curcontext"

# populate lists
(( ${+virtualenv} )) ||
virtualenvs=(${(f)"$(pyenv virtualenvs --bare)"})

(( ${+availableVersions} )) || {
    availableVersions=(${(f)"$(pyenv install --list)"#  })
    shift availableVersions # remove 'Available versions:' header
    availableVersions=( "${availableVersions[@]//((#s)[[:space:]]##|[[:space:]]##(#e))/}" ) # remove leading and trailing spaces
}

(( ${+installedVersions} )) ||
installedVersions=(${(f)"$(pyenv versions --bare)"})


_pyenv() {
    local ret=1
    
    _arguments -A "-*" \
        '(- *)--help[Display help for a command]' \
        '1: :_pyenv_subcommands' \
        '*::arg:->args' \
    && ret=0
    
    if [[ "$state" == "args" ]]; then
        local -a cmds=(
            activate help install latest local shell shims uninstall
            versions virtualenv virtualenv-delete virtualenvs whence which
        )
        cmds=( ${(M)cmds:#${words[1]}*} )
        if (( ${#cmds} == 1 )) || [[ $cmds[1] = install ]]; then
            cmd="$cmds[1]"
            curcontext="${curcontext%:*:*}:pyenv-${cmd}:"
            _call_function - _pyenv_${cmd//-/_} || _message 'no more arguments'
        else
            _message "unknown pyenv command: $words[1]"
        fi
    fi
    
    return ret
}

#####
# Helper completers
#

(( $+functions[_pyenv_subcommands] )) ||
_pyenv_subcommands() {
    local -a commands=(
        '--version:Display the version of pyenv'
        'activate:Activate virtual environment'
        'commands:List all available pyenv commands'
        'deactivate:Deactivate virtual environment'
        'exec:Run an executable with the selected Python version'
        'global:Set or show the global Python version(s)'
        'help:Display help for a command'
        'hooks:List hook scripts for a given pyenv command'
        'init:Configure the shell environment for pyenv'
        'install:Install a Python version using python-build'
        'latest:Print the latest installed or known version with the given prefix'
        'local:Set or show the local application-specific Python version(s)'
        'prefix:Display prefixes for Python versions'
        'rehash:Rehash pyenv shims (run this after installing executables)'
        'root:Display the root directory where versions and shims are kept'
        'shell:Set or show the shell-specific Python version'
        'shims:List existing pyenv shims'
        'uninstall:Uninstall Python versions'
        'version:Show the current Python version(s) and its origin'
        'version-file:Detect the file that sets the current pyenv version'
        'version-name:Show the current Python version'
        'version-origin:Explain how the current Python version is set'
        'versions:List all Python versions available to pyenv'
        'virtualenv:Create a Python virtualenv using the pyenv-virtualenv plugin'
        'virtualenv-delete:Uninstall a specific Python virtualenv'
        'virtualenv-init:Configure the shell environment for pyenv-virtualenv'
        'virtualenv-prefix:Display real_prefix for a Python virtualenv version'
        'virtualenvs:List all Python virtualenvs found in `$PYENV_ROOT/versions/*`.'
        'whence:List all Python versions that contain the given executable'
        'which:Display the full path to an executable'
    )
    [[ $debug ]] && commands+="$curcontext"
    _describe -t commands 'command' commands "$@"
}

(( $+functions[_pyenv_help] )) ||
_pyenv_help() {
    _arguments -A "-*" \
    '(- *)--help[Display help for a command]' \
    '1: :_pyenv_subcommands'
}

(( $+functions[_pyenv_activate] )) ||
_pyenv_activate() {
    (( $CURRENT == 3 )) && return
    [[ "${words[2]}" == "--unset" ]] && return
    local -a args=(
        '--unset[Unsets the activated virtualenv]'
    )
    _arguments -S -A "-*" $args[@]
    _describe -t versions 'versions' virtualenvs "$@"
}

(( $+functions[_pyenv_install] )) ||
_pyenv_install() {
    if  [[ "${words[2]}" == "--list" ]] || \
        [[ "${words[2]}" == "--version" ]] || \
        [[ "${words[2]}" == "-l" ]] || 
        [[ "${words[2]}" == "-v" ]]; then
            return
    fi
    local -a args=(
        '(-l --list)'{-l,--list}'[List all available versions]'
        '(-f --force)'{-f,--force}'[Install even if the version appears to be installed already]'
        '(-s --skip-existing)'{-s,--skip-existing}'[Skip if the version appears to be installed already]'
        '(-k --keep)'{-k,--keep}'[Keep source tree in $PYENV_BUILD_ROOT after installation (defaults to $PYENV_ROOT/sources)]'
        '(-p --patch)'{-p,--patch}'[Apply a patch from stdin before building]'
        '(-v --verbose)'{-v,--verbose}'[Verbose mode: print compilation status to stdout]'
        '(-g --debug)'{-g,--debug}'[Build a debug version]'
        '--version[Show version of python-build]'
    )
    _arguments -S -s $args[@]
    _describe -t versions 'versions' availableVersions "$@"
}

(( $+functions[_pyenv_latest] )) ||
_pyenv_latest() {
    local -a args=(
        '(-k --known)'{-k,--known}'[Select from all known versions instead of installed]'
    )
    local -a commands=(
        '2'
        '3'
    )
    _arguments -S $args[@]
    _describe -t commands 'command' commands "$@"
}

(( $+functions[_pyenv_local] )) ||
_pyenv_local() {
    if [[ $words[2] != "--unset" ]]; then
        local -a args=(
            '(-f --force)'{-f,--force}'[Do not verify that the versions being set exist]'
            '--unset[Unset the current version]'
        )
        _arguments -S $args[@]
        _describe -t versions 'version' installedVersions "$@"
    fi
}

(( $+functions[_pyenv_shell] )) ||
_pyenv_shell() {
    if [[ $words[2] != "--unset" ]]; then
        local -a args=(
            '-[Restore previously set version]'
            '--unset[Unset `PYENV_VERSION` (restoring environment state to before first `pyenv shell` call)]'
        )
        _arguments -S $args[@]
        _describe -t versions 'version' virtualenvs "$@"
    fi
}

(( $+functions[_pyenv_uninstall] )) ||
_pyenv_uninstall() {
    local -a args=(
        '(-f --force)'{-f,--force}'[Do not verify that the versions being set exist]'
    )
    _arguments -S $args[@]
    _describe -t versions 'version' installedVersions "$@"
}

(( $+functions[_pyenv_versions] )) ||
_pyenv_versions() {
    local -a args=(
        '--bare[Show paths only]'
        '--skip-aliases[Do not show versions that are aliases]'
        '--skip-envs[Do not show versions that are envs]'
    )
    _arguments -S $args[@]
}

(( $+functions[_pyenv_virtualenv] )) ||
_pyenv_virtualenv() {
    local -a args=(
        '(-f --force)'{-f,--force}'[Install even if the version appears to be installed already. Skip prompting for confirmation]'
        '--version[Show version of pyenv-virtualenv]'
    )
    _arguments -S $args[@]
}

(( $+functions[_pyenv_virtualenv_delete] )) ||
_pyenv_virtualenv_delete() {
    local -a args=(
        '(-f --force)'{-f,--force}'[Attempt to remove the specified virtualenv without prompting for confirmation]'
    )
    _arguments -S $args[@]
    _describe -t versions 'versions' virtualenvs "$@"
}

(( $+functions[_pyenv_virtualenvs] )) ||
_pyenv_virtualenvs() {
    local -a args=(
        '--bare[Show paths only]'
        '--skip-aliases[Do not show versions that are aliases]'
    )
    _arguments -S $args[@]
}

(( $+functions[_pyenv_whence] )) ||
_pyenv_whence() {
    local -a args=(
        '(- *)--help[Display help for a command]'
        '--path[List all Python versions that contain the given executable (show full path)]'
    )
    _arguments -S $args[@]
}

(( $+functions[_pyenv_which] )) ||
_pyenv_which() {
    local -a args=(
        '(- *)--help[Display help for a command]'
        '--nosystem[when you don''t need to search command in the
        system environment]'
    )
    _arguments -S $args[@]
}

_pyenv "$@"
```

PR #1646, opened 2020-07-01
allow tcl-tk as argument or try with homebrew by default Replacement for #1409

tillhainbach:

### Prerequisite
  -#1125
For anyone trying to use tcltk from homebrew before (or if you had tcltk installed from homebrew tkinter would fail I believe)
#1125
- replaces #1409 (and me being stupid in #1645)

### Description
From @grahamannett in #1409:
"I previously added the regex match for this, but having to pass the config opts everytime while other packages use homebrew by default made this seem worthwhile. If there is anything that would make this more streamlined, or a better place to put it in python-build then let me know.
This should hopefully allow more people to use pyenv and things like matplotlib or whatever that may previously had issues."

### Tests
- [x] My PR adds the following unit tests (if any):
Me: In addition to @grahamannett implementation of the option I fixed the tests in plugin/python-build/test/build.bats and added two additional test for the new option (analogous to the test used for the readline option)


grahamannett:
wow. thanks so much for doing this!  i think those tests were exactly what I was trying to figure out but never got around to it

tillhainbach:
No, if one wants to be 100% sure that tkinter works correctly one has to actually call `python -m tkinter` (you don’t need the -c stuff) and check that the Windows looks ok. The tests only verify that the python-build scripts sets the appropriate flags. From there, one has to trust that it gets links properly by the building script. In my opinion, anything more (like checking that it is actually link) would be beyond the scope of this repo.

grahamannett:
this is such a much needed PR!

grahamannett:
@joshfriend this is a really needed PR for people who might not know how to trouble shoot this issue and fixes the tests you had asked me to fix in https://github.com/pyenv/pyenv/pull/1409.

xrisk:
can this be merged please?

plugins/python-build/bin/python-build
```diff
@@ -1486,6 +1487,29 @@ use_xcode_sdk_zlib() {
...
 
+use_tcltk() {
+   # check args passed, use that otherwise check for homebrew tcl-tk
+   local tcltk_libdir="$(brew --prefix tcl-tk 2>/dev/null || true)"
+   local regex_tcltk_libs_args="(--with-tcltk-libs='([^']+)')"
+
+   if [[ "$PYTHON_CONFIGURE_OPTS" =~ $regex_tcltk_libs_args ]]; then
+     local tcltk_libs_match="${BASH_REMATCH[1]}"
+     local tcltk_match_quoted="${tcltk_libs_match//--with-tcltk-libs=/}"
+     PYTHON_CONFIGURE_OPTS="${PYTHON_CONFIGURE_OPTS//$tcltk_libs_match/}"
```
native-api, line 1498: This extracting-and-readding is limiting and unreliable. In particular, it requires exactly quoted single quotes and exactly around the value -- which is something one has no way to know or expect.

Better properly parse the configure values into an array, e.g. with `xargs` as per https://stackoverflow.com/questions/37372225/convert-a-string-into-an-array-with-bash-honoring-quotes-for-grouping.

plugins/python-build/bin/python-build
```diff
@@ -1486,6 +1487,29 @@ use_xcode_sdk_zlib() {
...
+
+     echo "python-build: use tcl-tk from \$PYTHON_CONFIGURE_OPTS"
+     package_option python configure "--with-tcltk-libs='${tcltk_match_quoted}'"
+   elif [ -d "$tcltk_libdir" ]; then
+     # get the version from the folder that homebrew versions
+     local tcltk_version_long="$(ls "$(brew --cellar tcl-tk)")"
+     local tcltk_version="${tcltk_version_long%.*}"
+     local tcltk_flags="--with-tcltk-includes=-I$tcltk_libdir/include --with-tcltk-libs=-L$tcltk_libdir/lib -ltcl$tcltk_version -ltk$tcltk_version"
+     echo "python-build: use tcl-tk from homebrew"
```
native-api, line 1507: Using-from-Homebrew logic should probably be placed into a separate function like with other library searches.

native-api (commented):
Should we really prefer Tcl/Tk from Homebrew? XCode SDK has its own copy, what's the problem with it?

tillhainbach:
Citing from: [python.org](https://www.python.org/download/mac/tcltk/)
>Apple 8.5.9
This release is included in macOS 10.7 through at least macOS 10.14. As of this writing, there are at least two known issues with Tk 8.5.9 that are present in Apple 8.5.9 Tk but fixed in more recent upstream 8.5 releases. The more serious problem is an immediate crash in Tk when entering a composition character, like Option-u on a US keyboard. (This problem is documented as Tk bug 2907388.) There is also the more general problem of input manager support for composite characters (Tk bug 3205153) which has also been fixed in more recent Tcl/Tk 8.5 releases. You can avoid these problems by using a current python.org installer or by using a third-party distribution of Python that does not use Apple 8.5.9 Tk.

plugins/python-build/bin/python-build
```diff
@@ -1516,6 +1517,60 @@ use_xcode_sdk_zlib() {
...
+
+  # remove quotes, because there mess up compilations
+  PYTHON_CONFIGURE_OPTS="${PYTHON_CONFIGURE_OPTS//"''"/}"
+  PYTHON_CONFIGURE_OPTS="${PYTHON_CONFIGURE_OPTS//'""'/}"
+
+  echo "python-build: use tcl-tk from \$PYTHON_CONFIGURE_OPTS"
+  # echo "PYTHON_CONFIGURE_OPTS=${PYTHON_CONFIGURE_OPTS}"
+  package_option python configure --with-tcltk-libs="${tcltk_libs}"
+  IFS="$OLDIFS"
+}
```
native-api, line 1560: Why is this extraction-and-readding needed again?  We only need to detect if the user provided a custom value -- if they did, we don't do anything, if they didn't, we add a default.

plugins/python-build/bin/python-build
```diff
@@ -1516,6 +1517,60 @@ use_xcode_sdk_zlib() {
   fi
 }
 
+use_homebrew_tcltk() {
+  # get the version from the folder that homebrew versions
+  local tcltk_version_long="$(ls "$(brew --cellar tcl-tk)")"
```
native-api, line 1522: This looks like it would break horribly if Homebrew isn't present (e.g. in Linux).

native-api (changes requested):

plugins/python-build/bin/python-build
native-api: Actually, such a situation could use a test.

plugins/python-build/bin/python-build
tillhainbach: See #1125. 
Gist: The way python-build works write now is that it somehow does not pass multipart options correctly to python configure.
But I guess you are correct that a better way would be to fix the parsing of PYTHON_CONFIGURE_OPTS for any multipart options (not just tcl-tk).

plugins/python-build/bin/python-build
tillhainbach: This function is only called by `use_tcltk` if `local tcltk_libdir="$(brew --prefix tcl-tk 2>/dev/null || true)"` is defined which entails that the `brew`-command is available on the machine. If someone tries to call that function anywhere else it would break dozens of tests.

plugins/python-build/bin/python-build
tillhainbach: @native-api: How should we proceed about this? From my point of view fixing the general issue of #1125 is out of scope of this PR, however it solves 2 things:
(1) provide a working (e.g. homebrew) default for tcl-tk libs 
(2) improve the work-around for the problem described in #1125 by reliably parsing `PYTHON_CONFIGURE_OPTS` using `xargs` explicitly for tcl-tk. The old workaround for passing multipart tcl-tk to `python-build` used regex and was merge in #1397.

Since tcl-tk from homebrew is a long requested feature, I would argue in favour of merging this PR and opening an additional PR for refactoring the parsing of `PYTHON_CONFIGURE_OPTS` using `xargs`. I can set up a new draft PR for it, summarising the relevant information there.

native-api: > From my point of view fixing the general issue of #1125 is out of scope of this PR

Okay, fair enough

native-api: Let's settle on this then, shall we?

* Make the detection reliable (current regex is not) and split the detection part and reinsertion part
* Add a comment to the extract-and-reinsert code that it's a workaround for #1125 and should be removed once it's fixed.
* Resolve the conflicts with master

Then we can merge this.

tillhainbach: Just for clarification:

> Make the detection reliable (current regex is not) and split the detection part and reinsertion part

What do you mean by that?
Currently, this just checks if the tcl-tk-flag `"--with-tcltk-libs="` is a substring of `$PYTHON_CONFIGURE_OPTS`
```
if [[ "$PYTHON_CONFIGURE_OPTS" =~ $tcltk_ops_flag ]]; then
...assume the user has set sensible values for tcl-tk-libs
```

Should I rewrite it to use `*`wildcards? `if [[ "$PYTHON_CONFIGURE_OPTS" == *"$tcltk_ops_flag"* ]];`

native-api: > What do you mean by that?

Detect if any of the args that `$PYTHON_CONFIGURE_OPTS` would be split into (honoring quotes) would be `--with-tcltk-libs=<whatever>`. The regex only detects if any of them contains this substring anywhere.

I admit that the chance of this substring randomly occurring elsewhere is pretty low...

native-api (approved):

PR #1651, opened 2020-07-03
Fix #1647 (Incorrect checksum for Python 3.7.8 gzip package)

hadret:
This is a minor copy/paste issue where checksum for the file was not bumped up for the given version (and only for tgz archive).

Closes #1647

Issue #1647, opened 2020-07-02
Incorrect checksum for Python 3.7.8 gzip package


kkietava:
The checksum for Python 3.7.8 gzip package is incorrect. It's actually a copy-paste leftover from the previous version. When trying to install 3.7.8, I get this error:

    checksum mismatch: Python-3.7.8.tar.gz (file is corrupt)
    expected 8c8be91cd2648a1a0c251f04ea0bb4c2a5570feb9c45eaaa2241c785585b475a, got 
    0e25835614dc221e3ecea5831b38fa90788b5389b99b675a751414c858789ab0

Afer downloading the package from https://www.python.org/downloads/release/python-378/ and comparing its MD5 checksum, I can verify that the actual checksum in the installation error message is correct:

    $ md5sum -c <<< '4d5b16e8c15be38eb0f4b8f04eb68cd0  Python-3.7.8.tgz'
    Python-3.7.8.tgz: OK

    $ sha256sum Python-3.7.8.tgz
    0e25835614dc221e3ecea5831b38fa90788b5389b99b675a751414c858789ab0  Python-3.7.8.tgz

PR #1652, opened 2020-07-03
Fix has_tar_xz_support function on FreeBSD.

hadret:
While working on a fix for #1647 (PR #1651) I noticed that tgz archive was used even though my FreeBSDs tar is capable of operating with J flag handling xz archives. Reason for that is that the current implementation of the `has_tar_xz_support` function returns always 1 due to the following error:

```
~ tar Jc /dev/null
tar: Failed to open '/dev/sa0'
```

Easy fix that works universally on all(? I checked Linux, macOS and FreeBSD) platforms is in this PR.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -504,7 +504,7 @@ download_tarball() {
 }
 
 has_tar_xz_support() {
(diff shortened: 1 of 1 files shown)


joshfriend:
❤️ Thanks

PR #1654, opened 2020-07-06
CPython 3.9.0b4

arcivanov:
CPython 3.9.0b4


arcivanov:
@joshfriend there are important seg faults fixed in b4 and this is needed for further testing.

arcivanov:
@yyuu @ghedsouza

Issue #1628, opened 2020-06-10
System python installations are not detected by pyenv


aspiers:
Several major Linux distributions have been providing simultaneous installs of Python 2 and 3 for quite a long time now, including Ubuntu, openSUSE, RHEL, and Mint.  According to #1585, the same holds true for Mac OS 10.15.4, and probably many other Mac OS versions too.  In these cases, typically:

- `/usr/bin/python` symlinks to either `python2` or `python3`, or may not exist at all
- `/usr/bin/python2` provides some version of Python 2
- `/usr/bin/python3` provides some version of Python 3

However, typically `pyenv` will only support use of `/usr/bin/python` as the single system version, so if the symlink is missing it doesn't find any system `python` **[UPDATE: this bit will be fixed if #1655 is merged]**, and even if it's not missing, if both 2 and 3 are present, it will only find one of them.  This has led to bug reports such as #1029, #1366, #1585, and #1613, the number of which alone suggests to me that this is worth addressing.  But in addition to these, there are valid use cases for **[UPDATED to clarify]** pyenv supporting the choice between at least two different *simultaneously installed* system versions:

In #1613 [**@b12thompson**](https://github.com/b12thompson) commented on [June 7, 2020 3:09 AM](https://github.com/pyenv/pyenv/issues/1613#issuecomment-640145379 "2020-06-07T02:09:05Z - Replied by Github Reply Comments"):
> 1.  **System python installations are not detected by pyenv**:
> 
> I was able to partially replicate the error in Linux Mint 19.3. The system Python 2 install was detected but the Python 3 version was not, which for me isn't a big deal. In my opinion, system installations of Python should be reserved for internal OS libraries and third-party packages. The user shouldn't really care about using the system installations of Python.
> 
> I'm not trying to nullify the error by saying that "you shouldn't be using the system installation anyway". Perhaps there are good use cases that I'm missing and don't completely understand. If someone is willing to clarify then we can proceed with an appropriate bug fix.

[...]

Issue #1366, opened 2019-07-04
System Python does not exist


omenos:
### System Info
- [ ] Platform information: __RHEL 8.0__
- [ ] OS architecture: __x86_64__
- [ ] pyenv version: __1.2.12-2-geb68ec94__
- [ ] Python version: __3.7.3, 2.7.16__
- [ ] C Compiler information: __8.2.1__

## Description
Starting with RHEL 8, there is no system level Python distribution by default. There is only a platform specific one not intended to be used by anything but system tools (like `dnf`). This python version is stored in `/usr/libexec`. This is to help prevent users from being locked into a system level Python they don't want to use, and instead install the Python version they want from the AppStreams (like modules in Fedora). Currently that's 3.6.8 and 2.7.15.

On my system I had yet to install either of those global python versions before install `pyenv` and Python 3.7.3 and 2.7.16. A package I installed pulled in the python36 stream and thus installed the `python3` binary and its associated files in `/usr/bin`. If called directly it works just fine, as intended. However, `pyenv` does not seem to recognize its existence.

```bash
$ pyenv versions
* 2.7.16 (set by /home/mrochefort/.pyenv/version)
* 3.7.3 (set by /home/mrochefort/.pyenv/version)

$ pyenv global system
pyenv: system version not found in PATH
```

It would appear from this (not sure how else to debug this) that global python versions installed post-`pyenv` installation aren't picked up by default. I've tried rehashing to no avail.. I can try a uninstall/reinstall of `pyenv`, but just not at the moment.

native-api:
You can set `$PYENV_DEBUG` and [`export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'`](https://wiki-dev.bash-hackers.org/scripting/debuggingtips#making_xtrace_more_useful) to debug.

The error is at https://github.com/pyenv/pyenv/blob/610346802358780ebccafdf112ba0c92b6cbaef0/libexec/pyenv-prefix#L37. But for "`system`", prefix shouldn't matter...

draeath:
This still occurs on RHEL 8.2 (not that I expected that to change)

I believe this is because it is only looking for 'python' and not 'python3' - there is no python installed at all by default on RHEL8, let alone python2.

```
+ [pyenv-versions:105] PYENV_VERSION=system
+ [pyenv-versions:105] pyenv-which python
+ [pyenv-versions:109] shopt -s nullglob
[...]

PR #1655, opened 2020-07-07
"system" python - support cases where python3 is in PATH but not python

draeath:
### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1366
  - https://github.com/pyenv/pyenv/issues/1628

### Description

Simple fixes such that a user who does not have a 'python' binary in their path, but does have a 'python3' binary see their system python in `pyenv versions` and does not receive warnings about no system python installations.

### Tests

N/A

### DIFF
--- a/libexec/pyenv-prefix
+++ b/libexec/pyenv-prefix
@@ -30,7 +30,7 @@ OLDIFS="$IFS"
 { IFS=:
   for version in ${PYENV_VERSION}; do
     if [ "$version" = "system" ]; then
-      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python 2>/dev/null)"; then
+      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python 2>/dev/null)" || PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 2>/dev/null)"; then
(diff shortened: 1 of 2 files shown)


draeath:
fixes #1366
fixes #1628

aspiers:
As mentioned in https://github.com/pyenv/pyenv/issues/1628#issuecomment-655573600, I fully support this PR and think it is a great step in the right direction, although by itself it is not the complete fix for #1628 which I had envisioned.

draeath:
in #1366 it's suggested that this should handle cases of `python2` as well. I did think of this, but decided not to do so (see my comment in that issue for more of an explanation).

However if it's decided this is wanted, it shouldn't be too difficult to add - especially if there's a better way to handle this than the way I've done here.

native-api:
https://github.com/draeath/pyenv/pull/1 PR to this PR fixes the `python2` detection.

draeath:
I've never had a PR inside a PR, so I don't know if I need to do anything more than merge on my end.

native-api:
> 
> 
> I've never had a PR inside a PR, so I don't know if I need to do anything more than merge on my end.

You don't. By merging a nested PR, you push its commits to this PR's branch so they just appear here like any other additional commit to a PR branch.

native-api:
@anton-petrov Could you also close the associated issues?

anton-petrov:
Done.

PR #1672, opened 2020-07-24
Add support for Jython 2.7.2

electricsam:
## Description
- Add plugin to support Jython 2.7.2


plugins/python-build/share/python-build/jython-2.7.2
```diff
@@ -0,0 +1,14 @@
+require_java
+unrequire_python27
+install_jar "jython-2.7.2" "https://repo1.maven.org/maven2/org/python/jython-installer/2.7.2/jython-installer-2.7.2.jar#db48721f97cc3d586761c9adfb24c105" jython
```
joshfriend, line 3: please use a SHA256 checksum

joshfriend (changes requested):

plugins/python-build/share/python-build/jython-2.7.2
electricsam: Updated with the SHA256 checksum

Issue #1473, opened 2019-12-03
PYTHON_BUILD_MIRROR_URL desn't take effect.


hongyi-zhao:
I want to let pyenv use the customized mirror url for downloading the python source code as follows:
 
```
$ export PYTHON_BUILD_MIRROR_URL=https://cdn.npm.taobao.org/dist/python
$ pyenv install -v 2.7.7

But, I found the following result:
$ pgrep -af curl
45088 curl -qsILf https://cdn.npm.taobao.org/dist/python/2983e3cd089b30c50e2b2234f07c2ac4fb8a5df230ab8f2e1133a1d8b208da78
```

Why the env PYTHON_BUILD_MIRROR_URL desn't take effect?

Regards

zhangnew:
use `PYENV_PYTHON_MIRROR_URL` instead, see https://github.com/pyenv/pyenv/pull/1380

hongyi-zhao:
Failed with this variable:
```
$ export PYENV_PYTHON_MIRROR_URL="http://npm.taobao.org/mirrors/python"
$ pyenv install -v 2.7.7
/tmp/python-build.20200326150548.21656 ~/.repo/github.com/pyenv/pyenv.git
Downloading Python-2.7.7.tar.xz...
-> https://www.python.org/ftp/python/2.7.7/Python-2.7.7.tar.xz
```

Mark24Code:
mark

Issue #1377, opened 2019-08-01
Support for pull-through cache-mirror of python packages


DenverJ:
We would like to configure pyenv to use an internal pull-through cache like Artifactory for downloading the python install files. Artifactory is configured with remote URLs and handles caching of objects locally on first request by a client.

PYTHON_BUILD_MIRROR_URL doesn't work for this purpose because it computes the hash of the file and tries to download that. This does not fit with the above use-case because artifactory preserves the original path and filename.

Perhaps another option - PYTHON_BUILD_CACHE_URL which simply replaces http://www.python.org with the specified value (which could be a hostname + path)?

guyskk:
For example download python tarball from https://npm.taobao.org/mirrors/python, it's very useful in China mainland.

gitfourteen:
> For example download python tarball from https://npm.taobao.org/mirrors/python, it's very useful in China mainland.

China is ZERO speed for now.

➜  ~ pyenv install 3.8.0
python-build: use openssl@1.1 from homebrew
python-build: use readline from homebrew
Downloading Python-3.8.0.tar.xz...
-> https://www.python.org/ftp/python/3.8.0/Python-3.8.0.tar.xz
error: failed to download Python-3.8.0.tar.xz

BUILD FAILED (OS X 10.15.1 using python-build 20180424)

Issue #1575, opened 2020-04-01
Specify install path to python tar 


alaverty:
I am trying to use pyenv from behind a corporate firewall :
```
[root@xxxxxxxxx plays]# pyenv install 2.7.8
Downloading Python-2.7.8.tar.xz...
-> https://www.python.org/ftp/python/2.7.8/Python-2.7.8.tar.xz
Installing Python-2.7.8...
```

Is there any way we can specify the path to the python tar for installation? Say for example  :

```
pyenv install https://myserver/Python-2.7.8.tar.xz
```

joshfriend:
You can use `PYTHON_BUILD_MIRROR_URL` as described [here](https://github.com/pyenv/pyenv/blob/089b828b0f92c40ee510e1aafbea29c189bd96e6/plugins/python-build/README.md#special-environment-variables), though there are several other issue reports that this environment variable is not working currently. I have not confirmed these reports myself so I'd encourage you to try anyways 😄 

If you do encounter a bug, please search for the open issues related to this and follow them for updates

alaverty:
Awesome, thanks will give it a go!

Issue #803, opened 2017-01-11
It seems PYTHON_BUILD_MIRROR_URL environment variable doesn't work


andyyoung01:
I installed pyenv through pyenv-installer according [README.rst](https://github.com/yyuu/pyenv-installer/blob/master/README.rst). Then I wanted to install Python-2.7.7 using pyenv, but the network was so slow. 
So I used [the pyenv's official mirror of Python archives](http://yyuu.github.io/pythons/) to download Python archives. I followed the instruction of the page, using command like this:
```bash
$ export PYTHON_BUILD_MIRROR_URL="http://mirrors.sohu.com/python/"
$ pyenv install -v 2.7.7
/tmp/python-build.20170111175043.4891 ~/.pyenv
Downloading Python-2.7.7.tar.xz...
-> https://www.python.org/ftp/python/2.7.7/Python-2.7.7.tar.xz
^Cerror: failed to download Python-2.7.7.tar.xz

BUILD FAILED (CentOS Linux 7 using python-build 20160602)

```
According to the output of the command, pyenv still download the archive from official website. So I cancelled the download with Ctrl-C. Was there any mistake of my configuration?
My computer's environment was like this:
```bash
$ uname -a
Linux centos7 4.5.2 #2 SMP Wed May 4 20:00:06 CST 2016 x86_64 x86_64 x86_64 GNU/Linux
$ pyenv --version
pyenv 1.0.6
```
I also tried other version of pyenv:
```
$ cd $HOME/.pyenv  
$ git fetch --tags
$ git checkout v20160726 
```
And opened a new terminal, issued the same command to install Python-2.7.7. It seems that pyenv still download archives from python official website. How can I change that?

seozed:
Look here!
http://www.tuicool.com/articles/jAFzYrV

andyyoung01:
@seozed Thanks!
I also found the document about "Package download mirrors" at [here](https://github.com/yyuu/pyenv/tree/master/plugins/python-build). It describe that "Package mirror URLs are constructed by joining this variable with the SHA2 checksum of the package file."

yyuu:
I hope everything worked now. I guess that the problem was incorrect file naming on mirror site.

ilovemysillybanana:
Can someone provide a walk through on how to do this? I'm trying to set this up now but that website linked by @seozed does not load for me.

leafonsword:
@seozed this method still not work : http://www.tuicool.com/articles/jAFzYrV

[...]

PR #1673, opened 2020-08-03
Add support for PYTHON_BUILD_MIRROR_URL when checksums do not exist

jamescurtin:

### Prerequisite
* [N/A] Please consider implementing the feature as a hook script or plugin as a first step.
* [N/A] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  - Closes https://github.com/pyenv/pyenv/issues/1473
  - Closes https://github.com/pyenv/pyenv/issues/1377
  - https://github.com/pyenv/pyenv/issues/1575
  - https://github.com/pyenv/pyenv/issues/803

### Description

Fixes an issue where `PYTHON_BUILD_MIRROR_URL` is not used if the mirror does not have the same checksum that is expected. In some cases, the mirror is not expected to have a matching checksum (as in the case of a corporate mirror behind a firewall), which causes errors.

By setting `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM`, the `PYTHON_BUILD_MIRROR_URL` will be used but will not attempt to match a checksum.

### Tests
  - package is fetched from mirror when checksum is invalid if SKIP_CHECKSUM set

### DIFF
--- a/plugins/python-build/README.md
+++ b/plugins/python-build/README.md
@@ -109,6 +109,8 @@ You can set certain environment variables to control the build process.
   downloaded package files.
 * `PYTHON_BUILD_MIRROR_URL` overrides the default mirror URL root to one of your
   choosing.
+* `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM`, if set, does not append the SHA2
+  checksum of the file to the mirror URL.
 * `PYTHON_BUILD_SKIP_MIRROR`, if set, forces python-build to download packages from
   their original source URLs instead of using a mirror.
 * `PYTHON_BUILD_ROOT` overrides the default location from where build definitions
@@ -181,6 +183,10 @@ You can point python-build to another mirror by specifying the
 `PYTHON_BUILD_MIRROR_URL` environment variable--useful if you'd like to run your
 own local mirror, for example. Package mirror URLs are constructed by joining
 this variable with the SHA2 checksum of the package file.
+
+If the mirror being used does not have the same checksum (*e.g.* with a
+pull-through cache like Artifactory), you can set the
+`PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM` environment variable.
 
 If you don't have an SHA2 program installed, python-build will skip the download
 mirror and use official URLs instead. You can force python-build to bypass the
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -480,7 +480,12 @@ reuse_existing_tarball() {
 }
 
 download_tarball() {
-  local package_url="$1"
+  local official_source="www.python.org/ftp/python"
+  if [ -n "$PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM" ]; then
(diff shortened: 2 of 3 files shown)


jamescurtin:
👋  Checking in on this PR: would someone please be able to review? Thank you!

jamescurtin:
Hi: wanted to check in again--would a maintainer be able to review this PR please? Thank you! (CC @joshfriend sorry for the ping 😄 )

joshfriend:
why would the download from a mirror have a different checksum?

jamescurtin:
The downloaded file will have the same checksum--the issue with the current implementation is the assumption that the checksum is always included in the filepath of the mirrored distribution. For example:

`3.8.5` can be downloaded from `https://www.python.org/ftp/python/3.8.5/Python-3.8.5.tar.xz`.

My `PYTHON_BUILD_MIRROR_URL` is `https://mirror.example.com/python/`, and the source distribution is available at `https://mirror.example.com/python/3.8.5/Python-3.8.5.tar.xz`.

Based on [this logic in `python-build`](https://github.com/pyenv/pyenv/blob/f44cb70d938446b396a676dd76f1b1b28031a9e6/plugins/python-build/bin/python-build#L393), there is an assumption that the path to the mirrored distribution is https://mirror.example.com/python/e3003.../Python-3.8.5.tar.xz, where `e3003...` is the checksum of the file. This may be the case for some mirrors, but is not the case when configuring a remote repository mirror using Artifactory (as is my situation). 

Therefore, the behavior introduced with `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM` is to skip templating the checksum into the URL of the download. The [`verify_checksum` function is still called](https://github.com/pyenv/pyenv/blob/f44cb70d938446b396a676dd76f1b1b28031a9e6/plugins/python-build/bin/python-build#L492) when downloading the tarball from the mirror, so you remain protected against corrupted or poisoned distributions.

joshfriend:
Ah I understand now :) thank you ❤️

uzxmx:
Just want to add a comment:

Because of this [code snippet](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/bin/python-build#L421-L422), it actually tries downloading the tarball twice. The first time it tries downloading with the checksum appended, but fails. The second time without checksum, hence succeeds.

So I think the name `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM` may not be a good name, because it doesn't skip, but actually substitutes the original url with `PYTHON_BUILD_MIRROR_URL`.

```
  if [ -n "$PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM" ]; then
      local package_url="$(echo "$1" | sed -e "s|.*//${URL_BASE:-$official_source}|$PYTHON_BUILD_MIRROR_URL|g")"
  else
      local package_url="$1"
  fi
```

In order to use the name `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM`, maybe [here](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/bin/python-build#L387-L396) is a good place to do some edits.

zengzuo613:
The Alibaba Cloud image path does not comply with the official agreement

Issue #1681, opened 2020-08-19
PyPy release page moved from bitbucket to python.org


ashwinvis:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Arch Linux
- [x] OS architecture (e.g. amd64): amd64
- [x] pyenv version: 1.2.20-1-g70635dad
- [x] Python version: 3.8.1
- [x] C Compiler information (e.g. gcc 7.3):  gcc 10.1.0
- [x] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

```sh
❯ PYENV_DEBUG=1 pyenv install -v pypy3.6-7.3.1                                                                                                                                              
+ [pyenv:22] enable -f /home/avmo/.pyenv/bin/../libexec/pyenv-realpath.dylib realpath
+ [pyenv:29] '[' -z '' ']'
++ [pyenv:31] type -p greadlink readlink
++ [pyenv:31] head -1
+ [pyenv:31] READLINK=/usr/bin/readlink
+ [pyenv:32] '[' -n /usr/bin/readlink ']'
+ [pyenv:57] '[' -z '' ']'
+ [pyenv:58] PYENV_ROOT=/home/avmo/.pyenv
+ [pyenv:62] export PYENV_ROOT
[...]

PR #1682, opened 2020-08-19
Fix PyPy download links

ashwinvis:

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1681

### Description

PyPy has migrated away from Bitbucket hosting to Heptapod. Sources and build
archives are now hosted at downloads.python.org.

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/pypy-2.0-src
+++ b/plugins/python-build/share/python-build/pypy-2.0-src
@@ -1,4 +1,4 @@
 #require_gcc
 prefer_openssl11
 install_package "openssl-1.1.0j" "https://www.openssl.org/source/old/1.1.0/openssl-1.1.0j.tar.gz#31bec6c203ce1a8e93d5994f4ed304c63ccf07676118b6634edded12ad1b3246" mac_openssl --if has_broken_mac_openssl
-  install_package "pypy-2.0-src" "https://downloads.python.org/pypy/pypy-2.0-src.tar.bz2#d92dfd418beac915d3efc28f8a2090f3c13a89ec653419deff3d7112c5c111f3" "pypy_builder" verify_py27 ensurepip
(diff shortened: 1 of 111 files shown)


plugins/python-build/share/python-build/pypy-2.0-src
```diff
@@ -1,4 +1,4 @@
 #require_gcc
 prefer_openssl11
 install_package "openssl-1.1.0j" "https://www.openssl.org/source/old/1.1.0/openssl-1.1.0j.tar.gz#31bec6c203ce1a8e93d5994f4ed304c63ccf07676118b6634edded12ad1b3246" mac_openssl --if has_broken_mac_openssl
-install_package "pypy-2.0-src" "https://bitbucket.org/pypy/pypy/downloads/pypy-2.0-src.tar.bz2#d92dfd418beac915d3efc28f8a2090f3c13a89ec653419deff3d7112c5c111f3" "pypy_builder" verify_py27 ensurepip
+  install_package "pypy-2.0-src" "https://downloads.python.org/pypy/pypy-2.0-src.tar.bz2#d92dfd418beac915d3efc28f8a2090f3c13a89ec653419deff3d7112c5c111f3" "pypy_builder" verify_py27 ensurepip
```
squeaky-pl, line 4: I was wondering if those whitespace changes on the beginning of the line were intentional.

ashwinvis: It happened because awk discards them. I agree, they have to be fixed.

But apart from that, I would be grateful if someone can verify the installation works and the checksums (?) are correct.

ashwinvis:
I think this is good to go from my side.

joshfriend:
Thanks!

I obviously did not check every single version to see if it installed, but I did download a few random tarballs from the new URLs to verify that they were valid URLs and checked over the diff to make sure no checksums were altered.

matclayton:
Any chance of pushing out a new release (ideally to brew) so we can use these. Alternatively is there a known workaround in the meantime?

ashwinvis:
+1 for a new release.

> Alternatively is there a known workaround in the meantime?

The fix has been merged on the master branch. Clone pyenv and use it?

Issue #1683, opened 2020-08-26
Anaconda coreutils package creates buggy shims in pyenv


miketrumpis:
I like using pyenv to manage a mixture of from-scratch and Miniconda environments. Today I set up a miniconda recipe that included "coreutils" (it packages [these commands][1]). The problem here is that at some point pyenv creates shims for all these commands, which leads to the spawning of endless new bash processes. I don't know enough about the pyenv scripts to narrow down *which* of these shims is a problem. I suspect it's the ``[`` test operator (why?!?). I don't know if this is really a pyenv issue, but I guess it counts as a bug if the system can allow a shim that breaks itself.

My system

```bash
$ uname -a
Darwin denali.local 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64
```

Pyenv version at commit 58c776a

Here are the steps to reproduce. 

1. basic pyenv setup for a bash shell

```bash
$ git clone https://github.com/pyenv/pyenv.git ~/.pyenv
$ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
$ echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
$ echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
```

Since bash breaks, I would recommend having .bash_profile open for editing in a separate shell or other window before proceeding!! That way you can go back and comment out pyenv-init parts and fix the damage.

2. (exec new shell) and ``pyenv install miniconda3-latest``
3. ``pyenv shell miniconda3-latest`` and ``conda create -n conda-env coreutils``

At this point any new shell will populate ``$(PYENV_ROOT)/shims`` with a bunch of junk shims from GNU coreutils. These shims, presumably, interrupt the ``pyenv init`` process. 

For anyone reproducing, it is helpful here to ``CTRL-C`` to end the runaway recursion. You'll need to delete the shims or just clear out the ``$(PYENV_ROOT)`` to recover.

  [1]: https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands

joshfriend:
[...]

PR #1685, opened 2020-08-27
exclude bin files from coreutils

miketrumpis:
I added the bin/* contents of the anaconda metadata (coreutils-8.31-h0b31af3_0.json) to default.list

This is preventing bash-crashing shims raised in #1683.


yyuu (approved):
lgtm

PR #1691, opened 2020-09-14
python-build: Only use explicit path as definition if it's not a plain name

native-api:
This fixes erroneous behavior if there happens to be an entry in the current directory
with the same name as the defnition.
To explicitly use a file in the current directory as a definition, specify it as "./file_name".

I ran into this issue when installing and testing multiple versions, saving test output into a file with the same name as the version for simplicity.

---

### Prerequisite
  N/A: bugfix
  N/A: affects python-build
  N/A

### Description

This fixes erroneous behavior if there happens to be an entry in the current directory
with the same name as the defnition.
To explicitly use a file in the current directory as a definition, specify it as "./file_name".

### Tests
 There don't seem to be any tests for python-build in tests/. Should there?

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1871,7 +1871,7 @@ done
 DEFINITION_PATH="${ARGUMENTS[0]}"
 if [ -z "$DEFINITION_PATH" ]; then
   usage 1 >&2
-elif [ ! -f "$DEFINITION_PATH" ]; then
+elif [[ ! "$DEFINITION_PATH" =~ \/ || ! -f "$DEFINITION_PATH" ]]; then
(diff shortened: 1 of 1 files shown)


native-api:
@pyenv/pyenv-core-committers Any feedback/opinion on this change? Since it's mine and potentially contentious, I'm hesitant to merge it with none at all.

native-api:
After much consideration, I decided that the PR violates Python Zen, koan 12 ("In the face of ambiguity, refuse the temptation to guess."). The existing code makes a guess (prefers a file to a stock script) and my change guesses the other way round.

If this is to be frixed, we'd rather need to introduce a switch that will explicitly tell Python-Build to use a file path instead of a stock script.

PR #1692, opened 2020-09-14
Fix download links for everything (incl. PyPy and Stackless)

native-api:
Seeing that `pypy-dev` no longer builds at https://travis-ci.com/github/travis-ci/cpython-builder/jobs/384173201 because PyPy had moved away from BitBucket,

I've checked _all_ the links and checksums in `plugins/python-build/share/python-build` with the scripts in https://github.com/pyenv/pyenv/pull/1692#issuecomment-711413160.

This PR fixes all the found invalid links to valid ones and deletes entries for releases that are no longer available for download.

---

### Prerequisite
  - N/A

### Description

See above.

### Tests
  I can add a test for installing (or at least downloading/HEAD checking URLs for) all the archives if you wish.


joshfriend:
#1682 already fixed pypy download links, can you update this with only changes to stackless?

native-api:
@joshfriend I was checking a codebase with #1682 applied and started working on this PR after I couldn't install a few versions. So I can say with confidence that it didn't fix everything.

plugins/python-build/share/python-build/pypy-2.3.1-src
```diff
@@ -1,4 +1,4 @@
 #require_gcc
 prefer_openssl11
 install_package "openssl-1.1.0j" "https://www.openssl.org/source/old/1.1.0/openssl-1.1.0j.tar.gz#31bec6c203ce1a8e93d5994f4ed304c63ccf07676118b6634edded12ad1b3246" mac_openssl --if has_broken_mac_openssl
-install_package "pypy-pypy-32f35069a16d" "https://bitbucket.org/pypy/pypy/get/release-2.3.1.tar.bz2#3fd10d97c0177c33ed358a78eb26f5bf1f91b266af853564b1a9d8c310a1e439" "pypy_builder" verify_py27 ensurepip
+install_package "pypy-pypy-32f35069a16d" "https://downloads.python.org/pypy/pypy-2.3.1-src.tar.bz2#6b10efab2be5502b475a25181870616e9ff83296b225aa2c7a7ec617c0ff78ef" "pypy_builder" verify_py27 ensurepip
```
joshfriend, line 4: seems weird that the checksum would change, why isnt the binary the same?

native-api: No idea. Maybe it was repackaged with different time stamps or something. Ask the project members if you want to know.

All that matters for Pyenv is that the new site is still an official one -- so the downloads are presumed trustworthy -- and since Pyenv checks the signatures, we have to update them if they change.

native-api:
That said, I didn't check signatures for _all_ the download links, only for those that I fixed (because the logic for this in `python-build` is rather complicated and there's no easy way to reuse it). Shall I do that in this PR?

native-api:
Checked checksums in addition to URLs with the following code:

```
#!/bin/bash -e
set -o pipefail
declare -A URLS
grep * -Pe '(?<=")http[^"]+(?=")' --only-matching | \
while IFS=':' read -r FILE URLLINE
do
  HASH=${URLLINE##*#}; if [[ $HASH == "$URLLINE" ]]; then HASH=""; fi
  if [[ -n $HASH ]]; then
    if [[ ${#HASH} -eq 32 ]]; then
        SUM=md5sum
    else
        SUM=sha256sum
    fi
    echo "$HASH *t.bin" >t.sum
  fi
  URL=${URLLINE%#*}
  if [[ -n ${URLS[$URL]} ]]; then continue; fi;
  echo "$FILE:$URL"
  if 
    (
        curl -sSfL -o t.bin "$URL" && \
        if [[ -n $HASH ]]; then
            "$SUM" --check t.sum
        fi
    ) 2>&1 \
    | tee out.txt
  then
    URLS[$URL]=1
  else
    echo "$FILE:$URL $(cat out.txt)" >>failed.txt
  fi
done
```

Checked generated URLs in the latest PyPy scripts with the following code:

```
#!/bin/bash -e
set -o pipefail
declare -A URLS

function get_version () {
  eval "$( function echo () { true; }
    function exit () { true; }
    source "$1";
    command echo VERSION=$VERSION;
    command echo PYVER=$PYVER;
  )"
}


grep pypy*-7* -Pe '(?<=")http[^"]+(?=")' --only-matching | \
while IFS=':' read -r FILE URLLINE
do
  get_version "$FILE"
  HASH=${URLLINE##*#}; if [[ $HASH == "$URLLINE" ]]; then HASH=""; fi
  if [[ -n $HASH ]]; then
    if [[ ${#HASH} -eq 32 ]]; then
        SUM=md5sum
    else
        SUM=sha256sum
    fi
    echo "$HASH *t.bin" >t.sum
  fi
  URL=${URLLINE%#*}
  URL="$(eval "echo \"$URL\"")"
  if [[ -n ${URLS[$URL]} ]]; then continue; fi;
  echo "$FILE:$URL"
  if 
    (
        curl -sSfL -o t.bin "$URL" && \
        if [[ -n $HASH ]]; then
            "$SUM" --check t.sum
        fi
    ) 2>&1 \
    | tee out.txt
  then
    URLS[$URL]=1
  else
    echo "$FILE:$URL $(cat out.txt)" >>failed.txt
  fi
done
```

PR #1707, opened 2020-10-05
Add CPython 3.9.0

linusg:
Replaces the current latest 3.9.0 (rc2) with 3.9.0 stable, released today.


joshfriend:
#1706 was submitted first so i merged that one, than you though!

PR #1708, opened 2020-10-06
Travis CI: Add Xcode 12 on macOS 10.15.5

cclauss:

### Prerequisite

### Description
Add the current macOS and current Xcode to the testing.
* https://docs.travis-ci.com/user/reference/osx/#macos-version
### Tests
Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ dist:
 
 # https://docs.travis-ci.com/user/reference/osx/#macos-version
 osx_image:
-- xcode9.4.1
-- xcode10.3
-- xcode11.3
+- xcode9.4.1  # macOS 10.13
+- xcode10.3   # macOS 10.14.4
+- xcode11.3   # macOS 10.14.6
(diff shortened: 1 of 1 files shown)


anton-petrov:
I suggest to more complex changes. Upgrade from Xenial to Focal, for example. As for xcode12u, the best solution would be `xcode12` (Xcode 12.0.1).  And another important point: the more we add an environment for testing, the longer the build time will be.

cclauss:
Travis jobs should run in parallel so the impact of adding a new env might not be a linear increase in runtime.  If turnaround time is important then perhaps we should move from Travis CI to GitHub Actions.

cclauss:
Unfortunately, Travis CI still does not provide access to macOS Big Sur (v11)
ProductName:	Mac OS X
ProductVersion:	10.15.7

anton-petrov:
Currently runs only 2 parallel vm's for pyenv, as know...

PR #1711, opened 2020-10-07
Add LDFLAGS for zlib on macOS >= 1100

lambda-9:
Add LDFLAGS for Xcode SDK zlib on macOS >= 1100 to resolve build issues on Big Sur.

### Prerequisite
  n/a
  n/a - this code does not appear in rbenv
  - https://github.com/pyenv/pyenv/issues/1643
Note: The issue is for general macOS 11 Big Sur beta build issues. Users must also have downloaded the Xcode 12.2 beta from Apple and have the macOS 11 SDK installed in addition to the Big Sur beta itself.

### Description
This is a 3 line change to python-build that affects macOS 11.0 Big Sur (currently in beta). Big Sur no longer packages dynamic libraries in /usr/lib as detailed in https://github.com/pyenv/pyenv/issues/1643#issuecomment-704223840 . This requires python-build to link against $(xcrun --show-sdk-path)/usr/lib on macOS >= 11.0 when using the macOS-shipped zlib. This allows python-build to correctly build python again on macOS 11 (Big Sur).

### Tests
n/a

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1493,6 +1493,9 @@ use_xcode_sdk_zlib() {
   if [ -d "$xc_sdk_path" ]; then
     echo "python-build: use zlib from xcode sdk"
     export CFLAGS="-I${xc_sdk_path}/usr/include ${CFLAGS}"
+    if is_mac -ge 1100; then
+      export LDFLAGS="-L${xc_sdk_path}/usr/lib ${LDFLAGS}"
+    fi
   fi
 }
 


archoversight:
Now that Big Sur is available for realz, it would be great to have this in pyenv :-)

armenzg:
Hello all,
Anyone know when the next release of `pyenv` will happen?
Thank you so much for your hard work!

tillhainbach:
@joshfriend This PR may actually cause the build to fail compiling readline from home-brew. At least on both of my macs running on macOS Big Sur 11.0.1. It will print
```sh
WARNING: The Python readline extension was not compiled. Missing the GNU readline lib?
```
The error could not have been caught by travis-ci because the pipeline does not include macOS 11 (yet).
I commented-out the added lines and exported the LDFLAG manually with: 
```
export LDFLAGS='-L /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib/'
```
This time, pyenv installed/compiled python with no errors or warnings.

So, it turns out that `${xc_sdk_path}` was using the MacOSX11.0.sdk from Xcode.app (Version 12.2 (12B45b)) and the sdk in `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib/` was still 10.15.
Can someone confirm similar issues?

tillhainbach:
So I did some more digging and apparently the order in LDFLAGS is somehow relevant (?)

```bash
    if is_mac -ge 1100; then
      export LDFLAGS="-L${xc_sdk_path}/usr/lib ${LDFLAGS}"
    fi
```

--> Warning ...

```bash
    if is_mac -ge 1100; then
      export LDFLAGS="${LDFLAGS} -L${xc_sdk_path}/usr/lib"
    fi
```
--> no warning

tillhainbach:
Edit: I updated the travis setup and the python-build-script to force travis to build against macOS SDK 11.0 (osx_image xcode12.2) and was able to reproduce the warning that readline extension was not compiled. However, this is not treated as a build error and hence the checks passed. But you can see the warning in the travis-logs. See example here: https://github.com/tillhainbach/pyenv/pull/8 and logs here https://travis-ci.com/github/tillhainbach/pyenv/jobs/457193096#L9265

So, it should be considered merging #1754

hjmallon:
@tillhainbach did you consider any other methods than adding a linker flag? It seems homebrew are using -isysroot for this stuff instead, I'n not sure whether that should make much difference.

PR #1713, opened 2020-10-10
CPython 3.10.0a1

nedbat:
This adds CPython 3.10.0a1.


nedbat:
a2 is out now, and I guess you don't want recipes for alphas?

joshfriend:
Thanks, Ned 😄

Recipes for alphas/betas/RCs are welcome (we try to only keep the most recent one for each version). I seem to be the only maintainer that is active on the project recently and my recently expanded family takes more of my time and energy than usual.

Sorry your contribution was basically ignored, that's not intentional.

nedbat:
@joshfriend "recently expanded family" is a good reason! Congratulations (I'm assuming that phrase means a baby...)! I didn't harbor any ill-will in the first place.

PR #1736, opened 2020-11-18
Add GraalPython 20.3

timfel:
This adds the recently released GraalPython 20.3 to the available install scripts.


timfel:
@joshfriend friendly ping :)

PR #1747, opened 2020-12-05
rehash: try to sleep in 0.1 sec steps when acquiring lock

scop:
### Prerequisite
  - N/A

### Description

POSIX sleep(1) doesn't provide subsecond precision, but many others do. I looked around (Ubuntu, FreeBSD, Solaris, BusyBox) and failed to find one that actually doesn't.

Doesn't apply to rbenv, touches pyenv specific code.

Not sure if 0.1 is the amount we want by default, open to changing it.

### Tests

N/A

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -45,8 +45,8 @@ for (( i=1; i<="${PYENV_REHASH_TIMEOUT:-60}"; i++ )); do
     acquired=1
     break
   else
-    # POSIX sleep(1) doesn't provides time precision of subsecond
-    sleep 1
+    # POSIX sleep(1) doesn't provide subsecond precision, but many others do
+    sleep 0.1 2>/dev/null || sleep 1
   fi
 done
 
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -40,7 +40,8 @@ if [ ! -w "$SHIM_PATH" ]; then
 fi
 
 unset acquired
-for (( i=1; i<="${PYENV_REHASH_TIMEOUT:-60}"; i++ )); do
+start=$SECONDS
+while (( SECONDS <= start + ${PYENV_REHASH_TIMEOUT:-60} )); do
   if acquire_lock 2>/dev/null; then
     acquired=1
     break


libexec/pyenv-rehash
```diff
@@ -45,8 +45,8 @@ for (( i=1; i<="${PYENV_REHASH_TIMEOUT:-60}"; i++ )); do
     acquired=1
     break
   else
-    # POSIX sleep(1) doesn't provides time precision of subsecond
-    sleep 1
+    # POSIX sleep(1) doesn't provide subsecond precision, but many others do
+    sleep 0.1 2>/dev/null || sleep 1
```
chrahunt, line 49: This loop expects the sleep to take 1 second. As-written, it will give up in 1/10th of the expected time if `sleep` supports subsecond precision.

chrahunt (commented):
I'm not familiar with where this is used. What is the intended user-experience outcome of reducing this sleep?

scop:
> I'm not familiar with where this is used. What is the intended user-experience outcome of reducing this sleep?

It's used in `pyenv init` by default which is intended to be called on shell startup. User experience is reduced shell startup time when the lock cannot be acquired immediately.

scop:
(I'm aware of `--no-rehash` and I'm using that due to other rehash slowness personally.)

libexec/pyenv-rehash
scop: Valid, point, fixed in 173fa22e0746c4db3de69c5201b21f416987a052.

anton-petrov:
I just tested it in various environments, including macOS - everything works fine. As for the `sleep` in macOS -- this function is implemented using _nanosleep()_ that is not accurate on macOS, look at [discussion](https://travis-ci.community/t/sleep-functions-are-not-accurate-on-macos/6122). In any case, the changes you made affect performance, so, I see no reasons to not include this improvements.

scop:
Hm, I read your comment as positive and "no reasons to not include this improvements", but nevertheless the PR was closed without merging and without different but corresponding fixes in master. Was that an oversight or did I miss something?

anton-petrov:
Oh, I'm sorry, that was a mistake :(((

anton-petrov:
All done. Merged your code into the master.

scop:
Thanks!

PR #1749, opened 2020-12-05
Use associative arrays for performance on bash >= 4

scop:

### Prerequisite
  - N/A

### Description

This makes use of bash associative arrays for speedups, see individual commit messages for details. These are quite noticeable at shell startup.

Will submit the same PR's to rbenv.

~Note: associative arrays are available starting from bash 4.0. I didn't find a reference which bash versions should pyenv work with, but bash 4.0 is already pretty old. If that's a no go, the implementation could be reworked so that it defines different function implementations for different bash versions.~ Reworked.

### Tests

N/A

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -123,34 +123,62 @@ make_shims() {
   done
 }
 
-registered_shims=" "
+if ((${BASH_VERSINFO[0]} > 3)); then
 
-# Registers the name of a shim to be generated.
-register_shim() {
-  registered_shims="${registered_shims}${1} "
-}
+  declare -A registered_shims
 
-# Install all the shims registered via `make_shims` or `register_shim` directly.
-install_registered_shims() {
-  local shim file
-  for shim in $registered_shims; do
-    file="${SHIM_PATH}/${shim}"
-    [ -e "$file" ] || cp "$PROTOTYPE_SHIM_PATH" "$file"
-  done
-}
+  # Registers the name of a shim to be generated.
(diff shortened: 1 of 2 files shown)


yyuu:
As far as I know, macOS is still shipping its BSD-based system along with bash 3.x even in 2021 (I'd speculate it is because of license concern, but, I'm not confident). Dependency to bash 4+ features might not be a very good idea until this situation to be changed 🤔 

```
% sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H114
% /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)
Copyright (C) 2007 Free Software Foundation, Inc.
```

scop:
I don't think that will ever change, and it would be a shame to keep the rest of the world stuck in the past just because of that. On the other hand, newer bash versions are available for macOS, they're just not there by default.

scop:
Oh, but then again, I didn't even remember that I already _did_ rework this so that it still works with bash < 4 as well, check out the patch contents.

anton-petrov:
Since in macOS 10.15 Catalina the default shell is zsh, the situation is changing to the good. But we can't just lose compatibility with other systems that still use bash 3. So, can we drop bash 3 support? In any way, this is an important case to be discussed. Maybe we need compatibility layer, or something similar. For example, in [pyenv-init ](https://github.com/pyenv/pyenv/blob/master/libexec/pyenv-init) do some checks for the shell version.

Issue #1136, opened 2018-04-13
bz2 and readline extension warnings during python 3.6.5 installation 


sunbearc22:
### Issue: 
I receive the two warnings. What should I do to avoid these warnings?

```
$ pyenv install 3.6.5
Downloading Python-3.6.5.tar.xz...
-> https://www.python.org/ftp/python/3.6.5/Python-3.6.5.tar.xz
Installing Python-3.6.5...
WARNING: The Python bz2 extension was not compiled. Missing the bzip2 lib?
WARNING: The Python readline extension was not compiled. Missing the GNU readline lib?
Installed Python-3.6.5 to /home/sunbear/.pyenv/versions/3.6.5
```

### Description
Platform information: Ubuntu Linux 16.04.3
OS architecture: amd64
pyenv version: latest at this point of time (just installed pyenv)
Python version: 3.6.5
C Compiler information:  gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609

outof-step:
You're missing build dependencies, you should install those. This is not a bug. 
Instructions are in the wiki under Ubuntu/Debian/Mint https://github.com/pyenv/pyenv/wiki#suggested-build-environment

PR #1750, opened 2020-12-07
Put prerequisite for installation before install

Almenon:
This prevents people from jumping the gun and trying to install a python version right after brew install.

### Prerequisite
* [N/A] Please consider implementing the feature as a hook script or plugin as a first step.
* [N/A] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  - https://github.com/pyenv/pyenv/issues/1136

### Description
See top

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,10 @@ We'd recommend to install pyenv-virtualenv as well if you have some plan to play
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


joshfriend:
good call

PR #1763, opened 2020-12-23
Add miniconda3 4.8.2, 4.8.3, & 4.9.2

pc-fmarin:

### Prerequisite

### Description

Update miniconda 4.8.2, 4.8.3, & 4.9.2 w/ py7, py8, and py9 versions

### Tests


anton-petrov (approved):
All versions is up to date.

yyuu (commented):
At the first glance, I was surprised with the naming of those new definition files (e.g. use of underscore), however, it seems like it's just following the naming convention of upstream as they are. I think, just using the upstream naming would be good and intuitive for most uses. Any opinions/insights? 🤔

anton-petrov:
> At the first glance, I was surprised with the naming of those new definition files (e.g. use of underscore), however, it seems like it's just following the naming convention of upstream as they are. I think, just using the upstream naming would be good and intuitive for most uses. Any opinions/insights? 🤔

I fully agree that file naming should be standard, a general approach should be followed, we shouldn't use underscores in filenames, IMHO

anton-petrov (changes requested):
It might be worth changing the naming of the files a bit...? Like miniconda3-**3.7**-4.8.3, where 3.7 is the python version.

anton-petrov (commented):
It might be worth changing the naming of the files a bit, like miniconda3-**3.7**-4.8.3? Where 3.7 is the python version, 4.8.3 - version of miniconda.

anton-petrov (approved):

Issue #1649, opened 2020-07-03
Shims directory appears multiple times in PATH


ggicci:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Ubuntu Linux 18.04
- [x] OS architecture (e.g. amd64): amd64
- [x] pyenv version: pyenv 1.2.19-4-g366c9801
- [x] Python version: 3.8.0
- [ ] C Compiler information (e.g. gcc 7.3): 
- [ ] Please attach verbose build log as gist
 
Directory `${PYENV_ROOT}/shims` appears multiple times in `PATH`. We can avoid this by checking the existence of the shims path in `PATH` before appending.

native-api:
WFM. Shims is only on the PATH once if I install Pyenv using the official instructions. Perhaps you're doing some custom stuff.

Integralist:
I have the same problem, which is caused by reloading my `.bashrc` file (which I do multiple times during the day, whether I'm making changes or I'm loading up a new tmux instance -- which subsequently loads .bashrc).

Because I have the lines...

- `eval(pyenv init -)`
- `eval(pyenv virtualenv-init -)`

...the code internally executed by those is mutating the $PATH by just prefixing to the path without checking it first.

So I've solved this by _inlining_ all the code from those two eval's and then replacing the path modification lines with...

```bash
pathmunge /Users/integralist/.pyenv/shims
[...]

PR #1769, opened 2021-01-05
Prevent duplicated shims in PATH variable

felipecrs:
This happens when someone adds the shims folder to the PATH variable manually.

I'm going to submit this to the upstream, so we don't need to merge this now, but we can discuss.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -84,13 +84,16 @@ fi
 
 mkdir -p "${PYENV_ROOT}/"{shims,versions}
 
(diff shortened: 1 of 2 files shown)


native-api:
This is being discussed in https://github.com/pyenv/pyenv/issues/1649 where alternative approaches are also considered.

cclauss:
#1838 adds the shims directory to $PATH in our GitHub Actions

anton-petrov:
Closed due to #1898

PR #1774, opened 2021-01-09
Added anaconda3-2020.11

anton-petrov:
Added support for Anaconda 2020.11

### Prerequisite

### Description

### Tests


plugins/python-build/share/python-build/anaconda3-2020.11
```diff
@@ -0,0 +1,19 @@
+case "$(anaconda_architecture 2>/dev/null || true)" in
+"Linux-ppc64le" )
+  install_script "Anaconda3-2020.11-Linux-ppc64le" "https://repo.continuum.io/archive/Anaconda3-2020.11-Linux-ppc64le.sh#870535ada0a8ae75eeda8cd2bf7dde853ac9f4949b20e1b5641f1843a655f3b8" "anaconda" verify_py38
```
chrahunt, line 3: Nice catch switching these to `verify_py38`. According to the release notes 3.8 was used by default in 2020.07, I'll file a separate PR to update that.

chrahunt (approved):
Thank you!

Checked:

- diff with 2020.07 script: URLs, digests, and `verify_*` has been updated
- digests match artifacts
- URLs match product and are coming from the same domain as the previous script

PR #1766, opened 2020-12-30
Anaconda3-2020.11, Miniforge3-4.9.2-4, and Apple M1 support in anaconda_architecture()

anton-petrov:
1. I added support for Anaconda 2020.11. (Linux x86_64, Linux ppc64, macOS x86_64)
2. Added Miniforge3-4.9.2-4 support (Linux x86_64, Linux ppc64, macOS x86_64, macOS arch64), so now we can install and use this python distribution too! This repository holds a minimal installer for conda specific to conda-forge. It is comparable to Miniconda, but with conda-forge set as the default channel and an emphasis on supporting various CPU architectures.
3. I made some changes for supporting new architecture Apple M1 on MacOS in python-build script, specially for Miniforge and future Anaconda installation.

All works fine, tested and using on my Macbook Air M1.

Make sure you have checked all steps below.

### Prerequisite
* [x] Please consider implementing the feature as a hook script or plugin as a first step.
  * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins](https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a plugin if possible.
* [x] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv will be imported to pyenv too, eventually.
  * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with rbenv.
* [x] My PR addresses the following pyenv issue (if any)
  - https://github.com/pyenv/pyenv/issues/XXXX

### Description
- [x] Here are some details about my PR

### Tests
- [x] My PR adds the following unit tests (if any)

PR #1775, opened 2021-01-09
Added arm64 architecture support in python-build for macOS 

anton-petrov:

### Prerequisite

### Description

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -995,7 +995,13 @@ build_package_activepython() {
 
 anaconda_architecture() {
   case "$(uname -s)" in
-  "Darwin" ) echo "MacOSX-x86_64" ;;
+  "Darwin" )
+    case "$(uname -m)" in
+    "arm64" ) echo "MacOSX-arm64" ;;
+    "i386" | "x86_64" ) echo "MacOSX-x86_64" ;;
+    * ) return 1 ;;
+    esac
+    ;;
   "Linux" )
     case "$(uname -m)" in
     "armv7l" ) echo "Linux-armv7l" ;;
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -998,8 +998,7 @@ anaconda_architecture() {
   "Darwin" )
     case "$(uname -m)" in
     "arm64" ) echo "MacOSX-arm64" ;;
-    "i386" | "x86_64" ) echo "MacOSX-x86_64" ;;
-    * ) return 1 ;;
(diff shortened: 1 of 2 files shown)


plugins/python-build/bin/python-build
```diff
@@ -995,7 +995,13 @@ build_package_activepython() {
 
 anaconda_architecture() {
   case "$(uname -s)" in
-  "Darwin" ) echo "MacOSX-x86_64" ;;
+  "Darwin" )
+    case "$(uname -m)" in
+    "arm64" ) echo "MacOSX-arm64" ;;
```
chrahunt, line 1000: Is it safe to assume that macOS users on ARM always want the ARM version? If some dependencies do not support ARM, they may prefer x86_64 (executed via Rosetta).

plugins/python-build/share/python-build/anaconda3-2020.11
```diff
@@ -0,0 +1,19 @@
+case "$(anaconda_architecture 2>/dev/null || true)" in
```
chrahunt, line 1: This looks unrelated to the arm64 change. If we pull this file into it's own PR then it is easier to review. Apologies, I did not communicating this well on #1766.

plugins/python-build/bin/python-build
anton-petrov: Hi @chrahunt, the approach I suggested is safe and here's why: macOS users on ARM can run programs in x86_64  mode on Rosetta 2 translator quite simple and transparent with the command `arch x86_64`. For example, `arch x86_64 pyenv install 3.9.1` build and install python for x86_64 architecture. In this way, the "arch" command runs commands under Rosetta 2. This approach is checked and tested on my MacBook Air M1 - everything works well! In fact, this is the only (correct) approach, that supported by Apple, and, may be, it would be incorrect to add any new parameters for pyenv, such as `pyenv install 3.9.1 x86_64`ю

chrahunt: Thank you for the explanation! I agree, in that light this seems like the right approach.

plugins/python-build/bin/python-build
```diff
@@ -995,7 +995,13 @@ build_package_activepython() {
...
+    "i386" | "x86_64" ) echo "MacOSX-x86_64" ;;
+    * ) return 1 ;;
```
chrahunt, line 1002:
```suggestion
    * ) echo "MacOSX-x86_64" ;;
```

to preserve the original behavior - it may be unlikely that someone is relying on it, but it's less work for now to not ask the question.

PR #1785, opened 2021-01-13
more general installation readme instructions

John15321:

### Description

I added command for pyenv installation for `Zsh`, `Ubuntu` etc. So its easier to do when installing and you can just copy and paste. I did it after having to install pyenv multiple times on different machines. Please check if the commands are all right 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ easy to fork and contribute any changes back upstream.
 
    - For **Fish shell**:
      ~~~ fish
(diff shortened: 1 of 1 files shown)


anton-petrov:
Have you tested this in all environments?

John15321:
> Have you tested this in all environments?
I did it in all except the `Fish` one. As I do not have access to FIsh nor experience with it

PR #1790, opened 2021-01-25
add manpage

karthek-git:
add manpage for pyenv

### DIFF
--- a/pyenv.1
+++ b/pyenv.1
@@ -1,0 +1,549 @@
+.\" (C) Copyright 2020 karthek <mail@karthek.com>,
+.\"
+.TH PYENV 1 "12 Dec 2020" "PYENV"
+.SH NAME
+pyenv \- Simple Python version management
+.SH SYNOPSIS
+.B pyenv
+\fI\,<command> \/\fR[\fI\,<args>\/\fR]
+.SH DESCRIPTION
+pyenv lets you easily switch between multiple versions of Python\. It's simple, unobtrusive, and follows the UNIX tradition of single\-purpose tools that do one thing well\.
+.P
+ To start using pyenv
+.IP "1." 3
+\fBAppend\fR the following to \fB$HOME/.bashrc\fR
+.P
+.RS 15
+source /usr/share/pyenv/pyenv_user_setup.bash
+.RE
+.\"OR
+.\"\fBsh echo \-e \if command \-v pyenv 1>/dev/null 2>&1; then\en eval "$(pyenv init \-)"\enfi' >> ~/\.bashrc\fR
+.RS 3
+.P
+.nh
+Appending this line enables shims. Please make sure this line is placed toward the end of the shell configuration file since it manipulates \fBPATH\fR during the initialization\. 
+.hy
+.TP 13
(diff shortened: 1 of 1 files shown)


native-api (commented):
How would this be synchronized with the README.md and COMMANDS.md?
How would this get compiled and get into MANPATH if the program is installed with `git clone`?

karthek-git:
> How would this be synchronized with the README.md and COMMANDS.md?
the one who changes them also changes manpage
> How would this get compiled and get into MANPATH if the program is installed with `git clone`?
this doesnt need any compilation, its nroff and user can view with man -l /path/to/pyenv.1

native-api:
I'm not sure about having to maintain a separate instance of documentation.
Especially since it's distinct from both existing ones (so have to decide by hand what does where each time), and cannot be used the standard way (`man pyenv`).

Maintainers, any thoughts?

karthek-git:
It can be used in a standard way if it's installed in standard directory /use/share/man/man1 

I made it for debian package

anton-petrov:
Need some instructions in README.MD, how to install man page to use it by standard way `man pyenv`.
In this case it would be a good option to have manpage for pyenv.
@karthek-git, can you write them?

adigitoleo:
+1 for man page. It's a CLI tool, `man` is the community standard for CLI documentation... (at least on linux)

The point is to not need a browser and internet to get detailed docs (from README or whatever) when you need them. If it's really a bother, there are tools to convert man pages to PDF and html and probably to markdown (something like [this](https://github.com/mle86/man-to-md)).

> Need some instructions in README.MD, how to install man page to use it by standard way

Eh, that's kind of pointless IMO since it's done by packagers/distros but if you insist:

1. Check output of `manpath`, it should contain `/usr/share/man`
2. Compress the manpage with `gzip /path/to/pyenv.1`
3. Create `/usr/share/man/man1` if necessary (with `mkdir`)
3. `sudo cp /path/to/pyenv.1.gz /usr/share/man/man1`
4. Update the database `sudo mandb`

I think that's how it's done on most linux distros, but as I said this is kind of distro dependent.

Just my 2c

karthek-git:
this usually isn't a problem if you install it through your package manager
but anyway 
there are a lot of ways to use this manpage easily and here are some of them:

**1. INSTALLING TO MANPATH**
just cp to any directory in your $MANPATH.
(need to be superuser)

_NOTE: never mess with /usr/share/man though, as this directory is to be managed by your package manager, only install it into /usr/local/share or /usr/local/man based on your manpath_

Ex.
`cp pyenv.1 /usr/local/share/man/man1/`
(optional) compress it before cp.

**2.MODIFY MANPATH**
modify $MANPATH shell environment variable to include /path/to/pyenv/man/
Ex.
`echo 'MANPATH=$MANPATH:/path/to/pyenv/man/' >> ~/.bashrc`

**3.CREATE COMMAND ALIAS**
create shell command alias to man -l /path/to/pyenv.1
so that you can invoke simply by one word command
EX.
`echo 'alias mpyenv="man -l /path/to/pyenv.1"' >> ~/.bashrc`

Issue #1813, opened 2021-02-13
"Using script's directory as PYENV_DIR if shim is invoked with a script argument" feature is undocumented and broken


native-api:
Closes https://github.com/pyenv/pyenv/issues/1807

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): all
- [x] OS architecture (e.g. amd64): all
- [x] pyenv version: 1.2.22-61-ga0b02db7
- [x] Python version: all
- [x] C Compiler information (e.g. gcc 7.3): N/A
- [x] Please attach verbose build log as gist: see https://github.com/pyenv/pyenv/issues/1807#issue-803346431

#### Background

As of now, if a Pyenv shim is invoked with a script parameter, Pyenv uses that script's directory to determine the Python version to use rather than the current directory. The feature was initially added in Rbenv in https://github.com/rbenv/rbenv/commit/283e67b57e8ab0bbbe504aab6866729b0035186a .
**The problems with this feature are:**
* **It's undocumented and thus surprising**
* **It's broken and it's unclear if it can be fixed at all**

---

#### It's undocumented and thus surprising

* The use case seems to be:
  * If a script is under a project tree with `.python-version`, the user probably intends to use that Python version whenever they run scripts from that project, even if they invoke them from elsewhere

[...]

Issue #1807, opened 2021-02-08
Python version changes depending on command line arguments


kazuya-morishita:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Amazon Linux 2 AMI (HVM), SSD Volume Type
- [x] OS architecture (e.g. amd64): x86_64
- [x] pyenv version: 1.2.21
- [x] Python version: 2.7.18 and 3.7.6
- [x] C Compiler information (e.g. gcc 7.3): 7.3.1 20180712 (Red Hat 7.3.1-9) (GCC)
- [ ] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

### Issue
Python version changes depending on command line arguments.

```
[ec2-user@******** foo]$ python bar.py -a ./without-pv/baz.txt
3.7.6 (default, Dec  8 2020, 06:20:44)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-9)]

[ec2-user@******** foo]$ python bar.py -a ./with-pv/baz.txt
2.7.18 (default, Aug 27 2020, 21:22:52)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-9)]
```

### Details
The bar.py is as below. That code displays python version by `print(sys.version)` regardless of the command line arguments.
```python
# bar.py
[...]

PR #1814, opened 2021-02-14
Remove the "Using script's directory as PYENV_DIR if shim is invoked with a script argument" feature

native-api:
### Prerequisite
  * For Ruby and its usecases which are different than Python, the feature may be not breaking things so much. I involved the feature's author in the issue below.
  - Closes https://github.com/pyenv/pyenv/issues/1813

### Description

Deletes the subj feature. This is one of the alternatives.

### Tests
 N/A

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -61,24 +61,6 @@ else
 fi
 export PYENV_ROOT
 
-# Transfer PYENV_FILE_ARG (from shims) into PYENV_DIR.
-if [ -z "${PYENV_DIR}" ]; then
-  if [ -n "${PYENV_FILE_ARG}" ]; then
-    if [ -L "${PYENV_FILE_ARG}" ]; then
-      PYENV_DIR="$(abs_dirname "${PYENV_FILE_ARG}")"
-    else
-      PYENV_DIR="${PYENV_FILE_ARG%/*}"
-    fi
-    export PYENV_DIR
-    unset PYENV_FILE_ARG
-  fi
-else
-  [[ $PYENV_DIR == /* ]] || PYENV_DIR="$PWD/$PYENV_DIR"
-  cd "$PYENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$PYENV_DIR'"
-  PYENV_DIR="$PWD"
-  cd "$OLDPWD"
-fi
-
 if [ -z "${PYENV_DIR}" ]; then
   PYENV_DIR="$PWD"
 fi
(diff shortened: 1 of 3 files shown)


llehtahw:
FAKE FEATURE, should have been removed, for years!
Thank you @native-api , it's almost the most beautiful PR I've met, although maybe no chance to be merged.

Anyway, advice to pyenv users, as a pyenv user:
* NEVER use `pyenv local` command.
* Delete all `.python-version` files on your disk.
* Use virtualenv if needed.

anton-petrov:
Requires a common team solution to merge.

libexec/pyenv
```diff
@@ -61,24 +61,6 @@ else
...
-    unset PYENV_FILE_ARG
-  fi
-else
-  [[ $PYENV_DIR == /* ]] || PYENV_DIR="$PWD/$PYENV_DIR"
-  cd "$PYENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$PYENV_DIR'"
-  PYENV_DIR="$PWD"
-  cd "$OLDPWD"
-fi
-
 if [ -z "${PYENV_DIR}" ]; then
```
yyuu, line 64: This is kinda off-topic, however, duplicating rbenv's behavior on handling of `RBENV_DIR` here might make sense at extent. At least it seems it's verifying the availability of given `RBENV_DIR` before exporting. It might be safer 🤔 

https://github.com/rbenv/rbenv/blob/d604acb78aeba583be95f08d45eeae430372beb9/libexec/rbenv#L60-L68

yyuu (approved):
I am sorry to be late :bow: I don't think there's not much existing & effective use case of this behavior. From that standpoint, I am fine with removing this behavior to prevent it behaving weirdly at some condition. This is just my guts feeling, though.

libexec/pyenv
native-api: All the `PYENV_DIR` handling code that I delete here is duplicated by the code just below that is missing in Rbenv.

What the deleted code does boils down to `PYENV_DIR = $(realpath PYENV_DIR)`.

Since that code below is sketchy, I'm fine with restoring the Rbenv lookalike and sorting the duplication out in a differrent PR.

native-api: > Since that code below is sketchy, I'm fine with restoring the Rbenv lookalike and sorting the duplication out in a differrent PR.

Whoops, can't do that!

Rbenv's lookalike code was removed in 19e2b9588a3824e007a7689f9a918336768e120f to fix a breakage!

anton-petrov: Since this feature is broken by design, it should be removed. 
I have looked at the proposed code for a long time, and  have no questions.

anton-petrov (approved):

native-api:
@anton-petrov Could you close https://github.com/pyenv/pyenv/issues/1807 as well?
The magic word in https://github.com/pyenv/pyenv/issues/1813#issue-807820300 didn't work :(

Issue #1349, opened 2019-05-29
[Feature Request] Support resolving python versions to some major or minor version


lynncyrin:
### What

I would like pyenv to support resolving python versions based on the latest major or minor version. In some fashion, I want to be able to say "python 2" or "python 3.4" and have pyenv make a decision about which specific python version (eg. `2.7.10` or `3.4.0`) that should be.

I imagine some "wildcard" or "latest" identifier may possibly be involved for the sake of reducing ambiguity (which I only mention because those are the words I searched while looking for this).

### Why

I, and at least a dozen other people, have used pyenv and came to the conclusion that this is a feature that fits well with pyenv. https://github.com/pyenv/pyenv/issues/602 and https://github.com/pyenv/pyenv/issues/541 provide evidence of that being the case.

### Requested Feedback

It's clear that in the past this feature has been considered, and determined to be not desirable. By creating this issue, I'm requesting that you (the maintainers) re-appraise that decision 🙂 I'll take this issue being closed as "we are explicitly deciding [again] that this feature isn't desired" and l leave the conversation there 👍

sztomi:
I agree - this is a feature we are missing, too. Tox doesn't differentiate between 3.6.4 and 3.6.8 by default (both are "py36") because they are expected to be backwards-compatible.

As it currently stands, I have a hard time updating the python versions to the latest point versions on our CI nodes because many of our projects have .python-versions that pin that version.

@joshfriend I'm curious about the previous reasoning for not adding this feature. If there is openness to the idea after all, I'd be happy to try and contribute a PR (no idea if this is over my head, I don't know the codebase at all).

ssbarnea:
I totally agree with @sztomi that pinning python minor versions is non maintainable. It gets even worse if you are one of those contributing to many python projects. In at least 9/10 cases when someone creates a `.python-versions`they only mean: i want `py27,py36,py37` -- they do not case about which patch release or if they ever care is to have a `>=3.6.1` because they want to avoid a bug with preivous release. So far I never seen a single case where someone wanted 3.6.1 deliberately when 3.6.2 was already available.

Another reason for accepting wildcards or ranges is cross platform availability. Maybe version 3.6.2 does not compile yet on MacOS but compiles on Linux, so bumping it would break other users.

Another issue is lack of scalability. Current design is forcing open-source contributors working on projects that use pyenv to keep an unrealistically big list of python versions installed because various projects update the file with their own speed (which sometimes can be even... not updating them at all).

[...]

Issue #602, opened 2016-05-15
Add "latest" aliases


nicktimko:
A small usability proposal: add "latest" (release) aliases, so something like `pyenv install ...`
- `2`, `2.x`, `2-latest` would install (as of now) 2.7.11, or
- `3.4`, `3.4.x`, `3.4-latest` would install 3.4.4.

I'm semi-indifferent to what scheme to use, as long as it's short and simple. Currently there are `X.Y` releases in the list, but could move them to `X.Y.0`?

Also perhaps when asking for a "latest" version, force an update of the version list?
##

tylucaskelley:
An alternative in the meantime (only tested on Mac OS 10.11):

```
PYTHON2_VERSION="$(pyenv install -l | grep -e '2.[0-9].[0-9]' | grep -v - | tail -1)"
PYTHON3_VERSION="$(pyenv install -l | grep -e '3.[0-9].[0-9]' | grep -v - | tail -1)"
```

As of today, they return 2.7.11 and 3.5.1 respectively, the latest versions available for each major Python release.

yyuu:
Basically I don't like to add such _implicit_ feature in pyenv itself. I believe this can be implemented as a plugin with [providing install hook](https://github.com/yyuu/pyenv/wiki/Authoring-plugins).

I confirmed that installing this hook script at `~/.pyenv/pyenv.d/install/latest.bash` worked on GNU/Linux at least.

```sh
DEFINITION_PREFIX="${DEFINITION%%@*}"                                                          
DEFINITION_TYPE="${DEFINITION_PREFIX%%-*}" # TODO: support non-CPython versions                
if [[ "${DEFINITION}" != "${DEFINITION_PREFIX}" ]]; then                                       
  DEFINITION_CANDIDATES=($(python-build --definitions | grep -F "${DEFINITION_PREFIX}" | grep "^${DEFINITION_TYPE}" | sed -e '/-dev$/d' -e '/-src$/d' | sort -t. -k1,1r -k 2,2nr -k 3,3nr || true))
  DEFINITION="${DEFINITION_CANDIDATES}"                                                        
  VERSION_NAME="${DEFINITION##*/}"
fi
```

With it, you can install the _latest_ version like this.

```
$ pyenv install -v 3.5@latest
```

yyuu:
Actually I'm not sure how much portable `sort(1)`'s version sorting is. I guess it doesn't work well with BSD toolchain.

[...]

PR #1831, opened 2021-02-25
Hook script to add latest suffix for install command

anton-petrov:
The main idea of this PR is to add the ability to install the latest version of python, using such syntax:

`pyenv install 3.9:latest`
`pyenv install 3.8:latest`
`pyenv install 3:latest`
`pyenv install pypy3.7-7.3:latest`
and so on…

As for the `:` symbol, i relied on docker syntax as the most clear and used for its images versioning.

Added code from [yyuu comment](https://github.com/pyenv/pyenv/issues/602#issuecomment-219896567) as a starting point to be able to install the latest version.

### Prerequisite
  - https://github.com/pyenv/pyenv/issues/1349

### Description
- [x] Despite the recommendation to implement this as a plugin, it seems to me that this functionality should be int the pyenv repo, since it is requested by many users. Also it doesn't change the main logic of pyenv, nor other users workflows.

### Tests
- No tests.

### DIFF
--- a/pyenv.d/install/latest.bash
+++ b/pyenv.d/install/latest.bash
@@ -1,0 +1,7 @@
+DEFINITION_PREFIX="${DEFINITION%%:*}"
+DEFINITION_TYPE="${DEFINITION_PREFIX%%-*}" # TODO: support non-CPython versions
+if [[ "${DEFINITION}" != "${DEFINITION_PREFIX}" ]]; then
+  DEFINITION_CANDIDATES=($(python-build --definitions | grep -F "${DEFINITION_PREFIX}" | grep "^${DEFINITION_TYPE}" | sed -e '/-dev$/d' -e '/-src$/d' | sort -t. -k1,1r -k 2,2nr -k 3,3nr || true))
+  DEFINITION="${DEFINITION_CANDIDATES}"
+  VERSION_NAME="${DEFINITION##*/}"
+fi


brechtm:
Thanks for this!

I was hoping the _:latest_ suffix would also be supported in _.python-version_ files. Alas, setting "3.9:latest" in _.python-version_ results in:
```
pyenv: version `3.9' is not installed (set by /Users/brechtm/Documents/Code/rinohtype/.python-version)
pyenv: version `latest' is not installed (set by /Users/brechtm/Documents/Code/rinohtype/.python-version)
```

Just wondering whether this is planned or perhaps already possible by means of other syntax.

xmo-odoo:
I've encountered a pair of issues when trying to use `:latest` on non-cpython as well (but in the basic "install" context):

- `pypy:latest` gives me `pypy-stm-2.5.1` which is somewhat unexpected: it's a 6 years old version of the mostly-abandoned STM effort.
- even going down as far as `pypy3.8:latest` (at which point that takes more effort than copying the version I want) it still does not find the semingly-current-latest `pypy3.8-7.3.7` and tries installing `pypy3.8-7.3.6`.

Also `:latest` might be a bit of an UB but it tries installing stackless-3.7.5. Maybe `:latest` alone should be special-cased and refuse resolving anything? Possibly try printing a few prefixes?

PR #1844, opened 2021-03-08
Update GET_PIP_URL

mateumann:
The bootstrap.pypa.io URLs for pip installation changed.  The pip maintainers ask to use an upgraded URL.


martin-denizet (approved):
Thanks, seems to be the change we need indeed!

mateumann:
@anton-petrov care to review?

Issue #1743, opened 2020-11-27
Some suggestions for friendly prompts during installation


2lovecode:
When I install using local cache，I download Python-3.3.0.tgz file and put it to ~/pyenv/cache.
but when I execute command "pyenv install 3.3.0",  it still download the file and don't use the cache file.

This confuses me and it took me a lot of time to find the problem.

Eventually I found that the problem was the file name is wrong, it must be Python-3.3.0.tar.gz! I should rename Python-3.3.0.tgz to Python-3.3.0.tar.gz! so frustrating！

So, I strongly recommend when using local cache to install, provide file name verification function, and give a friendly prompt！This is useful and necessary

native-api:
I'm not sure what you expect that "file name verification function" to do. Second-guessing the user, trying to fuzzy-match existing files somehow? Fuzzy matching is a double-edged sword and is fundamentally prone to errors.

What if two legitimate downloaded files have similar names? What if you name the file Python-3.3.tgz -- or, worse, confuse the version? This is an exercise in futility.

You are not supposed to put files into the cache yourself, `pyenv install` is supposed to do that -- since the file must be _the exact one_ specified in the installation script for that release, complete with a specific checksum. If your file has a different name, chances are, it's some different file so it wouldn't fit anyway (if this was the same file, there'd be no reason to change its name).

2lovecode:
> I'm not sure what you expect that "file name verification function" to do. Second-guessing the user, trying to fuzzy-match existing files somehow? Fuzzy matching is a double-edged sword and is fundamentally prone to errors.
> 
> What if two legitimate files to download have similar names? What if you name the file Python-3.3.tgz -- or, worse, confuse the version or make a typo? This is an exercise in futility.
> 
> You are not supposed to put files into the cache yourself, `pyenv install` is supposed to do that -- since the file must be _the exact one_ specified in the installation script for that release, complete with a specific checksum. If your file has a different name, chances are, it's some different file so it wouldn't fit anyway (if this was the same file, from the same official source, there'd be no reason to change its name).

Because of network reasons, I cannot download the python package through pyenv install.
So I have to get the installation package from other methods and put it in the cache directory. (local installation maybe)
Maybe the file name is wrong.

What I want is just a prompt, not a fuzzy-match. 

For example, when Python-3.3.0.tar.gz cannot be found in the cache,  if can give a friendly prompt like "Python-3.3.0.tar.gz package not found", it can help users to find the real problem.

[...]

PR #1908, opened 2021-05-08
Report cache filename at download

native-api:

### Prerequisite
* [x] Please consider contributiprefer ng the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  - Closes https://github.com/pyenv/pyenv/issues/1743

### Description

In certain cases, a user wants to know the cached filename to add the file themselves,
see https://github.com/pyenv/pyenv/issues/1743 .
Since we report both a filename and a URL anyway, there's no reason to report a wrong one.

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -417,8 +417,8 @@ fetch_tarball() {
    fi
 
   if ! reuse_existing_tarball "$package_filename" "$checksum"; then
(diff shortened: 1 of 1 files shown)


native-api (commented):
If we're going to _guarantee_ that the reported name is the cache name, need to test that...

Issue #1642, opened 2020-06-25
The setup for fish env appends pyenv path every login.


kobayashi:
Setup for fish env described in `README` appends pyenv path to fish_variables in every fish-shell starting.

```
set -Ux PYENV_ROOT $HOME/.pyenv
set -Ux fish_user_paths $PYENV_ROOT/bin $fish_user_paths
```

### Description
- [ ] Platform information (e.g. Ubuntu Linux 16.04): OSX 10.14.5 (18F132)
- [ ] OS architecture (e.g. amd64): 
- [ ] pyenv version: pyenv 1.2.19-1-gece59ca8
- [ ] Python version:  3.7.7

This is because universal variables are stored in the file .config/fish/fish_variables. The set command should be like this.
```
set -Ux fish_user_paths $PYENV_ROOT/bin
```

berislavlopac:
The problem with the solution suggested above is that it will overwrite the value of `$fish_user_paths` if it already exists. I would recommend using the following line instead:

```
contains $PYENV_ROOT/bin $fish_user_paths; or set -Uax fish_user_paths $PYENV_ROOT/bin
```

PR #1916, opened 2021-05-10
Fix fish installation instructions

faho:
### Prerequisite

Not applicable, rbenv does not have these instructions to begin with.

* [x]  My PR addresses the following pyenv issue (if any)
  -  https://github.com/pyenv/pyenv/issues/1642 (as one part of it)

### Description

The instructions for fish are broken at the moment in numerous respects, they

1. Recommended appending to a universal variable in config.fish, which
will baloon the variable and end up breaking the shell
2. Recommended using ~/.profile for fish, which it does not use

This changes them to be in line with what I'd recommend. See also https://stackoverflow.com/questions/67473902/pyenv-no-longer-sets-paths-correctly-when-activating-virtual-environments/67476295#67476295 where this tripped someone up quite badly.

For easier reading, this is the message it ends up printing:

```
# Add pyenv executable to PATH by running
# the following interactively:

set -Ux PYENV_ROOT $HOME/.pyenv
set -Ux fish_user_paths $PYENV_ROOT/bin $fish_user_paths

# Load pyenv automatically by appending
# the following to ~/.config/fish/config.fish:

pyenv init - | source
pyenv init --path | source
```

which, as far as I know, should be correct.

### Tests

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -100,8 +100,8 @@ function help_() {
       echo "# Load pyenv automatically by appending"
       echo "# the following to ~/.config/fish/config.fish:"
       echo
-      echo 'pyenv init - | source'
       echo 'status is-login; and pyenv init --path | source'
+      echo 'pyenv init - | source'
       ;;
     * )
       echo "# Add pyenv executable to PATH by adding"
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -83,58 +83,51 @@ function help_() {
   ksh )
     profile='~/.profile'
     ;;
-  fish )
-    profile='~/.config/fish/config.fish'
-    ;;
   * )
     profile='your profile'
     ;;
   esac
 
-  { echo "# Add pyenv executable to PATH by adding"
-    echo "# the following to ~/.profile:"
-    echo
+  {
     case "$shell" in
     fish )
+      echo "# Add pyenv executable to PATH by running"
+      echo "# the following interactively:"
+      echo
       echo 'set -Ux PYENV_ROOT $HOME/.pyenv'
       echo 'set -Ux fish_user_paths $PYENV_ROOT/bin $fish_user_paths'
-      ;;
-    * )
-      echo 'export PYENV_ROOT="$HOME/.pyenv"'
-      echo 'export PATH="$PYENV_ROOT/bin:$PATH"'
-      ;;
-    esac
-    echo
-    echo "# Load pyenv automatically by appending"
-    echo "# the following to ${profile}:"
-    echo
-    case "$shell" in
-    fish )
+      echo
+      echo "# Load pyenv automatically by appending"
+      echo "# the following to ~/.config/fish/config.fish:"
+      echo
       echo 'pyenv init - | source'
-      ;;
-    * )
-      echo 'eval "$(pyenv init -)"'
-      ;;
-    esac
-    echo
-    echo "# and the following to ~/.profile:"
-    echo
-    case "$shell" in
-    fish )
       echo 'pyenv init --path | source'
       ;;
     * )
+      echo "# Add pyenv executable to PATH by adding"
+      echo "# the following to ~/.profile:"
+      echo
+      echo 'export PYENV_ROOT="$HOME/.pyenv"'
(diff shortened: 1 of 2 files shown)


libexec/pyenv-init
```diff
@@ -83,58 +83,51 @@ function help_() {
...
-    echo "# the following to ~/.profile:"
-    echo
+  {
     case "$shell" in
     fish )
+      echo "# Add pyenv executable to PATH by running"
+      echo "# the following interactively:"
+      echo
       echo 'set -Ux PYENV_ROOT $HOME/.pyenv'
       echo 'set -Ux fish_user_paths $PYENV_ROOT/bin $fish_user_paths'
```
native-api, line 98: These two need to only run once, even if you start a nested shell so that the `PATH` entries don't duplicate.

libexec/pyenv-init
```diff
@@ -83,58 +83,51 @@ function help_() {
...
+      echo
+      echo "# Load pyenv automatically by appending"
+      echo "# the following to ~/.config/fish/config.fish:"
+      echo
+      echo 'pyenv init - | source'
+      echo 'status is-login; and pyenv init --path | source'
```
native-api, line 104: `pyenv init --path` needs to come first, and it also only needs to run once as per above.

native-api (changes requested):
* See https://github.com/pyenv/pyenv/issues/1914#issuecomment-836607473 on what needs to be done. I've no experience with Fish so need someone else's expertise on how to better do that.
* You've lumped all the instructions together. Previously, the ones for adding it to PATH and installing it to the shell were separate
* Would prefer to avoid duplicated code if possible
* Please also update the instructions in the README

faho:
>You've lumped all the instructions together. Previously, the ones for adding it to PATH and installing it to the shell were separate

Fish doesn't have the .bashrc/.profile distinction. There is no other configuration file to add `pyenv init --path` to.

Instead what you'd do is to check `status is-login` to only do something for login shells.

libexec/pyenv-init
faho: It is already only run for login shells, just like for other shells via .profile, so it's exactly as "once" as for those.

The alternative is to make the code it prints check if anything needs to be done, e.g. via checking if the entries are already in.

E.g. this currently prints

```fish
set -gx PATH '/shims' $PATH
```

If it just did

```fish
contains -- /shims $PATH; or set -gx PATH /shims $PATH
```

it would only add the shims directory if necessary. That way this wouldn't have to be conditioned on login shells. The same could be done for `pyenv init -`.

I can change the order.

native-api: > It is already only run for login shells...

Okay then.

> I can change the order.

Please do -- because `pyenv init -` does use the result of `pyenv init --path`.

faho:
>What I meant is that the pyenv init - step was described in a separate section.
Moving them all together resulted in much duplicated code.

I'm not sure what duplicated code you're referring to?

--------------

Also, just to head this off:

> set -Ux PATH $PYENV_ROOT/shims $PATH

A universal $PATH won't work. $PATH is inherited from the parent, and inherited variables take precedence over globals. That is why $fish_user_paths exists - its value is added to $PATH (in a deduplicated manner).

native-api:
> I'm not sure what duplicated code you're referring to?

The section headers.
(Not so much, actually...)

native-api:
> The section headers.
> (Not so much, actually...)

Okay, scratch that.

faho:
>Okay, now that you've addresed all of the concerns --
could you now synchronise the README with the changes

I have now. Note that 91dcaf42cdc3e38b0cf7ea24fc7bf0d377380c63 also adds instructions for when fish isn't the login shell - this happens when people keep $SHELL as e.g. bash, because many operating systems insist on putting important configuration into /etc/profile and friends, and instead start fish via their terminal.

In that case they would have to follow the ~/.profile instructions *in addition to* the fish instructions (well, technically the path-setting could be skipped, but since it only applies to login shells already...).

There is also the option of not caring about loginshell-ness entirely and instead checking if the components are included. That's easy - `contains -- /path/to/wherever $PATH; or set PATH /path/to/wherever $PATH`. But it's a bigger change that might warrant a separate discussion/PR.

native-api (approved):

Issue #1918, opened 2021-05-11
[FeatureRequest] static shim


loynoir:
# Scenario
There is a cli tool `foobar` at python 2.7, not support 3.7.

And likely will not.

To avoid things got breaked, installed within virtualenv.

# Background
slightly checked [source](https://github.com/pyenv/pyenv/blob/fd3c891d766a091f2dd560701953de722c2b757d/libexec/pyenv-rehash#L64) seems not implemented, so here is a feature request.

# Prepare
```txt
$ git clone \
    https://github.com/pyenv/pyenv-virtualenv.git \
    $(pyenv root)/plugins/pyenv-virtualenv
```

# Actual

Install package
```txt
$ pyenv virtualenv 2.7.10 foobar
$ env \
    PYENV_VERSION=2.7.10/envs/foobar \
    pip install foobar
```

Usage
```txt
$ env \
    PYENV_VERSION=2.7.10/envs/foobar \
    foobar --help
```
Every time call it, needs to explicitly specific with an older python version.

# Expected
Install package
```txt
$ pyenv virtualenv 2.7.10 foobar
$ env \
    PYENV_SHIM_VERSION_LOCK=y \
    PYENV_VERSION=2.7.10/envs/foobar \
    pip install foobar
```

Usage
```txt
$ foobar --help
```
Every time call it, no need to explicitly specific with an older python version.

native-api:
Looks like **what you _actually_ want is to be able to run a tool that belongs to a Pyenv-based installation from elsewhere, as if it was installed globally.**

For this, **a working solution would be to make a separate set of manually-configured "static shims" in another directory:**

* Create another directory and place it into `PATH` _behind_ `shims`
* Add shims there _for the needed executables only_ that would run with a statically defined `PYENV_VERSION`
* Add commands to manage this set: view status, add/overwrite, delete

This way, you'll be able to run a static shim from anywhere _and_ override by installing the cli into the active version: the regular shim will fall through to the static one if the executable is not present in the active version.

This looks like a job for a plugin to me since this is a significant new feature that's only of interest to a subset of users.

loynoir:
Thanks for your reply. Yeah, `static shim` might be more proper name. Got plugin 
(loynoir/pyenv-shim-version-lock)[https://github.com/loynoir/pyenv-shim-version-lock] to make `static shim` at last to fit my need.

PR #1919, opened 2021-05-11
static shim

loynoir:

### Prerequisite
* [] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  - https://github.com/pyenv/pyenv/issues/1918

### Description
if `PYENV_SHIM_VERSION_LOCK`
1) not set or empty, generate 100% same shim
2) set to y, shim will use same version as install.
3) set to n, shim will record the version using while install.

### Tests
- [] My PR adds the following unit tests (if any)


native-api (changes requested):
* Will make a mess of the shims. When you switch the version, some shims will change, some not, some are stuck at one version, some at another, and you'll never know which is which.
* Will block you from using the same CLI from another version _in any way_
* Will be overwritten at startup and at any other time when `rehash` is run without the `LOCK` envvar set (see `remove_outdated_shims`).
   * If the envvar is always set, will break version switching 'cuz the locking will be applied tio _all_ shims, including `python*`, `pip*` etc.

loynoir:
Static shim is design for outdated package cli.

So, switch version is designed to not work under any conditions.

But, yes, both implements lack of ability to show which shim is locked. 🙃

At last plugin implement fit my need.

Issue #1915, opened 2021-05-10
WARNING: `pyenv init -` no longer sets PATH.


The-Judge:
Hi everyone,

I am aware of #1906 , but the solution seems to not do the trick for me:

I set everything to `.bashrc` and `.profile` as suggested in the output of `pyenv init`, but still there is this warning printed, causing issues with sftp and similar:

```
~ $ ssh judgepi /bin/true
WARNING: `pyenv init -` no longer sets PATH.
Run `pyenv init` to see the necessary changes to make to your configuration.
~ $
```

There are no other login/shell relevant files than `.bashrc` and `.profile`:
```
root@judgepi:~# ls -al | egrep -v ^d
total 152
-rw-------  1 root root 11209 May 10 20:02 .bash_history
-rw-r--r--  1 root root   873 May 10 19:52 .bashrc
-rw-r--r--  1 root root   748 Apr  3 09:48 .bashrc.backup
-rw-------  1 root root   102 Apr 22 11:29 .lesshst
-rw-------  1 root root   204 Nov 27 19:06 .netrc
-rw-r--r--  1 root root   252 May 10 19:40 .profile
-rw-------  1 root root    44 Nov 26 19:54 .python_history
-rw-r--r--  1 root root    66 Jul 28  2019 .selected_editor
-rw-------  1 root root 20183 May 10 20:00 .viminfo
lrwxrwxrwx  1 root root    30 Dec 12  2019 .virtualenvs -> /root/.local/share/virtualenvs
-rw-r--r--  1 root root   255 Apr 20 22:37 .wget-hsts
root@judgepi:~#
```

### Description
- Platform information (e.g. Ubuntu Linux 16.04): Raspbian GNU/Linux 10 (buster)
- OS architecture (e.g. amd64): armv7l
- pyenv version: 1.2.27-29-gfd3c891d
- Python version: 3.7.3
- C Compiler information (e.g. gcc 7.3): gcc 8.3.0
- [.bashrc and .profile gist](https://gist.github.com/The-Judge/94c9c599a980b08a65aa7ad9a9d956d6)

aiguofer:
I was just trying to solve the same issue. The problem is that the warning is just a "dumb" message:

```
❯ pyenv init -
echo 'WARNING: `pyenv init -` no longer sets PATH.'
echo 'Run `pyenv init` to see the necessary changes to make to your configuration.'
export PYENV_SHELL=zsh
[...]

Issue #1913, opened 2021-05-10
path manipulations need to be placed in ~/.zprofile for zsh user


hrzhu:
I'm using zsh on Ubuntu 20.04. I need to place the following path manipulations in ~/.zprofile instead of ~/.profile to make pyenv work and suppress the recently added warning.

```
export PYENV_ROOT="$HOME/.pyenv"
export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init --path)"
```

europ:
After pyenv update, I had the same problem. @hrzhu who shared the solutions that can suppress the warning(s). Thanks!

#### Problem
```
WARNING: `pyenv init -` no longer sets PATH.
Run `pyenv init` to see the necessary changes to make to your configuration.
WARNING: `pyenv init -` no longer sets PATH.
Run `pyenv init` to see the necessary changes to make to your configuration.
```

#### Solution
You just need to create a `~/.zprofile` with the following content (see below).

#### ~/.zshrc
```bash
# kept content in ~/.zshrc
export PYENV_ROOT="$HOME"/.pyenv
export PATH="$PYENV_ROOT"/bin:"$PATH"
eval "$(pyenv init -)"
eval "$(pyenv virtualenv-init -)"
```

#### ~/.zprofile
```bash
# added content to ~/.zprofile
export PYENV_ROOT="$HOME/.pyenv"
export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init --path)"
```

#### Question
How can I configure pyenv using only `~/.zshrc` (without `~/.zprofile`)?

native-api:
See https://github.com/pyenv/pyenv/issues/1914#issuecomment-836607473 for what needs to be done at startup.
I don't have any experience with `zsh` outside of OSX so can't say how it's best to implement this for Zsh.

Doesn't it read `~/.profile` like Sh and Bash do? Maybe it should since it's Sh-compatible?
Is it your login shell or not? If it is, what happens if you enter a GUI session -- do any of Zsh-specific files have an effect on it?

native-api:
Okay. Then, I see that if I run `zsh` in a GUI session in a terminal window in OSX, it thinks that it's a login shell -- even though it's technically not.

Does it behave the same way in Linux? (If it doesn't, we'll need to put commands into _both_ `~/.profile` and `~/.zprofile`.)

native-api:
[...]

PR #1898, opened 2021-05-04
Split startup logic into PATH and everything else

native-api:
Make sure you have checked all steps below.

### Prerequisite
* [x] Please consider implementing the feature as a hook script or plugin as a first step.
  * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins](https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a plugin if possible.
* [ ] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv will be imported to pyenv too, eventually.
  * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with rbenv.
* [x] My PR addresses the following pyenv issue (if any)
  - Closes https://github.com/pyenv/pyenv/issues/1649

### Description
- [x] Here are some details about my PR

* Implements https://github.com/pyenv/pyenv/issues/1649#issuecomment-694388530 .
* Adds a warning on shell startup if ~./profile` entry is not present -- to facilitate migration
* (unrelated) fixes expanding `PATH` in one of test descriptions

### Tests
- [x] My PR adds the following unit tests (if any)

* Test for the added warning

PR #1920, opened 2021-05-11
Update install instructions for Bash and Zsh

native-api:
* See https://github.com/native-api/pyenv/tree/bash_zsh_instructions#basic-github-checkout for the rendered README
* Clone https://github.com/native-api/pyenv/tree/bash_zsh_instructions and run `pyenv init [<shell>]` to see text instructions for `<shell>`:
```
pushd "$(pyenv root)"
git remote add pr https://github.com/native-api/pyenv
git fetch pr
git status    # see what branch/tag/commit you're currently at
git checkout bash_zsh_instructions

pyenv init

#undo the changes
git checkout <old branch/tag/commit from `git status` above>
git branch -D bash_zsh_instructions
git remote remove pr
popd
```

---

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/1913
  - Closes https://github.com/pyenv/pyenv/issues/1915

### Description

Updated README and instructions for Bash and Zsh based on user feedback.
Going to ask for reviews before merging.

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,12 @@ easy to fork and contribute any changes back upstream.
         ~~~ bash
         echo 'eval "$(pyenv init -)"' >> ~/.bashrc
         ~~~
+        
+        - **If your `/etc/profile` sources `~/.bashrc` (SUSE):**
+        
+          ~~~bash
+          echo 'if command -v pyenv >/dev/null; then eval "$(pyenv init -)"; done' >> ~/.bashrc
+          ~~~
 
       - For **Zsh**:
         ~~~ zsh
--- a/README.md
+++ b/README.md
@@ -206,9 +206,9 @@ easy to fork and contribute any changes back upstream.
 
 2. **Configure your shell's environment for Pyenv**
 
-   **Note:** The below instructions and code samples are intended for common shell setups.  
-   If you have an uncommon setup and/or they don't work for you,
-   use the guidance text and [Advanced Configuration](#advanced-configuration)
+   **Note:** The below instructions for specific shells are designed for common shell setups.  
+   If you have an uncommon setup and they don't work for you,
+   use the guidance text and the [Advanced Configuration](#advanced-configuration)
    section below to figure out what you need to do in your specific case.
    
    **MacOS note:** If you installed Pyenv with Homebrew, you don't need
--- a/README.md
+++ b/README.md
@@ -205,15 +205,42 @@ easy to fork and contribute any changes back upstream.
         cd ~/.pyenv && src/configure && make -C src
 
 2. **Define environment variable `PYENV_ROOT`** to point to the path where
-   pyenv repo is cloned and add `$PYENV_ROOT/bin` to your `$PATH` for access
-   to the `pyenv` command-line utility.
+   pyenv repo is cloned, add `$PYENV_ROOT/bin` to your `$PATH` for access
+   to the `pyenv` command-line utility, enable shims to be able to use Pyenv.
 
-   - For **bash**/**Zsh**:
-     ~~~ bash
-     echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.profile
-     echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile
-     echo 'eval "$(pyenv init --path)"' >> ~/.profile
-     ~~~
+   - For **bash**:
+
+      - **If your `~/.profile` sources `~/.bashrc` (Ubuntu):**
+         <!--This way is an alternative option and needn't be replicated to `pyenv init`-->
+         ~~~ bash
+         echo -e 'if shopt -q login_shell; then' \
+               '\n  export PYENV_ROOT="$HOME/.pyenv"' \
+               '\n  export PATH="$PYENV_ROOT/bin:$PATH"' \
+               '\n eval "$(pyenv init --path)"' \
+               '\nfi' >> ~/.bashrc
+         ~~~
+
+         Or, put these commands into `~/.profile` _before_ the part that sources `~/.bashrc`:
+
+         ~~~bash
+         export PYENV_ROOT="$HOME/.pyenv"
+         export PATH="$PYENV_ROOT/bin:$PATH"
+         eval "$(pyenv init --path)"
+         ~~~
+
+      - **If your `~/.profile` doesn't source `~/.bashrc` or doesn't exist (Debian):**
+         ~~~ bash
+         echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.profile
+         echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile
+         echo 'eval "$(pyenv init --path)"' >> ~/.profile
+         ~~~
+
+      If you have `~/.bash_profile`, make sure that it also executes the commands
+      that you've added above, this way or another.
+
+   - For **Zsh**:
+
+     Same as for Debian above, but add the commands to both `~/.profile` and `~/.zprofile`.
 
    - For **Fish shell**:
 
@@ -232,18 +259,16 @@ easy to fork and contribute any changes back upstream.
 
    - **Proxy note**: If you use a proxy, export `http_proxy` and `https_proxy` too.
 
-3. **Add `pyenv init` to your shell** to enable shims and autocompletion.
-   Please make sure `eval "$(pyenv init -)"` is placed toward the end of the shell
-   configuration file since it manipulates `PATH` during the initialization.
+3. **Add `pyenv` to your shell** to enable autocompletion and all the commands.
 
    - For **bash**:
      ~~~ bash
-     echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bashrc
+     echo 'eval "$(pyenv init -)"' >> ~/.bashrc
      ~~~
 
    - For **Zsh**:
      ~~~ zsh
-     echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.zshrc
+     echo 'eval "$(pyenv init -)"' >> ~/.zshrc
      ~~~
 
    - For **Fish shell**:
@@ -253,16 +278,13 @@ easy to fork and contribute any changes back upstream.
      ~~~
 
     **General warning**: There are some systems where the `BASH_ENV` variable is configured
-    to point to `.bashrc`. On such systems you should almost certainly put the above mentioned line
+    to point to `.bashrc`. On such systems you should almost certainly put the above-mentioned line
     `eval "$(pyenv init -)"` into `.bash_profile`, and **not** into `.bashrc`. Otherwise you
     may observe strange behaviour, such as `pyenv` getting into an infinite loop.
     See [#264](https://github.com/pyenv/pyenv/issues/264) for details.
 
-4. **Restart your shell so the path changes take effect.**
-   You can now begin using pyenv.
-    ```sh
-    exec "$SHELL"
-    ```
+4. **Restart your login session so the path changes take effect.**
+   E.g. if you're in a GUI session, you need to fully log out and log back in.
 
 5. [**Install Python build dependencies**](https://github.com/pyenv/pyenv/wiki#suggested-build-environment) before attempting to install a new Python version.
 
@@ -364,9 +386,9 @@ opposed to this idea. Here's what `pyenv init` actually does:
 To see exactly what happens under the hood for yourself, run `pyenv init -`.
 
 If you don't want to use `pyenv init` and shims, you can still benefit
-from pyenv's ability to install Python versions for you. Just run 
-`pyenv install` and you will find versions installed in 
-`$(pyenv root)/versions`, which you can manually execute or symlink 
+from pyenv's ability to install Python versions for you. Just run
+`pyenv install` and you will find versions installed in
+`$(pyenv root)/versions`, which you can manually execute or symlink
 as required.
 
 ### Uninstalling Python Versions
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ For example, you might have these versions installed:
 * `$(pyenv root)/versions/3.4.2/`
 * `$(pyenv root)/versions/pypy-2.4.0/`
 
-As far as pyenv is concerned, version names are simply the directories in
+As far as Pyenv is concerned, version names are simply directories under
 `$(pyenv root)/versions`.
 
 ### Managing Virtual Environments
--- a/README.md
+++ b/README.md
@@ -174,116 +174,131 @@ For pyenv to install python correctly you should [**install the Python build dep
 
 ### Homebrew on macOS
 
-   1. Consider installing with [Homebrew](https://brew.sh)
+   1. Consider installing with [Homebrew](https://brew.sh):
       ```sh
       brew update
       brew install pyenv
       ```
-   2. Then follow the rest of the post-installation steps under [Basic GitHub Checkout](https://github.com/pyenv/pyenv#basic-github-checkout), starting with #3 ("Add `pyenv init` to your shell to enable shims and autocompletion").
+   2. Then follow the rest of the post-installation steps under [Basic GitHub Checkout](https://github.com/pyenv/pyenv#basic-github-checkout), starting with #2 ("Configure your shell's environment for Pyenv").
 
-If you're on Windows, consider using @kirankotari's [`pyenv-win`](https://github.com/pyenv-win/pyenv-win) fork. (`pyenv` does not work on windows outside the Windows Subsystem for Linux)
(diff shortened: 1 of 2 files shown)


agoose77:
@native-api thanks for creating a PR to try and tackle this. 

### The Issue
I feel that with all of this work on documenting exactly what to do for each shell, we're trying to solve the wrong problem; shell RC files are confusing, and not always consistent between shells and environments.

All of the issues that have been raised about this have stemmed from the fact that `pyenv` has several locations that need to end up in `PATH`, and users either don't know of all of them, or don't understand why they need to go in different RC files. This is not a criticism of these users; I have to reference what each shell does when I reply to these issues.

### Proposed Change
I think that we should consider dropping the shell-specific approach entirely, and instead `pyenv` should just expose two commands:
```bash
pyenv init --path
```
and
```bash
pyenv init --shell
```

These are the two kinds of configuration that `pyenv` needs to set AFAICT. The `pyenv init --shell` command is an attempt to remove the visual precedence that `pyenv init -` has over `pyenv init --path`, and to make it clear that one is for shell functions, the other is for environment.

### API Changes
Of course, this means that things will change for users. I don't know what the policy for API change is for this project, but if we don't want to break `pyenv init -` outright, then the safest approach is just to add a new `--shell` flag and retain the existing `pyenv init -` behaviour (but mark as deprecated). However, I would perhaps go further, and just make `pyenv init -` return `cat <(echo "warning ...") <(pyenv init --path) <(pyenv init --shell)`. This way, users who haven't modified their RC files will be notified of the upcoming deprecation, but will not be confused by breakages. 

### Shell-Specific Help
I still think that the shell-specific information is useful, but perhaps it doesn't belong in the tool. If we repurpose `pyenv init -` as an old-style (deprecated) configuration command, then we can add some information about what the new-style looks like, e.g.

```bash
cat <<- 'EOF' 
Warning, `pyenv init -` will no longer work in future releases of pyenv.
To avoid breakages when this occurs, please move to the new configuration mechanism.

The old `pyenv init -` command has been split into two new commands:
`pyenv init --path` and `pyenv init --shell`. 

You should remove the existing `pyenv init -` code, and instead 
1. Add `pyenv init --path` to your `~/.profile` (or equivalent) file
2. Add `pyenv init --shell` to your `~/.bashrc` (or equivalent) file

In some cases you might find that this does not work. This is likely because shell configuration is hard! 
Please visit .... for more information with shell-specific guidance.
```

The-Judge:
No offense, but reading all the replies in #1915 to my initial request, it looks a bit as the thread got "hijacked" for "_all-things-shell-issues_" ;-)
Initially, I reported an issue for "bash" shell. Here's what I did:

1. I cloned #1920 to `/usr/src/pyenv` and cd'ed into it
2. `src/configure && make -C src` (successfully)
3. `export PYENV_ROOT="/usr/src/pyenv"`
4. `export PATH="${PYENV_ROOT}/bin:${PATH}"`
5. `which pyenv` (provides `/usr/src/pyenv/bin/pyenv`)
6. Both, `pyenv init` and `pyenv init bash` provide:
```
# Load pyenv automatically by appending
# the following to ~/.bash_profile:

eval "$(pyenv init -)"
```
Nothing more; nothing about setting the `PATH` or anything; which seems wrong already, isn't it?

Anyways: I removed everything pyenv related from my `.bashrc` and `.profile` already before running this. I started with 1:1 doing what `pyenv init` suggests, knowing this will fail without the PATH manipulation and - to very few surprises - it did:

```
-bash: pyenv: command not found
```

So, I also added this before that line in `.bash_profile`, making it look like this:

```
root@judgepi:~# cat .bash_profile
#source ${HOME}/.bash_pyenv
export PYENV_ROOT="/usr/src/pyenv"
export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init -)"

root@judgepi:~#
```

Actually, it looks like this did the trick! I can use pyenv and do no longer receive the WARNING when SSH'ing into it:

```
~ $ ssh pi /bin/true
~ $ echo $?
0
~ $
```

The-Judge:
PS: I like @agoose77 's idea! Debugging the global RC-landscape is out of scope for pyenv; at least "core". Supporting all shells and advice on them seems to be more related to sidekick-projects like [pyenv/pyenv-installer](https://github.com/pyenv/pyenv-installer).
As long as it is clear and transparent to the user what's the goal he/she has to take care about, I feel like pyenv's duty is done.

native-api:
> I did by modifying the PATH to that new location - didn't I?

Oh, sorry, I missed that part (I honestly thought that no-one would go for the trouble :-) ). Then it indeed should work.

How exactly did you "clone #1920"? I suspect that you didn't check out the right branch and ended up at `master` from my fork -- which is stale and unused.

README.md
```diff
@@ -205,15 +205,42 @@ easy to fork and contribute any changes back upstream.
         cd ~/.pyenv && src/configure && make -C src
 
 2. **Define environment variable `PYENV_ROOT`** to point to the path where
-   pyenv repo is cloned and add `$PYENV_ROOT/bin` to your `$PATH` for access
-   to the `pyenv` command-line utility.
+   pyenv repo is cloned, add `$PYENV_ROOT/bin` to your `$PATH` for access
+   to the `pyenv` command-line utility, enable shims to be able to use Pyenv.
```
julianfortune, line 209: Should there be an "and" here after the comma?

libexec/pyenv-init
```diff
@@ -91,21 +91,40 @@ function help_() {
...
     case "$shell" in
     fish )
       echo 'set -Ux PYENV_ROOT $HOME/.pyenv'
       echo 'set -Ux fish_user_paths $PYENV_ROOT/bin $fish_user_paths'
+      echo 'pyenv init --path | source'
       ;;
     * )
       echo 'export PYENV_ROOT="$HOME/.pyenv"'
       echo 'export PATH="$PYENV_ROOT/bin:$PATH"'
+      echo 'eval "$(pyenv init --path)"'
```
julianfortune, line 114: Thank you for grouping these lines together! Seeing two sections that both need to go in `.profile` was extremely confusing.

aiguofer:
I'm wondering if `pyenv init --path` is even needed at all? the installation is already asking users to add `export PATH="$PYENV_ROOT/bin:$PATH`, why not just ask them to add `export PATH="$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH"` (or the equivalent in each shell) instead?

I feel like the extra command to set the PATH is what's tripping people up as they think they need to replace the old command with that one.

libexec/pyenv-init
native-api: In fact, adding `shims` to `PATH` and the function to the shell have been presented as an optional step to offer a "bare minimal impact" installation option to help with Pyenv adoption -- so that prospective users aren't scared off by high upfront requirements and the fear that it would break their systems.

But now with `~/.profile` shenanigans added, two steps that involve `~/.profile` instead of one have made the instruction with the optional step too complicated.

Does the above still hold? Maybe it's time to stop worrying about that and lump all the parts together if practical?

README.md
```diff
@@ -205,15 +205,42 @@ easy to fork and contribute any changes back upstream.
...
+
+      - **If your `~/.profile` sources `~/.bashrc` (Ubuntu):**
+         <!--This way is an alternative option and needn't be replicated to `pyenv init`-->
+         ~~~ bash
+         echo -e 'if shopt -q login_shell; then' \
+               '\n  export PYENV_ROOT="$HOME/.pyenv"' \
+               '\n  export PATH="$PYENV_ROOT/bin:$PATH"' \
+               '\n eval "$(pyenv init --path)"' \
+               '\nfi' >> ~/.bashrc
+         ~~~
```
julianfortune, line 221: This `if` block solution is not working for me on Ubuntu 20.04 with bash. This is what I have in `.bashrc`:
```shell
if shopt -q login_shell; then
  export PATH="$PYENV_ROOT/bin:$PATH"
  eval "$(pyenv init --path)"
fi
```
If I echo my path I get:
```shell
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

agoose77: @julianfortune does your `.profile` contain a line that sources `.bashrc`?

julianfortune: @agoose77 Yes, [this](https://gist.github.com/julianfortune/27fc6218b86dfb55b5495045f93c07c1) is what is contained in .profile by default.

native-api: Okay, the `~/.bashrc` appending is not good enough, we need to come up with something better.
I'll split this off into a seprate PR and dissosiate this one from https://github.com/pyenv/pyenv/issues/1914.

julianfortune: What about a "C library ifndef"-style approach, something like:
```
if [ -z $PYENV_PATH_SET ]; then
  export PATH="$PYENV_ROOT/bin:$PATH" 
  eval "$(pyenv init --path)"
  export PYENV_PATH_SET='true'
fi

agoose77: @julianfortune yes, when you SSH with no commands you are given an interactive login shell, so your `.profile` is sourced (or other profile-like file) (see [here](https://web.archive.org/web/20121025183138/http://meta.ath0.com/2007/10/23/cleaning-up-bash-customizations/) or [here](http://www.solipsys.co.uk/new/BashInitialisationFiles.html) for more).

When you log in via the GUI, it varies per configuration. On Ubuntu with GDM only your `.profile` is explicitly sourced because `GDM` explicitly invokes `/bin/sh` (even if you change your default shell), and is launched as a login shell. 

So, for maximum compatibility, you end up wanting to have a `.bash_profile` that loads `.profile` and `.bashrc`, and then a much smaller `.profile` for just doing sh+bash compatible things.

native-api: The problem is, `~/.bashrc` in included when `~/.profile` is executed as `bash` (by a Bash login shell) -- but is _not_ included when it's run as `sh` (by a GUI manager).

My current ideas are:

1. Appends blocks to both `~/.bashrc` and `~/.profile` so that the first is executed in Bash mode and the 2nd one (below) in non-Bash mode:

    ```bash
    if [ -z "$BASH_VERSION" ]; then
      export PATH="$PYENV_ROOT/bin:$PATH" 
      eval "$(pyenv init --path)"
      export PYENV_PATH_SET='true'
    fi
    ```
    This has a downside of duplicated code. But this is the only solution achieveable with just appending.

2. Make an editing script (probably in Awk or Perl) that would insert the lines before the `if` block that sources `~/.bashrc`. Detect the necessary line with heuristics (e.g. a regex like `(^|\s)if\s.*\W(BASH_\w+|bash)\W`); if nothing found, append to the end.

    This will obviously break in environments with a very different `~/.profile` logic, and it'll be a rather complex command.

julianfortune:
> source pyenv init --path
> source pyenv init --shell 
> ```
> 
> What do you think?

@agoose77 This solution would work great for all of the use cases in my world! Just my 2 cents.

native-api:
@agoose77 @julianfortune @The-Judge 

> I feel that with all of this work on documenting exactly what to do for each shell, we're trying to solve the wrong problem; shell RC files are confusing, and not always consistent between shells and environments.
> <...>
> My rationale for dropping the per-shell documentation in `pyenv` itself is that it quickly explodes with the number of shells, OS default configurations (e.g. `.profile` sources `.bashrc` on Ubuntu), and login shells (e.g. I'm using GDM [which explicitly sources `~/.profile`](https://unix.stackexchange.com/a/360750) despite `SHELL=/usr/bin/zsh`). This gets _more_ confusing if users have non-standard configuration files already in use.

1. Not including ready-to-use invocations **would erect a high barrier to entry.** Even if they _can_ (which is also not the case quite often), users don't _want_ to have to think and waste time on it, they assume that if this is a production-ready software, someone has already done the thinking for them!
And rightfully so: if they have a common configuration, there _must_ be someone who already thought of a turn-key solution for it, so there's absolutely no reason to reinvent the wheel!
2. These "documenting exactly what to do for each shell" have been in the README for years and have been working fine so far.

**I think we can get the best of both worlds here if we include _both_ an explanation of what the effect should be, _and_ the ready-to-use examples for _common_ setups.**
This is what I just did in the update to this PR -- go see if it's to your satisfaction! :wink: 

> I think that we should consider dropping the shell-specific approach entirely, and instead `pyenv` should just expose two commands:
> 
> <..>
> pyenv init --path
> <...>
> pyenv init --shell

This is exactly what the interface is now (just with keeping the the old name for "`--shell`", see below).

> 
> These are the two kinds of configuration that `pyenv` needs to set AFAICT. The `pyenv init --shell` command is an attempt to remove the visual precedence that `pyenv init -` has over `pyenv init --path`, and to make it clear that one is for shell functions, the other is for environment.

I thought about renaming `-` to `--rc` -- both when writing #1898 and now -- and ultimately decided against it.

It's true that it'll make things _marginally_ clearer. But the drawbacks are:

* Should we force the users to rewrite their configurations _yet again,_ just after we've already frustrated them with this change and the broken and convoluted instructions?
    * That'll be very bad PR for the project. If we ever do that, we should rather first let the userbase calm down and forget the incident.
* Users don't deal with `pyenv init` _at all_ outside of the initial configuration step. So as long as it's clear what to put where, they don't really _care_ how these [incantations](http://www.catb.org/jargon/html/I/incantation.html) are spelled.
    * So all the frustration and the extra work that we'll be forcing onto all our users would be for nothing, really.
    * See also below for another reason why I _had_ to retain `-`. Concluding from it, the moment to rename it was during #1898, and now the opportunity is gone.

So my best idea for this -- if this is even worth doing, as per above -- is to introduce a new name as _an alias_ to `-`, and wait for a _long, long_ time before deprecating the original.
 
> Of course, this means that things will change for users. I don't know what the policy for API change is for this project, but if we don't want to break `pyenv init -` outright <...>

The change was made to fix https://github.com/pyenv/pyenv/issues/1649. We _needed_ to make a breaking change to force the users to change their configurations -- otherwise, they would keep falling victim to https://github.com/pyenv/pyenv/issues/1649 and would keep reporting it. _And_ we needed to retain `-` so that we could use the existing shell configuration to prominently notify them about what they need to do.

> Supporting all shells and advice on them seems to be more related to sidekick-projects like pyenv/pyenv-installer.

Actually, [`pyenv init` (without arguments) _is_ intended for Pyenv-installer](https://github.com/pyenv/pyenv-installer/blob/0a5ac813ba1ed660b201005f19be1e6a386d9ed5/bin/pyenv-installer#L61). Since the instructions are specific to Pyenv's version rather than Pyenv-Installer's, it's natuiral to keep them in Pyenv's codebase to avoid duplication.

agoose77:
@native-api thanks for the clarification. I agree on all points; I'm not pushing for any breaking changes any time soon. 

I like the `--path` `--shell` flags because they make it explicit that this is a two-phase initialisation. I agree that this shouldn't *break* anything; it should just be an alias with a long-term deprecation policy. Given that we've now made public changes to the pyenv initialisation APIs, I also concur that `pyenv init -` needs to remain stable.

aiguofer:
The updated instructions seem pretty complete to me. The only issue I can see is for anyone using a shell framework like prezto or oh-my-zsh. They generally have some plugin for `pyenv` which may do some, or all, of the initialization. For example, [prezto](https://github.com/sorin-ionescu/prezto/blob/master/modules/python/init.zsh) runs `eval "$(pyenv init - --no-rehash zsh)"`, while [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/pyenv/pyenv.plugin.zsh) attempts to do it all.

I imagine you don't need to go into detail about this, but maybe just a mention if you're using one of those frameworks you might only need the `PATH` changes (seems like that's all that would be needed for both of those frameworks).

shireenrao:
Thank you @aiguofer - even after following the changes mentioned, I was seeing issues. After seeing your comment on oh-my-zsh plugins I was able to resolve this by disabling the plugin.

native-api:
> For example, [prezto](https://github.com/sorin-ionescu/prezto/blob/master/modules/python/init.zsh) runs `eval "$(pyenv init - --no-rehash zsh)"`, while [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/pyenv/pyenv.plugin.zsh) attempts to do it all.

Those projects need to react to Pyenv changes.

native-api:
> I imagine you don't need to go into detail about this, but maybe just a mention if you're using one of those frameworks you might only need the `PATH` changes (seems like that's all that would be needed for both of those frameworks).

I feel that it's rather the responsibility of those projects to inform their users which exact configuration steps they're saving them from.

native-api:
> I imagine you don't need to go into detail about this, but maybe just a mention if you're using one of those frameworks you might only need the `PATH` changes (seems like that's all that would be needed for both of those frameworks).

I feel that it's rather the responsibility of those projects to inform their users which configuration steps they're saving them from.

aiguofer:
> > I imagine you don't need to go into detail about this, but maybe just a mention if you're using one of those frameworks you might only need the `PATH` changes (seems like that's all that would be needed for both of those frameworks).
> 
> I feel that it's rather the responsibility of those projects to inform their users which configuration steps they're saving them from.

Yeah that makes sense... Maybe just a mention to check how their framework handles `pyenv` then. I feel like a lot of users might unfortunately not be aware that their framework might already be doing something and thus think it's an issue with `pyenv` itself.

PR #1922, opened 2021-05-13
Bash automated install option

native-api:
See https://github.com/native-api/pyenv/tree/bash_automated_option#basic-github-checkout for the rendered README
(search that page for "automated installation" to quickly find the section).

---

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/1914

### Description

Adds an optional way for Bash fit for an automated installation

Based on https://github.com/pyenv/pyenv/pull/1920, needs to be merged after it

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,21 @@ easy to fork and contribute any changes back upstream.
             eval "$(pyenv init --path)"
             ~~~
 
+            <!--This is an alternative option and needn't be replicated to `pyenv init`-->
+            Alternatively, for an automated installation, you can run the following:
+            ~~~ bash
+            echo -e 'if shopt -q login_shell; then' \
+                  '\n  export PYENV_ROOT="$HOME/.pyenv"' \
+                  '\n  export PATH="$PYENV_ROOT/bin:$PATH"' \
+                  '\n eval "$(pyenv init --path)"' \
+                  '\nfi' >> ~/.bashrc
+            echo -e 'if [ -z "$BASH_VERSION" ]; then'\
+                  '\n  export PYENV_ROOT="$HOME/.pyenv"'\
+                  '\n  export PATH="$PYENV_ROOT/bin:$PATH"'\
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -174,95 +174,151 @@ For pyenv to install python correctly you should [**install the Python build dep
...
+            ~~~bash
+            export PYENV_ROOT="$HOME/.pyenv"
+            export PATH="$PYENV_ROOT/bin:$PATH"
+            eval "$(pyenv init --path)"
+            ~~~
+
+            <!--This is an alternative option and needn't be replicated to `pyenv init`-->
+            Alternatively, for an automated installation, you can run the following:
+            ~~~ bash
+            echo -e 'if shopt -q login_shell; then' \
```
agoose77, line 248: @native-api I'm wondering if it might be easier / more robust to just check for either a guard variable e.g. `[ -z "$PYENV_ROOT" ]` or test for `pyenv` with `! command -v pyenv &> /dev/null` instead of making assumptions about *why* a particular RC file has been sourced?

native-api: That's not an "assumption", that's actually a requirement :-) See the guidance text for this section in the README when this code needs to run.

> if it might be easier / more robust to just check

There's no such foolproof test. E.g. `command -v pyenv` will always succeed if `pyenv` is installed into a location that's already in `PATH` (e.g. with Homebrew).

The solution to #1649 that we chose is to conform to the UNIX design: make shell configuration commands "dumb" and rely on the shell configuration infrastructure to run them at the right moments.
The only alternative was to second-guess the user -- and thus inevitably guess wrong from time to time.

native-api: Actually, if you have a problem understanding how this chunk achieves the goal stated in the guidance text -- maybe I should add a brief explanation text?

agoose77: @native-api no, I'm not confused; I didn't see that you scoped this section to the distros that source bashrc from profile. With that context, I think this is as general + safe as we can be here.

Issue #1877, opened 2021-04-17
pyenv arm64 builds confused by x84 gettext in /usr/local


mikepqr:
I'm trying to follow homebrew's advice and supported workflow to configure a system with both arm64 and x86_64 brew packages, i.e. put an arm64 brew in /opt/homebrew and an x86 brew in /usr/local.

Both have a copy of the gettext package (which is a dependency of ... well ... pretty much everything).

The arm64 brew has pyenv. `pyenv install 3.9.4` fails with:

```
[...]

PR #1957, opened 2021-06-01
Append Homebrew custom prefix to search path

native-api:
Needed to find other Python deps (e.g. libintl) in Homebrew if it has
nonstandard prefix (e.g. Apple ARM64)

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/1877

### Description

Apple M1 install Homebrew into `/opt/homebrew`.
If x64 Homebrew is installed as well, ARM64 Python build looks there for libs fisrt and tries to link with those for the other architecture, resulting in error.
Append paths so as not to override any user-supplied paths.

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -766,6 +766,7 @@ build_package_standard_build() {
   local PACKAGE_CFLAGS="${package_var_name}_CFLAGS"
 
   if [ "$package_var_name" = "PYTHON" ]; then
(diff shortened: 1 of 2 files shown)


native-api:
I'm getting "`unstub <cmd>` failed" errors and there's no info to debug them; the bats version in the codebase also cannot do debug printing (to fd 3).

Looks like I'll have to refirbish the test framework before I'm able to progress with this.

PR #1960, opened 2021-06-01
Fixes README.md guide for fish shell

bytefluxio:

### Prerequisite
  **considered, but not applicable**
* [x] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project. (**not applicable**)
  - None - Thought it would be faster to just do the PR instead of mentioning the issue first

### Description

Tried to follow the setup instructions using fish, didn't work, so I went through the init script and noticed that with the parameter1 set to the shell in use, it used the fish commands instead of bash. That made it work. So I'm adjusting the README.md to make it easier for other fish users to get up and running with pyenv

### Tests
**none**

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ easy to fork and contribute any changes back upstream.
 
         And add this to `~/.config/fish/config.fish`:
         ~~~ fish
-        status is-login; and pyenv init --path | source
+        status is-login; and pyenv init --path fish | source
         ~~~
 
         If Fish is not your login shell, also follow the Bash/Zsh instructions to add to `~/.profile`.
@@ -309,7 +309,7 @@ easy to fork and contribute any changes back upstream.
       - For **Fish shell**:
         Add this to `~/.config/fish/config.fish`:
         ~~~ fish
-        pyenv init - | source
+        pyenv init - fish | source
         ~~~
 
       **General warning**: There are some systems where the `BASH_ENV` variable is configured


README.md
```diff
@@ -276,7 +276,7 @@ easy to fork and contribute any changes back upstream.
 
         And add this to `~/.config/fish/config.fish`:
         ~~~ fish
-        status is-login; and pyenv init --path | source
+        status is-login; and pyenv init --path fish | source
```
native-api, line 279: This makes no difference -- since this file can only be run with Fish!

native-api:
`pyenv init --path` autodetects the shell that it runs in and produces the appropriate shell commands. So the proposed change looks redundant -- and also exposes an internal interface (the optional `<shell>` argument) which we'd rather not include into official documentation so that we can change it freely.

Did autodetrection fail for you somehow?

bytefluxio:
I guess it did. Let me check some things and get back to you. My guess is omf pyenf might have put functions somewhere that weren't removed when I uninstalled it. So it does make sense to file a bug issue before doing a report I guess. If it's the auto detect that failed. I'll try to find out why and make a new PR for that.

bytefluxio:
**This is just an update post, no action needs to be taken. This PR will be linked to a bug issue I will create later. Not sure if today though**

```
 ~ $ export PATH="/c/Users/rjwes/.pyenv/shims:${PATH}u^C
 ~ $ status is-login; and pyenv init --path | source
- (line 1): Variables cannot be bracketed. In fish, please use "$PATH".
export PATH="/c/Users/rjwes/.pyenv/shims:${PATH}"
                                          ^
from sourcing file -
source: Error while reading file '<stdin>'
 ✘  ~ $ type -a pyenv
pyenv is /c/Users/rjwes/.pyenv/libexec/pyenv
 ~ $
 ```
 
 there was indeed a function defined in .pyenv/libexec/pyenv.init that fish sourced before. Didn't change anything though.
 So pyenv seems to not recognize that im running fish shell. I'll try to reproduce it.

native-api:
Autodetection checks the name of parent process: https://github.com/pyenv/pyenv/blob/e2125b27501888d4163aa526fbbc69c0c7de5082/libexec/pyenv-init#L42-L47

Looks like you're running Fish under a different name.

bytefluxio:
seems like my ps doesn't have an -o option.

Doesn't seem a pyenv bug, but a windows one according to [this slightly unrelated site](https://www.tutorialfor.com/questions-318000.htm)

There seem to be other options than using the `-o` option to extract the information.

@native-api: Would you accept a pr that changes this so that -o isn't needed or would you just wait for microsoft to fix their error? :)



mintty_ENhtqdk0Zb

mintty_I7cq6rwPzf

native-api:
> @native-api: Would you accept a pr that changes this so that -o isn't needed or would you just wait for microsoft to fix their error? :)

From your screenshots, this seems to be a limitation of Cygwin/MinGW rather than Windows.
We don't officially support Windows, see https://github.com/pyenv/pyenv#homebrew-on-macos.

If you can figure out a reliable way to work with `ps` without `-o`, I can accept a PR -- but you're very likely to see breakages elsewhere.

PR #1962, opened 2021-06-01
Windows 10 recognize shell

bytefluxio:
As discussed in [this pullrequest](https://github.com/pyenv/pyenv/pull/1960), pyenv can't correctly identify what shell it is running on windows 10, since the `ps -o` option is missing there.

This fix uses `awk` instead of the `-o` option. We could also check if the `-o` option worked (check if `$shell` is empty before using awk fallback), if you would want that instead of the current version, just tell me :)

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -39,7 +39,7 @@ done
 
 shell="$1"
 if [ -z "$shell" ]; then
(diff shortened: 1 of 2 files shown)


bytefluxio:
ah dang another commit sneaked in. let me fix the pr quickly

PR #1963, opened 2021-06-01
Fixes recognizing shell on windows10

bytefluxio:
As discussed in [this pullrequest](https://github.com/pyenv/pyenv/pull/1960), pyenv can't correctly identify what shell it is running on windows 10, since the `ps -o` option is missing there.

This fix uses `awk` instead of the `-o` option. We could also check if the `-o` option worked (check if `$shell` is empty before using awk fallback), if you would want that instead of the current version, just tell me :)

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -39,7 +39,7 @@ done
 
 shell="$1"
 if [ -z "$shell" ]; then
-  shell="$(ps -p "$PPID" -o 'args=' 2>/dev/null || true)"
+  shell="$(ps -p $PPID 2>/dev/null | awk 'END {print $NF}' 2>/dev/null)"
   shell="${shell%% *}"
   shell="${shell##-}"
   shell="${shell:-$SHELL}"
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -39,7 +39,10 @@ done
 
 shell="$1"
 if [ -z "$shell" ]; then
-  shell="$(ps -p $PPID 2>/dev/null | awk 'END {print $NF}' 2>/dev/null)"
+  shell="$(ps -p "$PPID" -o 'args=' 2>/dev/null || true)"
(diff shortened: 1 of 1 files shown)


bytefluxio:
> If you can figure out a reliable way to work with ps without -o, I can accept a PR -- but you're very likely to see breakages elsewhere.

Sounds like the fallback option would be better

native-api (changes requested):
See below.

This is why we don't support Windows. Too many things needs to be done differently and too many additional limitations.

Besides, alternate Pythons you install will not be "real" Win32 Pythons but rather run through a compatibility layer -- so you will lose Windows-specific functionality.

libexec/pyenv-init
```diff
@@ -40,6 +40,9 @@ done
 shell="$1"
 if [ -z "$shell" ]; then
   shell="$(ps -p "$PPID" -o 'args=' 2>/dev/null || true)"
+  if [ -z "$shell" ]; then
+    shell="$(ps -p $PPID 2>/dev/null | awk 'END {print $NF}' 2>/dev/null)"
```
native-api, line 44: * Will terminate Pyenv if the invocation fail for whatever reason
* Will silently produce "COMMAND" if `ps` doesn't print the info for whatever reason (e.g. process is not found or cannot be opened).

bytefluxio: Alright. I see it. I'll hopefully be back on a linux distribution soon, instead of having to get used to windows. Thanks for taking your time to answer me, though.

When I'm back on Linux, I'll see if there are any open issues, I could address here. Overall I'm new to contributing to open source and this experience has been very pleasant. Keep up the good work.

native-api: The docs recommend using https://github.com/pyenv-win/pyenv-win for Windows. It does installed "real" win32 Pythons.

PR #1977, opened 2021-06-08
Update setup instructions for debian, ubuntu etc.

LukeAI:
### Prerequisite
* [n/a ] Please consider implementing the feature as a hook script or plugin as a first step.
* [ n/a] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
* [x ] My PR addresses the following pyenv issue (if any)
https://github.com/pyenv/pyenv/discussions/1974#discussioncomment-839302

### Description

Edit setup instructions to avoid problem in ubuntu with shim not always being at front of `$PATH` if pip and other python packages installed in `~/.local/bin` (see link)

### Tests
- [n/a] My PR adds the following unit tests (if any)

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,20 @@ easy to fork and contribute any changes back upstream.
          echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile
          echo 'eval "$(pyenv init --path)"' >> ~/.profile
          ~~~
+            <!--This is an alternative option and needn't be replicated to `pyenv init`-->
+         Alternatively, for an automated installation, you can run the following:
+         ~~~ bash
+         echo -e 'if shopt -q login_shell; then' \
+                 '\n  export PYENV_ROOT="$HOME/.pyenv"' \
+                 '\n  export PATH="$PYENV_ROOT/bin:$PATH"' \
+                 '\n eval "$(pyenv init --path)"' \
+                 '\nfi' >> ~/.bashrc
+         echo -e 'if [ -z "$BASH_VERSION" ]; then'\
+                 '\n  export PYENV_ROOT="$HOME/.pyenv"'\
+                 '\n  export PATH="$PYENV_ROOT/bin:$PATH"'\
+                 '\n  eval "$(pyenv init --path)"'\
+                 '\nfi' >>~/.profile
+         ~~~
 
          - **If your `~/.profile` sources `~/.bashrc` (Debian, Ubuntu, Mint):**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,6 +232,20 @@ easy to fork and contribute any changes back upstream.
          echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.profile
          echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile
          echo 'eval "$(pyenv init --path)"' >> ~/.profile
+         ~~~
+            <!--This is an alternative option and needn't be replicated to `pyenv init`-->
+         Alternatively, for an automated installation, you can run the following:
```
native-api, line 237: This alternative is intended for Debian/Ubuntu case only, it won't work in other cases AFAIK. Is there a reason why is was moved?

README.md
```diff
@@ -240,22 +254,10 @@ easy to fork and contribute any changes back upstream.
...
-                  '\n eval "$(pyenv init --path)"' \
-                  '\nfi' >> ~/.bashrc
-            echo -e 'if [ -z "$BASH_VERSION" ]; then'\
-                  '\n  export PYENV_ROOT="$HOME/.pyenv"'\
-                  '\n  export PATH="$PYENV_ROOT/bin:$PATH"'\
-                  '\n  eval "$(pyenv init --path)"'\
-                  '\nfi' >>~/.profile
+            And put this line at the _bottom_ of `~/.profile` (_after_ sourcing `~/.bashrc` and any other `$PATH` manipulations):
+            ~~~bash
+            eval "$(pyenv init --path)"
```
native-api, line 260: Since it's at the bottom, it'll probably be clearer to just give an appending command without such long-winded explanations like in other cases.

README.md
LukeAI: I moved it because this won't work on Ubuntu, since appending the first two lines will mean that it doesn't come before ~/.bashrc is sourced. I figured it was for other OS's - but if that's not the case, maybe it should just be deleted?

README.md
LukeAI: ok, happy to do that

README.md
native-api: It was specifically designed for Ubuntu since it's not possible to reliably insert stuff into the middle of other unknown code -- and should work there.

LukeAI: but this will insert these lines at the end of the file? so they won't be inserted before ~/.bashrc is sourced which is required?

native-api: This solution gets around that requirement. If `~/.profile` is sourced as `bash`, the chunk in `~/.bashrc` will run -- still before `eval "$(pyenv init -)"`. If it's executed as `sh`, `~/.bashrc` won't be sourced and the chunk in `~/.profile` will run instead.

Issue #1980, opened 2021-06-10
Request: Update mtime of shim when installing a newer Python


cjolowicz:
Would it be possible for pyenv to update the mtime of a shim when a newer Python is installed?

Due to the caching logic in virtualenv >= 20, virtualenv currently uses an older pyenv-installed Python to generate environments even when a newer one is installed and activated (https://github.com/pypa/virtualenv/issues/2130). The reason for this appears to be that the mtime of the relevant shim does not change.

It seems to be a relatively straightforward change to teach `pyenv rehash` to copy the newest mtime to the shim. IIUC rehash is triggered automatically after an install. So this would make virtualenv's caching work out of the box with pyenv. I'll send a PR shortly. **Update:** See #1981

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Linux (debian:stable-slim image), Mac
- [x] OS architecture (e.g. amd64): amd64
- [x] pyenv version: 3.9.5
- [x] Python version: 3.7.3
- [x] C Compiler information (e.g. gcc 7.3): not relevant
- [x] Please attach verbose build log as gist : not relevant
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

native-api:
Could you link to and walk us through the corresponding Virtualenv logic that you're referring to?
To make sure and support your reasoning that the mtimes are used by it as you claim they are.

[...]

PR #1981, opened 2021-06-10
Update mtime of shims with their targets

cjolowicz:
### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/1980

### Description

This PR adds a step at the end of pyenv-rehash to update mtimes of shims, where appropriate:

1. List the full paths of all executables under `$PYENV_ROOT/versions/*/bin`.
2. For each of these executables:
3. Determine the full path of the corresponding shim (`$PYENV_ROOT/shims/$basename`)
4. If the executable is newer than the shim, copy its timestamp to the shim.

Step 4 uses bash's builtin `test -nt` to compare mtimes, and `touch -r` to copy the mtime. Note that `touch -r` is specified by POSIX. I have explicitly checked support for `touch -r` on Linux (GNU coreutils), FreeBSD, and MacOS.

### Tests
  - Add test that rehash updates mtime

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -110,6 +110,20 @@ make_shims() {
   done
 }
 
+copy_mtime_for_shims() {
+  local version file shim
+  pyenv-versions --bare --skip-aliases | \
+  while read -r version; do
+    for file in "${PYENV_ROOT}/versions/${version}/bin/"*; do
+      shim="${PYENV_ROOT}/shims/${file##*/}"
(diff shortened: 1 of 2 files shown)


native-api (changes requested):
As per https://github.com/pyenv/pyenv/issues/1980#issuecomment-858574638, it appears that you can't reliably fix the problem the way this PR suggests. Virtualenv will instead need to be fixed to check `pythonX.Y`'s version instead of/in addition to its mtime.

Issue #2006, opened 2021-07-12
python-build overwrites some MAKE_OPTS given as environment variable (CFLAGS_EXTRA) for micropython


yggdr:
### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04):
FreeBSD 13.0-p5
- [x] OS architecture (e.g. amd64):
amd64
- [x] pyenv version:
2.0.3
- [x] Python version:
micropython (all)
- [x] C Compiler information (e.g. gcc 7.3): 
gcc 10.3.0
- [ ] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

Due to an [issue with compiling micropython > 1.12 on FreeBSD](https://github.com/micropython/micropython/issues/7525) with `asdf` (which uses `pyenv`/`python-build` under the hood), I needed to give `make` an additional `CFLAGS_EXTRA` argument, like so:
`env MAKE=gmake MAKE_OPTS="CFLAGS_EXTRA=-Wno-float-conversion V=1" MAKE_INSTALL_OPTS="CFLAGS_EXTRA=-Wno-float-conversion V=1" ./python-build micropython-1.13 /tmp/TESTPY`
That however doesn't work, because line 893 overwrites `CFLAGS_EXTRA` again:
[...]

PR #2007, opened 2021-07-13
WIP Fix for #2006

yggdr:
WIP CFLAGS_EXTRA should now work for micropython. bats test currently not correct. Please see issue #2006 for details.


yggdr:
@native-api The fix you propose is not working.

The unittest for the function `build_package_micropython` now seemingly does the right thing, but only if the `CFLAGS_EXTRA` argument is passed (like I did in the test). With my `+=` version, that was indeed the correct test, but with your new version, it is no longer. That is because the program itself doesn't get `CFLAGS_EXTRA` passed, but `MAKE_OPTS`, which is a string containing the new `CFLAGS_EXTRA` definition. So we would need to operate that out of `MAKE_OPTS` to make your code work, which is what I wanted to avoid with my `+=` "hack".

The `+=` syntax is definitely correct syntax _within_ a Makefile, and although the GNU Make docs don't mention it, it seems to also work on the command line for overriding variables. I have also tested this. Do you have docs that state it is not legal syntax on the command line?

But now I see how `stub` is meant to work, thanks :)

native-api:
I made the code prepend stuff rather than append it so that user's input take priority in case of conflicts (I didn't check if the code assumes this chunk to prevail later on though; if it does, the chunk should indeed be appended).



> Do you have docs that state it is not legal syntax on the command line?

I checked [GNU Make command line docs](https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding) which only mention `=` and `:=`/`::=` as valid syntax. Even if it works, using undocumented syntax is risky without additional guarantees 'cuz it can break unpredictably and you don't know the details of how it's supposed to work (so if it breaks, you don't know what to do and who's at fault).

native-api:
Any progress on this? That is, if you wish to see it merged, of course.

PR #2020, opened 2021-08-01
Disable coreutils on M1 Apple Silicon with arm64

ttksm:

### Prerequisite
  * **[Author comment] I don't patch to rbenv.**
  - https://github.com/pyenv/pyenv/issues/1768#issuecomment-757552526
    - This issue has already closed but this patch will automate above the commented operation.

### Description
- Commands of coreutils installed with homebrew can negatively affect installation, which `pyenv install` on M1 Apple Silicon with arm64.
- Therefore, the patch temporarily excludes coreutils bin from PATH so that `pyenv install` does not reference the commands of coreutils.
  - [coreutils](https://www.gnu.org/software/coreutils/)
    - The path of coreutils bin, which is `$(brew --prefix coreutils)/libexec/gnubin`, overrides native path of commands included in coreutils.
    - The overriding allows direct access to commands without using `g*`.
- I don't think this patch satisfies best practices of high maintainability codes because blacklist method lacks versatility.
  - For the time being, I propose a patch as one of the solutions.

### Problem
- [ ] Platform information (e.g. Ubuntu Linux 16.04): macOS Big Sur 11.4
- [ ] OS architecture (e.g. amd64): arm64
- [ ] pyenv version: 2.0.4-1-gb005dcc7
- [ ] Python version: 3.9.6
- [ ] C Compiler information (e.g. gcc 7.3): I have tried the following two.
  - [ ] Apple clang version 12.0.5
  - [ ] Homebrew GCC 11.2.0
- [ ] Please attach verbose build log as gist: https://gist.github.com/ttksm/f287528cb664cc4bd8e1bbe2ea5d6dfe
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

### Tests


native-api:
Could you please showcase the problem that this is intended to fix -- with diagnostic info as per [the issue template](https://github.com/pyenv/pyenv/blob/master/.github/ISSUE_TEMPLATE.md)?

ttksm:
@native-api 
Thank you for your comment.
I've added information such as the environment in which the error occurred.
Please take a look when you get a chance.

native-api:
Thank you.

I see that the problem is in `./config.guess` outputting a result that Python's `./configure` is not able to work with.
I don't have Apple M1 hardware, so could you run the folowiing for me?
* The output of `/usr/bin/uname -a`, `guname -a`
* The output of `PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }' sh -x config.guess` in the Python build tree (it should be somewhere under your `$TEMP` after a failed build; if not, download and unpack the release tarball yourself and run the command there).

I also see that [the release 3.9.3 has been recalled due to having introduced an unexpected incompatibility](https://www.python.org/downloads/release/python-393/) so we should probably remove it, too.

* Could you check if other releases that officially support Apople Silicon -- 3.9.1+ and 3.8.10+ -- show the same problem?

ttksm:
Thank you.

>The output of /usr/bin/uname -a, guname -a

The results are as follows.

- `/usr/bin/uname -a`
```
Darwin BLMC009.local 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:27 PDT 2021; root:xnu-7195.141.2~5/RELEASE_ARM64_T8101 arm64
```
- `guname -a`
```
Darwin BLMC009.local 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:27 PDT 2021; root:xnu-7195.141.2~5/RELEASE_ARM64_T8101 arm64 arm64 MacBookPro17,1 Darwin
```

>The output of PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }' sh -x config.guess in the Python build tree

I uploaded log to [my gist](https://gist.github.com/ttksm/f287528cb664cc4bd8e1bbe2ea5d6dfe#file-debug_config-guess-log).

>Could you check if all other releases that officially support Apple Silicon -- 3.9.1+ and 3.8.10+ -- show the same problem?

I tried to install the following versions but it failed.
3.8.10, 3.8.11, 3.9.1, 3.9.2, 3.9.4, 3.9.5 and 3.9.6.

native-api:
Okay. The core reason is an old `config.sub` (a stock file inserted by `autoconf`) that doesn't support `arm64`. The issue has been reported in https://bugs.python.org/issue43878 and fixed in 3.10 only.

So we can selectively apply this workaroud to 3.9.1+ and 3.8.10+ only.

native-api (changes requested):
Could you change the logic to only selectively apply to the affected versions in the affected scenarios?
E.g. call this fuction not from general logic but from specific formulae.

native-api:
Another option is to instead patch the `config.sub` for affected versions to a newer version of it.
The patch will need to be duplicated for each affected formula though.

ttksm:
Thanks for your investigation and advices.
I changed the logic to only selectively apply to the affected versions (3.8.10+ and 3.9.1+ ~ 3.9.6).
Please re-review.

native-api (approved):

native-api:
I've made some refinements. Please check that all the changed scripts work on an Apple M1.

ttksm:
Thanks for your refinemnts.
I've checked that scripts work fine with revision https://github.com/ttksm/pyenv/commit/ff9720981852fbcedbe5fbf380612284a01da049.

PR #2021, opened 2021-08-02
Update install instructions for shell variables

tonywang10101:
* modified commands for bash, zsh, fish shell installation instructions


joshfriend:
Hey, thanks so much for this contribution and trying to make the install docs better! 

`$PYENV_ROOT/bin` is actually the only thing users need to add to their `$PATH` to set up pyenv. The `bin` directory is actually where the `pyenv` command itself lives, so you need that in the `PATH` to start. Calling `pyenv init --path` then emits a directive that adds the shims path to the `PATH`, e.g. `export PATH="/Users/josh/.pyenv/shims:${PATH}"`.

PR #2022, opened 2021-08-03
Remove 3.9.3

native-api:
It has been recalled due to introducing an unexpected incompatibility:
https://www.python.org/downloads/release/python-393/

### Prerequisite

### Description

### Tests


ulgens:
I didn't really understand why this is removed. It's been "recalled", okay, but still there might be people around explicitly install that broken version and try/do things. IMO, it should be kept, for historical reasons at least.

native-api:
@ulgens PSF strongly discourages people from using this versions because it causes trouble due to having introduced a binary incompatibility. They even removed a download for it from their download page.

If we kept it, we'd be contributing to that trouble.

ulgens:
@native-api Got it, thanks a lot.

PR #2045, opened 2021-09-02
Prevent addition of duplicate plugin dirs to PATH

PamelaM:
### Description
- [x ] Here are some details about my PR

`libexec/pyenv` adds the `.../plugins/*/bin` directories to the PATH. This leads to duplicated PATH entries when `bin_path` is the same as `PYENV_ROOT`.

This PR simply checks if `bin_path` != `PYENV_ROOT` before adding the bin_path `plugins/*/bin` directories to PATH

### Tests

### Prerequisite
  * _This is a trivial bug in libexe/pyenv and, therefore, not amenable to either a hook script or a plugin_
  *  _I checked rbenv and this code is specific to a difference between rbenv and pyenv, specifically: pyenv adds the bin_path plugins/*/bin directories, while rbenv does not_

  - _No existing pyenv issue_

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -77,9 +77,12 @@ export PYENV_DIR
 shopt -s nullglob
 
 bin_path="$(abs_dirname "$0")"
-for plugin_bin in "${bin_path%/*}"/plugins/*/bin; do
-  PATH="${plugin_bin}:${PATH}"
-done
+if [ "${bin_path%/*}" != "$PYENV_ROOT" ]; then
+  # Add the plugins paths for bin_path if it isn't in $PYENV_ROOT
+  for plugin_bin in "${bin_path%/*}"/plugins/*/bin; do
+    PATH="${plugin_bin}:${PATH}"
+  done
+fi
 for plugin_bin in "${PYENV_ROOT}"/plugins/*/bin; do
   PATH="${plugin_bin}:${PATH}"
 done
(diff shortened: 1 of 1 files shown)


native-api:
> when `bin_path` is the same as `PYENV_ROOT`

How can this happen?
The only way I see is if you add `pyenv/libexec` to `PATH` instead of `pyenv/bin`. That is not a supported configuration.

native-api:
> when `bin_path` is the same as `PYENV_ROOT`

How can this happen?
`bin_path` (the location of `pyenv` executable) is supposed to be either `$PYENV_ROOT/bin` (git checkout installation), or outside `$PYENV_ROOT` (Homebrew or other package manager installation).

PamelaM:
> > when `bin_path` is the same as `PYENV_ROOT`
> 
> How can this happen?
> `bin_path` (the location of the `pyenv` executable) is supposed to be either `$PYENV_ROOT/bin` (git checkout installation), or outside `$PYENV_ROOT` (Homebrew or other package manager installation).

This happened in a Docker container that pre-installed pyenv and set `ENV PYENV_ROOT=...`

Now that I think about it from outside my tunnel vision, I can see that that's not an good plan, and will have a conversation with that Dockerfile maintainer.

I think the fix is still "safe", but it does add complexity, so it's your call whether to include this or not.

native-api (approved):
Since the code performs the same operation twice for 2 paths, I guess it makes sense to check if they are equal.

PR #2060, opened 2021-09-11
macOS: avoid using binutils from Homebrew

puhitaku:

### Prerequisite
   - #2020 

### Related
 - https://github.com/bitcoin/bitcoin/issues/20825#issuecomment-753444519
    - Conversation of the same error and solution

### Description
Disabling coreutils is enough to get Python 3.8.10 ~ 3.8-dev and 3.9.1 ~ 3.9-dev to work on macOS. Disabling binutils from Homebrew was also necessary to build Python 2.7.18 on my Mac + Apple Silicon. Actually, this PR is a part of a fix for Python 2.7.18 + Apple Silicon build failure.

I've enhanced the (very fresh) function that was introduced in #2020 to disable binutils in Homebrew. Thanks!

#### Build confirmation after the change

Here is [a Gist](https://gist.github.com/puhitaku/81b6b0930948fc8e4811a9f850e475fe) that has all verbose build logs of:
 - 3.8-dev
 - 3.8.10
 - 3.8.11
 - 3.9-dev
 - 3.9.1
 - 3.9.2
 - 3.9.4
 - 3.9.5
 - 3.9.6
 
 #### Environment
 - MacBook Pro (Late 2020, M1)
 - macOS 11.5.2 (Big Sur)

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1526,23 +1526,26 @@ use_xcode_sdk_zlib() {
 # CPython 3.9.1+ and 3.8.10+ have old config.sub that doesn't support "arm64" arch
 # which is what Gnu uname (but not Apple uname) produces on Apple M1
 # (https://github.com/pyenv/pyenv/pull/2020#issuecomment-891911842)
-build_package_arm64_apple_disable_homebrew_coreutils() {
+# Binutils from Homebrew also affects the build and causes linker error
+build_package_arm64_apple_disable_homebrew_executables() {
   if is_arm64_apple; then
-    local brew_coreutils="$(brew --prefix coreutils 2>/dev/null || true)"
-    if [ -d "$brew_coreutils" ]; then
-      local after_list=()
-      # Bash 3 ignores IFS set for command when doing expansions so have to set it separately
-      IFS=':'; local -a list=(${PATH}); IFS="$OLDIFS"
-      for str in "${list[@]}"; do
-        if [[ "$str" != "$brew_coreutils"/* ]]; then
-          after_list+=("$str")
-        else
-          echo "python-build: excluding \`$str' from PATH"
-        fi
-      done
-      # Bash 3 ignores IFS set for command when doing expansions so have to set it separately
-      IFS=':'; export PATH="${after_list[*]}"; IFS="$OLDIFS"
-    fi
+    for pkg in coreutils binutils; do
+      local brew_executable="$(brew --prefix $pkg 2>/dev/null || true)"
+      if [ -d "$brew_executable" ]; then
+        local after_list=()
+        # Bash 3 ignores IFS set for command when doing expansions so have to set it separately
+        IFS=':'; local -a list=(${PATH}); IFS="$OLDIFS"
+        for str in "${list[@]}"; do
+          if [[ "$str" != "$brew_executable"/* ]]; then
+            after_list+=("$str")
+          else
+            echo "python-build: excluding \`$str' from PATH"
+          fi
+        done
+        # Bash 3 ignores IFS set for command when doing expansions so have to set it separately
+        IFS=':'; export PATH="${after_list[*]}"; IFS="$OLDIFS"
+      fi
+    done
   fi
 }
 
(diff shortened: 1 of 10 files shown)


native-api:
Could you demonstrate the probem with 2.7.18 and how Homebrew's binutils are at fault?

If they really are, we should, like in https://github.com/pyenv/pyenv/pull/2020, only disable them for affected versions, separately from coreutils -- i.e. not change the build environment more than necessary for each version.

native-api:
IIRC 2.7 is not supported on MacOS 11 at all.

native-api:
IIRC 2.7 is not supported on MacOS 11 at all. Does it even build in the stock environment?

puhitaku:
> Could you demonstrate the probem with 2.7.18 and how Homebrew's binutils are at fault?

I've created [a Gist](https://gist.github.com/puhitaku/de8667dd61d8246c8e2cb707723d79fe) for ensuring that both coreutils and binutils have to be disabled. The script is also attached.

Summary:
At pyenv 9bd87fa (master + 2.7.18 patch),
 - With all (coreutils and binutils are enabled): reproduced #2020 
 - Without binutils: reproduced #2020 
 - Without coreutils: ld warning and undefined symbol error that look so similar to https://github.com/bitcoin/bitcoin/issues/20825
 - Without all: build successfully

> If they really are, we should, like in #2020, only disable them for affected versions, separately from coreutils

I'll modify the code to disable binutils only for 2.7.18 to limit the effect of this PR. Applying this to other supported Python should be done in another PR I think.

FYI: [A brief test](https://gist.github.com/puhitaku/064d2cc443d13e88dcdf5040be332014) shows that disabling Homebrew binutils was also necessary for 3.8-dev

puhitaku:
I've pushed 929d066. I re-modified the code rather than reverting it back. If I should squash them, please tell me 😉

puhitaku:
> IIRC 2.7 is not supported on MacOS 11 at all. Does it even build in the stock environment?
It builds on macOS 11.5.2 with only a few lines of patch.

```patch
From 38289ec101bf6bff87e9b76de75928055aedd06c Mon Sep 17 00:00:00 2001
From: Takumi Sueda <puhitaku@gmail.com>
Date: Sat, 11 Sep 2021 14:06:11 +0900
Subject: [PATCH] Make configure recognize arm64 environment

---
 configure | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/configure b/configure
index 63d675312d..c40ec120de 100755
--- a/configure
+++ b/configure
@@ -8474,6 +8474,9 @@ fi
     	ppc)
     		MACOSX_DEFAULT_ARCH="ppc64"
     		;;
+    	arm64)
+    		MACOSX_DEFAULT_ARCH="arm64"
+    		;;
     	*)
     		as_fn_error $? "Unexpected output of 'arch' on OSX" "$LINENO" 5
     		;;
-- 
2.30.1
```

It comes without some internal things that I patched in #2061 

```
Failed to build these modules:
_ctypes            _Qt                _tkinter        
```

native-api:
> Without coreutils: ld warning and undefined symbol error that look so similar to Error Building from source on MacOS BigSur; linker issues bitcoin/bitcoin#20825

Please do not replace facts with your deductions (you can accompany them with those though). "Looks very similar" is not good enough to say with confidence that the specific configuration issue is the same and the best way to fix it is the same.

In particular, it also "looks very similar" to https://github.com/pyenv/pyenv/issues/1877.

Moreover, if binutils unconditionally breaks Apple's toolchain, I suspect we don't even need to support the case where they are on `PATH` -- the user then has a broken setup.

plugins/python-build/bin/python-build
```diff
@@ -1523,27 +1523,38 @@ use_xcode_sdk_zlib() {
...
-          after_list+=("$str")
-        else
-          echo "python-build: excluding \`$str' from PATH"
-        fi
-      done
-      # Bash 3 ignores IFS set for command when doing expansions so have to set it separately
-      IFS=':'; export PATH="${after_list[*]}"; IFS="$OLDIFS"
-    fi
-  fi
+    disable_homebre_executables coreutils
```
robbyoconnor, line 1552: typo.

Issue #882, opened 2017-03-26
Deprecation policy?


Samureus:
Currently there are +/-300 definitions inside python-build, some of them very old. There are projects, e.g. PyPy and Miniconda/Anaconda that update very often, or new source releases that no longer compile with old versions of Debian (e.g. Stackless) due to old dependencies.

It would be great to start publishing some sort of deprecation policy in advance, e.g, "distribution/package XX will no longer work after version XX", to at least follow upstream deprecation policies, like for example, Ubuntu 12.04 reaches EOL in April:
https://lists.ubuntu.com/archives/ubuntu-security-announce/2017-March/003777.html

Or, Fedora 23 is no longer supported, even though we still have instructions for it:
https://fedoraproject.org/wiki/Releases

Or what to do with CPython 2.1.3 - 2.6.9? Does anybody use any of these?

The idea is that we can at least start doing integration testing on pull requests using maybe #617 (maybe with containers? we don't have enough buzzwords) to avoid "it works on my machine" issues.

joshfriend:
I just removed the `3.2-dev` and `3.3-dev` targets since those branches have reached EOL (37299c88511158983ceae7c819cfa75855f09119), as well as some very old pypy-dev branches (734c341c468e531c16902534dde237288fc21a2b).

> Or what to do with CPython 2.1.3 - 2.6.9? Does anybody use any of these?

I would like to talk about these too, I see people keep trying to install extremely old pythons (#910 for example). I think it would be totally ok to remove all pythons pre-2.6.x. (I think people absolutely do still use 2.6.x ☹️). There are also versions like `2.7` which a person might assume installs the latest version of the 2.7.x branch, but really it installs a version from 2010. I believe that all 3.x branches pre-3.4 are also EOL now. It may be a good idea to keep the last release from those branches around though (3.3.6, 3.2.6, and 2.6.9 for example?) 

> PyPy and Miniconda/Anaconda that update very often

Yes, it feels like these multiply at an exponential rate 😝 .

Samureus:
>     PyPy and Miniconda/Anaconda that update very often
> > Yes, it feels like these multiply at an exponential rate 😝 .

I'd say we could treat PyPy and *Conda as "evergreen" releases the same as Chrome/Firefox work, that is, just keep the last 3 releases and drop the rest.

[...]

PR #2061, opened 2021-09-11
macOS: fix the build of Python 2.7.18 on macOS 11.5.2 (Big Sur) + Apple Silicon

puhitaku:

### Prerequisite
  - #1768 : the most related topic
  - #2020
  - #2060 

### Description
Python 2.7.x build on macOS + Apple Silicon (arm64) is broken at present. This PR backports the patch of Python 3.8.7 for darwin arm64 from [Homebrew/formula-patches](https://github.com/Homebrew/formula-patches/blob/master/python/3.8.7.patch) and fixes the build of 2.7.18.

This depends on #2060 for a successful build on my environment. Some Macs that have no binutils from Homebrew may successfully build it without #2060 but it's not confirmed.

The hand-ported patch is somewhat huge; while it produces a working build with _tkinter, _ctypes, and _Qt, I'm not sure if it's necessary to apply this patch to all Python 2.7.x since Python 2.x have EOL'ed, as discussed in #882. This fix should cover most of Python 2 usage for developers who fight against obsolete Python 2 code.

Here is [a Gist](https://gist.github.com/puhitaku/acc733615e32d009d82994115aca2e0d#file-python-2-7-18-apple-silicon-pyenv-verbose-fail-log) of failing build log and successful build log.

### Environment
 - MacBook Pro (Late 2020, M1)
 - macOS 11.5.2 (Big Sur)

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/patches/2.7.18/Python-2.7.18/0001-Detect-arm64-in-configure.patch
+++ b/plugins/python-build/share/python-build/patches/2.7.18/Python-2.7.18/0001-Detect-arm64-in-configure.patch
@@ -1,0 +1,41 @@
+From 5495fb4719fa67ea3abeda0434699c8b28c77bb3 Mon Sep 17 00:00:00 2001
+From: Takumi Sueda <puhitaku@gmail.com>
+Date: Sat, 11 Sep 2021 16:50:14 +0900
+Subject: [PATCH 1/7] Detect arm64 in configure
+
+---
+ configure    | 3 +++
+ configure.ac | 3 +++
+ 2 files changed, 6 insertions(+)
+
+diff --git a/configure b/configure
+index 63d675312d..c40ec120de 100755
+--- a/configure
++++ b/configure
+@@ -8474,6 +8474,9 @@ fi
+     	ppc)
+     		MACOSX_DEFAULT_ARCH="ppc64"
+     		;;
++    	arm64)
++    		MACOSX_DEFAULT_ARCH="arm64"
++    		;;
+     	*)
+     		as_fn_error $? "Unexpected output of 'arch' on OSX" "$LINENO" 5
+     		;;
+diff --git a/configure.ac b/configure.ac
+index efe6922b5d..df05bee87b 100644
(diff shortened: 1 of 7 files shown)


native-api:
Looks like there are several concerns here mixed up together. Let's sort them out one by one.

* The alleged issue with binutils we'll discuss in #2060 .
* Until and unless concensus in #882 says otherwise, I'm fine with accepting patches for EOL versions from interested users. We just don't make any promises from our side that any EOL version will work or will keep working.
  * However, I'd like to make sure that we don't patch what we don't absolutely need to.
  * So please demonstrate the problem that each of your patches resolves and its root cause.

puhitaku:
Then I'll try to build 2.7.18 with the patch and without disabling binutils (without installing binutils with Homebrew). Fortunately, I have a MacBook Air (M1) which has a almost stock OS installed. After this succeeds, I'll revert back or remove 7c9bfa5 and split the mixed topics.

puhitaku:
I've successfully built Python 2.7.18 on a newly installed Mac so please take a look at [this Gist](https://gist.github.com/puhitaku/95b0c10f9f5eb730f407bef60971efe4) for details.

Now it's concluded that the patch to 2.7.18 is actually needed while Homebrew packages (binutils and coreutils) are another discrete topic.

puhitaku:
I've reverted 7c9bfa5. Please tell me if I should *pair-annihilate* the commit and the revert. I'll force push squashed commits.

native-api:
> 
> 
> I've reverted [7c9bfa5](https://github.com/pyenv/pyenv/commit/7c9bfa5f8d2478af1b194cf50c9346df2bedb51d). Please tell me if I should _pair-annihilate_ the commit and the revert. I'll force push squashed commits.

It's better to have a nice-looking history in a PR when it's presented for review (that's not a strict requirement though -- e.g. if the changes only make sense together, they're going to be squashed together anyway, so why bother).

That said, some PR checks currently fail. Please look into that, too.

puhitaku:
I've pushed update commits for a fixed build. Please approve workflows. @native-api

Issue #2062, opened 2021-09-13
Incomplete installation doc when using with pyenv-virtualenv plugin ?


PierreRust:
When using pyenv-virtualenv with pyenv>2, the command ` pyenv activate <venv>` appears to work correctly (i.e. not error is reported) although the shims are not pointing to the venv's python and pip.

When running `pyenv init --path | source` it works fine, although the documentation only mention to run `pyenv init - | source`

This has been tested on ubuntu 20.04, pyenv 2.0.6-7 with fish shell, but it has also been reported with bash in pyenv/pyenv-virtualenv#401 

I'm reporting it here because, while the problem appears with a plugin, it seems to depend on a missing(?) instruction in the installation documentation.

native-api:
I've made a PR with updated setup instructions based on user feedback so far. Please review: https://github.com/pyenv/pyenv/pull/2067

Issue #2058, opened 2021-09-11
Improve zsh installation instructions


shniubobo:
In zsh, `typeset -U PATH` can be used to avoid duplicate values in `PATH` (see [this manual page](https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html)). Also, according to [its user guide](https://zsh.sourceforge.io/Guide/zshguide02.html#l24), this is the recommended way to set `PATH`.

Therefore, the following configuration, either in `.zshrc` or `.zshenv`, would work perfectly for zsh (no need to seperate any of them into `.zprofile`):

```zsh
typeset -U PATH
export PYENV_ROOT="$HOME/.pyenv"
export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init --path)"
eval "$(pyenv init -)"
```

I am willing to submit a PR to update `README.md` if this configuration is acceptable.

anton-petrov:
Please, submit a PR.

native-api:
I've made a PR with updated setup instructions based on user feedback so far. Please review: https://github.com/pyenv/pyenv/pull/2067

Since the suggested configuration ensures there are no duplicate PATH entries -- unless the users specifically intends so -- `typeset -U` seems like unnecessary complication.

shniubobo:
> Since the suggested configuration ensures there are no duplicate PATH entries -- unless the users specifically intends so -- `typeset -U` seems like unnecessary complication.

Yes, that makes sense.

Issue #1990, opened 2021-06-26
[documentation] incorrect instructions for fish shell 


pawelmhm:
Instructions for fish in readme.md file here on github say
```
Execute this interactively:

set -Ux PYENV_ROOT $HOME/.pyenv
set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
And add this to ~/.config/fish/config.fish:

status is-login; and pyenv init --path | source
```

then it says you should

```
pyenv init - | source
```

So here is what I get when I follow [readme.md](https://github.com/pyenv/pyenv), you can see I have pyenv root set, I have fish user paths, I have config.fish exactly as README.md says. I have virtualenv with python 3.9.1 and default Python system wide 3.8.5. I get warnings about no longer setting path, virtualenv is activated properly, but python is not correct, it should be 3.9.1.

I would be really grateful for instructions for fish users. I'm long time fish and pyenv user, and pyenv worked great for me for a long time.

![Screenshot from 2021-06-26 15-11-02](https://user-images.githubusercontent.com/2700942/123514051-e0cb8100-d690-11eb-8684-6338219789bf.png)

native-api:
@faho, could you advise?

Also see https://github.com/pyenv/pyenv/pull/1991 -- someone else had trouble with fish instructions, too.

faho:
The fish instructions are a reasonably direct translation of the bash/zsh ones. The only difference is that it suggests using universal variables for things that, to a first approximation, don't change. So instead of adding $PYENV_ROOT/bin to $PATH every time, you add it to $fish_user_paths once and fish then takes care that it's in $PATH.

There is a bit of weirdness in the suggestion to do things differently for login shells - in my opinion the concept of a "login shell" is basically broken, but that's also consistent between the shells.

It would be possible to change all of this to suggest changing it in config.fish, and to not require a login shell, but I don't think that's where the error lies.

I would assume the problem here is that you don't have a login fish running, so `pyenv init --path | source` isn't ever actually run?

The output of `set --show PATH` and `set --show fish_user_paths` would be useful.

pawelmhm:
> The output of set --show PATH and set --show fish_user_paths would be useful.
![Screenshot from 2021-06-27 19-22-33](https://user-images.githubusercontent.com/2700942/123553689-0afa6d00-d77d-11eb-8309-58aad093093b.png)

> I would assume the problem here is that you don't have a login fish running (because you never did chsh)

[...]

PR #2067, opened 2021-09-18
Update setup instructions in the Readme

native-api:
See the rendered README at https://github.com/native-api/pyenv/tree/readme#readme

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2062
  - Closes https://github.com/pyenv/pyenv/issues/2058
  - Closes https://github.com/pyenv/pyenv/issues/2068
  - Closes https://github.com/pyenv/pyenv/issues/1990

### Description

* Update the setup code in the README based on user feedback
* Remove setup code from installer output since it proved

Considered adding logic for pyenv-installer that would replicate the README (like RVM does). Since the logic proves rather complex and multi-case, let's wait for positivbe feedback from users on the README-driven logic before we can rely on it enough to apply it unattended.

**This is my last-ditch effort to make clear instructions. If these are still too hard to follow, I'll make it so that just `pyenv init -` works, too, as previously, making `--path` optional -- but that will yield poor results in some setups (e.g. in distros that add `~/.local/bin` to `PATH` in profile) where it's preferreable to add a `PATH` entry at a different moment.**

### Tests

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,146 +223,157 @@ easy to fork and contribute any changes back upstream.
 
    **Note:** The below instructions for specific shells are designed for common shell setups;
    they also install shell functions into interactive shells only.  
-   If you have an uncommon setup and/or special needs and they don't work for you,
-   use the guidance text and the [Advanced Configuration](#advanced-configuration)
+   If you have an uncommon setup and/or needs and they don't work for you,
+   use the [Advanced Configuration](#advanced-configuration)
    section below to figure out what you need to do in your specific case.
    
-   1. **Adjust the session-wide environment for your account.** Define
-   the `PYENV_ROOT` environment variable to point to the path where
-   you cloned the Pyenv repo, add the `pyenv` command-line utility to your `PATH`,
-   run the output of `pyenv init --path` to enable shims.
+   **General MacOS note:**
+   Make sure that your terminal app is configured to run the shell as a login shell
+   (especially if you're using an alternative terminal app and/or shell).
+   The configuration samples for MacOS are written under this assumption and won't work otherwise.
    
-      These commands need to be added into your shell startup files in such a way
-      that _they are executed only once per session, by its login shell._
-      This typically means they need to be added into a per-user shell-specific
-      `~/.*profile` file, _and_ into `~/.profile`, too, so that they are also
-      run by GUI managers (which typically act as a `sh` login shell).
+   - For **Bash**:
 
-      **MacOS note:** If you installed Pyenv with Homebrew, you don't need
-      to add the `PYENV_ROOT=` and `PATH=` lines.
-      You also don't need to add commands into `~/.profile` if your shell doesn't use it.
-   
-      - For **Bash**:
+      - **If your `~/.profile` sources `~/.bashrc` (Debian, Ubuntu, Mint):**
+
+         ~~~bash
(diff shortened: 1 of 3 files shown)


shniubobo:
Please note that WSL also runs shells as login shells by default.

```console
$ [[ -o login ]] && echo "login"
login
$ zsh
$ [[ -o login ]] && echo "login"  # no output
```

Tested with Windows Terminal 1.10.2383.0 and:
1. Ubuntu 18.04 WSL, zsh 5.4.2
2. ArchWSL, zsh 5.8

Maybe the config for MacOS also applies to WSL?

native-api:
> 
> 
> Please note that WSL also runs shells as login shells by default.
> 
> ```
> $ [[ -o login ]] && echo "login"
> login
> $ zsh
> $ [[ -o login ]] && echo "login"  # no output
> ```
> 
> Tested with Windows Terminal 1.10.2383.0 and:
> 
>     1. Ubuntu 18.04 WSL, zsh 5.4.2
> 
>     2. ArchWSL, zsh 5.8
> 
> 
> Maybe the config for MacOS also applies to WSL?

AFAICS, WSL is effectively an Ubuntu console session in a Windows terminal window. So instructions for Ubuntu should also apply to it, no need for a separate section.

shniubobo:
> AFAICS, WSL is effectively an Ubuntu console session in a Windows terminal window. So instructions for Ubuntu should also apply to it, no need for a separate section.

Of course instructions for Ubuntu work for WSL, and of course no extra section is needed.

But instructions for Ubuntu also work for MacOS (except that an extra note is needed to tell users to run the shell as a login shell). Why should there be a seperate section for MacOS? Because there is no need to add three extra lines into `.profile` [1]. This is the same as WSL.

What I meant by saying "maybe the config for MacOS also applies to WSL" was that, instead of writing:

> MacOS, if Pyenv is installed with a Git checkout:
> In MacOS, restarting terminal windows is enough

we could write:

> MacOS or WSL, if Pyenv is installed with a Git checkout:
> In MacOS and WSL, restarting terminal windows is enough

[1] I have never used MacOS, but I assume lines in `.profile` are ignored if zsh is the login shell, as they should be on other OSes, and thus these extra lines will not break pyenv; please let me know if I am wrong.

native-api:
> What I meant by saying "maybe the config for MacOS also applies to WSL" was that, instead of writing:
> 
> > MacOS, if Pyenv is installed with a Git checkout:
> > In MacOS, restarting terminal windows is enough
> 
> we could write:
> 
> > MacOS or WSL, if Pyenv is installed with a Git checkout:
> > In MacOS and WSL, restarting terminal windows is enough

That'd not be correct. As you can see, WSL is not a separate OS, it's merely another way to use a pre-existing Linux distro -- one of at least a few different ones as you've demonstrated yourself. So there isn't a single, separate environment called "WSL", and as such, there cannot be a separate, not to mention single, set of instructions for it, the instructions for whatever distro you're using through it apply instead.

Moreover, we do not officially support WSL (in particulae, do not test things on it as part of our maintenance) so I'm hesitant to treat it as a first-class citizen in the README as it would give users a false impression.

anton-petrov:
I'll add some instructions for WSL2 later (using it in my work too)...

PR #2073, opened 2021-09-21
GitHub actions tests

native-api:

### Prerequisite

### Description

* Add running `make test` in Github Actions now that Travis check no longer runs; fix failures.
* Add diagnostic output to test libraries hat makes it possible to diagnose errors.

### Tests

### DIFF
--- a/.github/workflows/pyenv_tests.yml
+++ b/.github/workflows/pyenv_tests.yml
@@ -1,0 +1,33 @@
+name: pyenv_tests
+on: [pull_request, push]
+jobs:
+  pyenv_tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-20.04
+          - ubuntu-18.04
+          - macos-11
+          - macos-10.15
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      # Normally, we would use the superbly maintained...
+      # - uses: actions/setup-python@v2
+      #   with:
+      #    python-version: ${{ matrix.python-version }}
+      # ... but in the repo, we want to test pyenv builds on Ubuntu
+      # - run: |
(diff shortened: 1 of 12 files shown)


joshfriend:
Any reason you kept the Travis CI config around still?

native-api:
@joshfriend I don't know why it was disabled -- that certainly wasn't me. So until that is clarified, I didn't think it was a good idea to touch it.

joshfriend:
I didn't disable it, but it's probably for the best given the Travis CI security incident, although I don't know how to revoke the github API key or whatever travis is using to create releasees for tags

native-api:
> given the Travis CI security incident
I read their bulletin. According to it, it only exposed secret variables to external PR builds (PR builds where the PR branch is from a fork) -- so they could leak if a PR had malicious code.
There have been no PRs with any suspect of malicious code in that timeframe here, and the secret value is included in such a way that it's hard to access from build logic (need to parse Travis' autogenerated script to my best knowledge, so code doing smth like that would be quite obvious).
> I don't know how to revoke the github API key or whatever travis is using to create releasees for tags
I don't see any kind of access keys in this repo's settings. Personal tokens can be used for releases. Since @anton-petrov is doing the releases, I guess it's his personal token encrypted in `.travis,yml`.

native-api:
> Since @anton-petrov is doing the releases, I guess it's his personal token encrypted in .travis.yml.

Checking Git Blame, @yyuu added the encrypted value. So it must be his.

As mentioned earlier, I believe it's highly unlikely that it has leaked -- yet he can revoke that token nevertheless just for the peace of mind.

yyuu:
Hmm, I couldn't figure out how to revoke my Travis token 🤔  Maybe uninstall & re-install the GitHub App for Travis works...? Maybe it's a better idea to consider migrating to GitHub Actions completely... 🙊

Issue #2070, opened 2021-09-20
`pyenv-build` calls `conda install --yes pip` and updates `conda` to a different version than desired


humitos:
Hi all! Thanks for this amazing project.

I found a problem with `conda` version not being the desired one. This happens because after `conda` is installed `pyenv-build` calls `conda install --yes pip` at this line https://github.com/pyenv/pyenv/blob/v2.0.6/plugins/python-build/bin/python-build#L1037 which makes conda to **updates all the packages installed in the base system**:

```
...

The following packages will be UPDATED:

  ...
  conda                                       4.7.12-py37_0 --> 4.10.3-py37h06a4308_0
  ...

...
```

Note that `conda-4.7.12` was updated to `conda-4.10.3` but that's not version that I told `pyenv-build` to install.

I'd suggest to not install `pip` (why `pip` is installed by default?) or find a way to tell conda to **not update other dependencies** (I tried passing `--no-update-deps` but it didn't work as expected).

Let me know if I can be useful here.

<details>
<summary>Click to see the full output</summary>

```
▶ asdf install python miniconda3-4.7.12
python-build miniconda3-4.7.12 /home/humitos/.asdf/installs/python/miniconda3-4.7.12
Downloading Miniconda3-4.7.12-Linux-x86_64.sh...
-> https://repo.anaconda.com/miniconda/Miniconda3-4.7.12-Linux-x86_64.sh
Installing Miniconda3-4.7.12-Linux-x86_64...
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /home/humitos/.asdf/installs/python/miniconda3-4.7.12

  added / updated specs:
    - pip


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    ca-certificates-2021.7.5   |       h06a4308_1         113 KB
    certifi-2021.5.30          |   py37h06a4308_0         139 KB
    conda-4.10.3               |   py37h06a4308_0         2.9 MB
    openssl-1.1.1l             |       h7f8727e_0         2.5 MB
    pip-21.0.1                 |   py37h06a4308_0         1.8 MB
    ------------------------------------------------------------
                                           Total:         7.4 MB

The following packages will be UPDATED:

  ca-certificates                               2019.8.28-0 --> 2021.7.5-h06a4308_1
  certifi                                  2019.9.11-py37_0 --> 2021.5.30-py37h06a4308_0
[...]

PR #2074, opened 2021-09-23
Don't update conda when installing pip

anton-petrov:
### Solution to tell conda to not update other dependencies when installing pip.
The main idea is to specify the existing version of the conda when installing the pip package.  Conda is the main tool for installing packages in miniconda & miniforge. So, some users does not always need the latest version of conda, for example. After installation of miniconda or miniforge, user can update all packages  manually or use such a setting as  `conda config --set auto_update_conda False`.
Thanks to @native-api, @humitos for the suggested solutions.

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2070

### Description
- [x ] Here are some details about my PR
- https://github.com/pyenv/pyenv/issues/2070#issue-1000827548

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1042,7 +1042,9 @@ build_package_anaconda() {
 
 build_package_miniconda() {
   build_package_anaconda "$@"
-  "${PREFIX_PATH}/bin/conda" install --yes "pip"
+  # Workaround to not upgrade conda when installing pip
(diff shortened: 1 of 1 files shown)


humitos (approved):
Happy to see this workaround implemented in pyenv! Thanks for the good discussion we had in the issue 💪🏼 

I can now remove my hack to apply the same patch from our code 😬

PR #2075, opened 2021-09-24
Add pyston-2.3.1 support

kmod:
We changed the directory structure for pyston 2.3.1 slightly, where we now
include a top-level "pyston-2.3.1" directory when you unpack the tarball.

My first time looking at pyenv so I'm not super confident in how I've laid things out.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -881,18 +881,6 @@ build_package_pyston() {
   build_package_copy
   mv "${PREFIX_PATH}/usr/bin" "${PREFIX_PATH}/bin"
   mv "${PREFIX_PATH}/usr/lib" "${PREFIX_PATH}/lib"
-  local bin
-  shopt -s nullglob
-  for bin in "usr/bin/"*; do
-    if [ -f "${bin}" ] && [ -x "${bin}" ] && [ ! -L "${bin}" ]; then
-      case "${bin##*/}" in
-      "python"* )
-        ( cd "${PREFIX_PATH}/bin" && ln -fs ../usr/bin/"${bin##*/}" "python" )
-        ;;
-      esac
-    fi
-  done
-  shopt -u nullglob
 }
 
 build_package_ironpython() {
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -858,8 +858,8 @@ build_package_jython_builder() {
   ( cd "dist" && build_package_jython )
 }
 
-build_package_pyston() {
-  # currently supported version 2.2 and upper
+build_package_pyston2.2() {
+  # currently supported version 2.2 and 2.3
   build_package_copy
   mkdir -p "${PREFIX_PATH}/bin" "${PREFIX_PATH}/lib"
   local bin
@@ -869,6 +869,25 @@ build_package_pyston() {
       case "${bin##*/}" in
       "pyston"* )
         ( cd "${PREFIX_PATH}/bin" && ln -fs "${bin##*/}" "python" )
+        ;;
+      esac
+    fi
+  done
+  shopt -u nullglob
+}
+
+build_package_pyston() {
+  # currently supported version 2.3.1 and higher
+  build_package_copy
+  mv "${PREFIX_PATH}/usr/bin" "${PREFIX_PATH}/bin"
+  mv "${PREFIX_PATH}/usr/lib" "${PREFIX_PATH}/lib"
+  local bin
+  shopt -s nullglob
+  for bin in "usr/bin/"*; do
+    if [ -f "${bin}" ] && [ -x "${bin}" ] && [ ! -L "${bin}" ]; then
+      case "${bin##*/}" in
+      "python"* )
+        ( cd "${PREFIX_PATH}/bin" && ln -fs ../usr/bin/"${bin##*/}" "python" )
         ;;
       esac
     fi
--- a/plugins/python-build/share/python-build/pyston-2.3.1
+++ b/plugins/python-build/share/python-build/pyston-2.3.1
@@ -1,0 +1,19 @@
+echo
+colorize 1 "WARNING"
+echo ": Pyston only runs on x86_64 platforms, and only has been tested on Ubuntu"
+echo "so, your mileage may vary with other Linux distributions."
+echo
+
+case "$(pyston_architecture 2>/dev/null || true)" in
+"linux64" )
+  install_package "pyston_2.3.1" "https://github.com/pyston/pyston/releases/download/pyston_2.3.1/pyston_2.3.1_portable.tar.gz#04ba4cb0249973870b8f198f05d6d22c054cb5b059865cacf7e623b3624949e8" "pyston" verify_py38 get_pip
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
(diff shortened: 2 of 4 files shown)


plugins/python-build/share/python-build/pyston-2.3.1
```diff
@@ -0,0 +1,19 @@
+echo
+colorize 1 "WARNING"
+echo ": Pyston only runs on x86_64 platforms, and only has been tested on Ubuntu"
+echo "so, your mileage may vary with other Linux distributions."
+echo
```
native-api, line 5: There's `require_distro()` to display a warning like this. No need to display it unless the user runs an unsupported distribution.
An an unsupported architecture case is already handled below.

plugins/python-build/bin/python-build
```diff
@@ -876,6 +876,25 @@ build_package_pyston() {
...
+  build_package_copy
+  mv "${PREFIX_PATH}/usr/bin" "${PREFIX_PATH}/bin"
+  mv "${PREFIX_PATH}/usr/lib" "${PREFIX_PATH}/lib"
+  local bin
+  shopt -s nullglob
+  for bin in "usr/bin/"*; do
+    if [ -f "${bin}" ] && [ -x "${bin}" ] && [ ! -L "${bin}" ]; then
+      case "${bin##*/}" in
+      "python"* )
+        ( cd "${PREFIX_PATH}/bin" && ln -fs ../usr/bin/"${bin##*/}" "python" )
```
native-api, line 890: Shouldn't it also create a `python3` and `python3.8` aliases in accordance with https://www.python.org/dev/peps/pep-0394/?

```
$ ls -lR ~/.pyenv/versions/pyston-2.3.1/{bin,usr}
/home/vmuser/.pyenv/versions/pyston-2.3.1/bin:
total 8856
-rwxrwxr-x 1 vmuser vmuser     257 сен 24 22:27 pip
-rwxrwxr-x 1 vmuser vmuser     257 сен 24 22:27 pip3
-rwxrwxr-x 1 vmuser vmuser     257 сен 24 22:27 pip3.8
lrwxrwxrwx 1 vmuser vmuser      19 сен 24 22:27 pyston -> python3.8-pyston2.3
lrwxrwxrwx 1 vmuser vmuser      19 сен 24 22:27 pyston3 -> python3.8-pyston2.3
lrwxrwxrwx 1 vmuser vmuser      19 сен 24 22:27 python -> python3.8-pyston2.3
-rwxr-xr-x 1 vmuser vmuser 9051080 сен 24 22:27 python3.8-pyston2.3
-rwxrwxr-x 1 vmuser vmuser     244 сен 24 22:27 wheel
<...>
```

Take a look at `build_package_symlink_version_suffix()` (it's automatically called from `verify_python*`). IIRC you create a `pythonX.Y` executable and it creates the rest for you.

plugins/python-build/bin/python-build
```diff
@@ -876,6 +876,25 @@ build_package_pyston() {
...
 }
 
+build_package_pyston() {
+  # currently supported version 2.3.1 and higher
+  build_package_copy
+  mv "${PREFIX_PATH}/usr/bin" "${PREFIX_PATH}/bin"
+  mv "${PREFIX_PATH}/usr/lib" "${PREFIX_PATH}/lib"
+  local bin
+  shopt -s nullglob
+  for bin in "usr/bin/"*; do
```
native-api, line 886: Not sure... won't this always be a no-op since you've just moved `usr/bin` away? What's the CWD here, and how do you know you can rely on it?

native-api: This code also looks overcomplicated. If all you need is to symlink `pythonX.Y-*` to `python`, you can just take the first result of `${PREFIX_PATH}/bin/python?.?-*`

native-api:
```
ln -s "$(basename "${PREFIX_PATH}/bin/python?.?-*")" "${PREFIX_PATH}/bin/python"
```

plugins/python-build/share/python-build/pyston-2.3.1
kmod: I copied this warning from the other pyston rules but I'm happy to change all of them if you'd like.

plugins/python-build/bin/python-build
kmod: I'm not an expert on this stuff but we're generally not providing any symlinks that typically point to a CPython install, such as "python3" or "python3.8", and instead call ourselves "pyston", "pyston3", and "python3.8-pyston2.3". This is more important for things like libraries, where we can't run the compiled modules in `/usr/lib/python3.8`, and need a separate place to put our modules.

Our interpretation is that our version number is "3.8-pyston2.3", so we still adhere to the "python$(VERSION)" convention, except we break anything that assumes version numbers are X.Y.

I'm happy to create a "python3.8" alias as part of the pyenv install if that's what the rest of the code expects, though personally I'd advocate for "python3.8-pyston2.3" getting the same treatment as "python3.8".

plugins/python-build/bin/python-build
kmod: Sounds good. Looks like the code might not be necessary at all; I tested to make sure that a `python` alias was being written, but clearly it's not being written by my non-functional code here, so maybe there's something else that's writing it? My guess is that the code I copied this from was only necessary in order to deal with the fact that we didn't used to have anything called `python*`, so maybe we can just remove it?

plugins/python-build/bin/python-build
native-api: The point of https://www.python.org/dev/peps/pep-0394/ is so that a script designed for a specific Python version can work with any Python of appropriate version without changes.
So if you don't care about anything not specifically written for Pyston, that indeed doesn't matter to you.

plugins/python-build/share/python-build/pyston-2.3.1
native-api: Displaying irrelevant warnings is certainly poor UX.

plugins/python-build/bin/python-build
native-api: It's your job to test if a code change works :-)

plugins/python-build/share/python-build/pyston-2.3.1
kmod: This isn't my project but I have a strong inclination that all of the Pyston installers should have the same behavior. Are you sure that you'd like this one to be different and use require_distro?

plugins/python-build/bin/python-build
kmod: I can be snarky too and say that I followed all of the steps that were provided :)

plugins/python-build/bin/python-build
kmod: I agree that it would be nice if the Pyston pyenv installation included python3 and python3.8 symlinks, but my understanding is that you're advocating that we provide those aliases in our installer. I don't believe that's the right course of action since that is how people refer to the system CPython and I don't think people want that to be redirected to Pyston (though hopefully some day we will get OS distributions to ship Pyston as the system Python). I'm happy to update this PR or the installer code so that it writes out the additional symlinks though.

plugins/python-build/bin/python-build
native-api: Looks like there was a misunderstanding. I wrote that solely because you seemed to be requesting me to test your changes for you. If after removing the code, the installer still works as intended then of course it's okay to remove it...

PR #2084, opened 2021-10-01
Fix installation instructions for Ubuntu.

tvalentyn:
### Description

Current instructions didn't work for me on Ubuntu installation.
Since `.bashrc` doesn't source `.profile` (it's the opposite), `$PATH` variables were not updated by the time `pyenv` is called. 

Proposing the fix that makes them work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ easy to fork and contribute any changes back upstream.
          a \
          export PATH="$PYENV_ROOT/bin:$PATH"
          a \
-         ' -e ':a' -e '$!{n;ba};}' ~/.profile
+         ' -e ':a' -e '$!{n;ba};}' ~/.bashrc
          echo 'eval "$(pyenv init --path)"' >>~/.profile
 
          echo 'eval "$(pyenv init -)"' >> ~/.bashrc


native-api:
The instructions have been updated fuurther since. Please see the current versions at https://github.com/pyenv/pyenv#readme.

native-api:
As per the instructions, you need to restart your login session before using Pyenv. That way, `.profile` will be sourced.

native-api:
As mentioned in https://github.com/pyenv/pyenv/pull/2067#issue-1000138452, we can lift the session restart requirement, but programs outside a terminal session will still not find Pyenv until you do so (but before v.2, they weren't able to find Pyenv at all). Will that be better?

tvalentyn:
> As per the instructions, you need to restart your login session before using Pyenv. That way, .profile will be sourced.

It still doesn't work after restarting login session, and .profile does not get sourced, perhaps this is specific to Ubuntu 16 that I am running this on.

native-api:
Are you working in GUI or console session? Do you have `~/.bash_profile`?

tvalentyn:
console session, no `.bash_profile`, only `.profile` and `.bashrc`.

native-api:
Then most probably, your shell is somehow run as a non-login shell. Check that, and what startup files it sources (e.g. by placing a debug print statement at the start of each one).

tvalentyn:
Thanks, yes, it was running as a non-login shell.

tvalentyn:
For my education, why would it be wrong to modify `$PATH` in `.bashrc` instead of `.profile`, given that `.profile` sources `.bashrc` ?

tvalentyn:
For example, also since we are adding `         echo 'eval "$(pyenv init -)"' >> ~/.bashrc`	- how will pyenv be resolved withing `.bashrc` if the pyenv path is not defined in `.bashrc` ?

native-api:
`.bashrc` can be sourced multiple times. So it's problematic to do `PATH` manipulations there, they should go into profile files instead. See https://unix.stackexchange.com/questions/609522/should-bashrc-commands-support-shell-reloading

native-api:
> how will pyenv be resolved withing `.bashrc` if the pyenv path is not defined in `.bashrc` ?

That's why the suggested setup puts adding Pyenv to PATH at the _start_ of `.profile`, before it sources `.bashrc`.

tvalentyn:
Thanks!

PR #2094, opened 2021-10-05
Disable conda package self-update

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2070

### Description
https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#update-conda-automatically-auto-update-conda is the option intended to prevent implicit `conda` package updates.

CAUTION: it seems to [still not protect from `conda update --all`](https://github.com/conda/conda/issues/5294)!! Pinning protects from it, but it's local to an environment. Since we shim `conda`, we can propagate the pin to other environments as they are created, but that'd be another feature.

### Tests
Test the feature + helper fns for Anaconda

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1059,9 +1059,7 @@ build_package_miniconda_autoupdate() {
 
 build_package_miniconda() {
   build_package_anaconda "$@"
-  # Workaround to not upgrade conda when installing pip
-  # see https://github.com/pyenv/pyenv/issues/2070
-  "${PREFIX_PATH}/bin/conda" install --yes "pip" "conda=$(${PREFIX_PATH}/bin/conda --version | cut -d ' ' -f 2)"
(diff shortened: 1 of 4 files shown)


anton-petrov:
@native-api I am opposed to changing the default settings for anaconda. End user doesn't even assume that auto-update is turned off for the conda! This will create a bunch of problems!
So, after installing the anaconda/miniconda, at the end of the installation process, we should return all the default settings.

humitos:
Thanks for taking the time to work on this!

I agree with @anton-petrov about not changing the defaults permanently. Also, I think that chunk of code could have a comment explaining why that is required.

native-api:
> End user doesn't even assume that auto-update is turned off for the conda!

@humitos That's strange. In https://github.com/pyenv/pyenv/issues/2070, you assumed exactly that. And for a good reason -- our naming system for Anaconda is based on that assumption!

Well, if you're fine with things the way they are, who am I to judge you... Let's wait for someone else to run into this and give us more feedback...

PR #2144, opened 2021-11-04
Use command -v instead of which in python-build test

scop:

### Prerequisite
  * https://github.com/rbenv/rbenv/pull/1359

### Description

`which` has been deprecated in Debian's debianutils 5.0+ (currently only
Debian unstable), use `command -v` instead like is already done
elsewhere.

### Tests

### DIFF
--- a/plugins/python-build/test/pyenv.bats
+++ b/plugins/python-build/test/pyenv.bats
@@ -183,7 +183,7 @@ OUT
 }
 
 @test "pyenv-install has usage help preface" {
-  run head "$(which pyenv-install)"
+  run head "$(command -v pyenv-install)"
   assert_output_contains 'Usage: pyenv install'
 }
 
@@ -212,6 +212,6 @@ OUT
 }
 
 @test "pyenv-uninstall has usage help preface" {
-  run head "$(which pyenv-uninstall)"
+  run head "$(command -v pyenv-uninstall)"
   assert_output_contains 'Usage: pyenv uninstall'
 }
--- a/test/test_helper.bash
+++ b/test/test_helper.bash
@@ -110,28 +110,13 @@ assert() {
   fi
 }
 
-which_a() {
-    if command -v which; then
-        # which -a is portable
-        which -a "$@"
-    elif command -v whereis; then
-        local where
-        # -b to whereis is not portable, not available e.g. in macOS 11
-        where=$(whereis -b "$@") || return $?
-        [ -z "$where" ] && return 1
-    else
-        echo "which -a or whereis -b required" >&2
-        return 127
-    fi
-}
-
 # Output a modified PATH that ensures that the given executable is not present,
 # but in which system utils necessary for pyenv operation are still available.
 path_without() {
   local path=":${PATH}:"
   for exe; do 
     local found alt util
-    for found in $(PATH="$path" which_a "$exe"); do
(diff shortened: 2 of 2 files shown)


native-api (changes requested):
These are not all the useages.

And what would be the replacement for `which -a` (which is also used)?

scop:
They were all I could find in a quick search, there's a lot of `pyenv which` that are obviously a different thing. But let me know what I missed in addition to the `which -a`.

One possibility for `which -a` would be `whereis -b`, done here now.

native-api:
> `which` has been deprecated in Debian's debianutils 5.0+ (currently only
> Debian unstable), use `command -v` instead like is already done
> elsewhere.

Could you also link to the source of that and relevant discussion? I'm quite interested in the rationale and officially suggested alternatives.

scop:
https://salsa.debian.org/debian/debianutils/-/commit/3a8dd10b4502f7bae8fc6973c13ce23fc9da7efb
https://salsa.debian.org/debian/debianutils/-/commit/610a65c6ef040872a124576ddb504c2d4918e880

I'm not aware of a related mailing list or such discussion (but would be surprised if it didn't exist).

native-api:
> https://salsa.debian.org/debian/debianutils/-/commit/3a8dd10b4502f7bae8fc6973c13ce23fc9da7efb

Okay, they suggest `type` and `command` as replacements and the rationale is:

> Since `type` and `command -v`
> were mandated by POSIX, this utility is no longer useful for
> maintainer scripts and thus will be removed from debianutils.

So they think it's not useful _for maintainer scripts._

---

The replacement for `which -a` is `type -aP`.
`whereis` as a bad replacement since its output has to be parsed and it can only use spaces as delimiters.

---

HOWEVER. I see this utility missing from `debianutils` `4.9.1`... but [present again in `4.11.2`](https://packages.ubuntu.com/impish/amd64/debianutils/filelist)! Looks like it was reintroduced?

scop:
> So they think it's not useful _for maintainer scripts._

As it stands, there's no other implementation of it in the repos, so the only implementation there actually is is scheduled for removal.

> The replacement for `which -a` is `type -aP`. `whereis` is a bad replacement since its output has to be parsed and it can only use spaces as delimiters.

Agree `whereis` is far from nice due to parsing, but also because it's not standard either. But `type -aP` isn't all that good either, because `-a` and `-P` to it are not portable: neither is in POSIX, zsh type doesn't have -P, FreeBSD sh type has neither.

> HOWEVER. I see this utility missing from `debianutils` `4.9.1`... but [present again in `4.11.2`](https://packages.ubuntu.com/impish/amd64/debianutils/filelist)! Looks like it was reintroduced? Edit: actually, it's present in `4.9.1`, too. They just moved it from `/bin` to `/usr/bin`.

Yes, it's "only" deprecated for now, not yet removed.

native-api:
> But `type -aP` isn't all that good either, because `-a` and `-P` to it are not portable: neither is in POSIX, zsh type doesn't have -P, FreeBSD sh type has neither.

Pyenv is written in Bash (except eval'ed blocks which are different for different shells) so we only have to care about different Bash versions.

scop:
But of course :facepalm:, thanks for the sanity check. Switched to `type -aP` now.

PR #2157, opened 2021-11-13
Support arm64 identifier from brew coreutils' uname for 2.7/3.7+.

chipx86:
### Description

During the initial phase of compilation, Python's `config.sub` attempts to determine whether the OS and machine identifier type returned by `uname` is a valid target. On M1 Macs, the system-default `uname` returns a value that's eventually parsed as `arm-apple-darwin21.0.1`, which passes the pattern matching.

However, if the coreutils Homebrew target is installed and its version of `uname` is in the path, the result is `arm64-apple-darwin21.0.1`. This isn't handled by the existing patterns, and fails the build.

This appears to be a problem on all shipping versions of Python, tested on Python 2.7.18, 3.6.15, 3.7.12, 3.8.12, 3.9.8, and 3.10.0 (the latest versions in each of their respective series).

This change patches `config.sub` to add `arm64-*` to the list of patterns, alongside `arm-*`, allowing compilation to proceed. Patches have been added for all the versions of Python mentioned above with the exception of 3.6.15, which will need additional work in order to compile.

### Tests

All existing unit tests pass.

Performed a clean build of each of the patched versions of Python on a 2021 M1 MacBook Pro and tested each.

### DIFF
--- a/plugins/python-build/share/python-build/patches/2.7.18/Python-2.7.18/0008-Support-arm64-identifier-from-brew-coreutils-uname-f.patch
+++ b/plugins/python-build/share/python-build/patches/2.7.18/Python-2.7.18/0008-Support-arm64-identifier-from-brew-coreutils-uname-f.patch
@@ -1,0 +1,38 @@
+From 88a8e34c7eb48dcb80d8c2717e6f3fa5ed049819 Mon Sep 17 00:00:00 2001
+From: Christian Hammond <christian@beanbaginc.com>
+Date: Fri, 12 Nov 2021 23:20:35 -0800
+Subject: [PATCH] Support "arm64" identifier from brew coreutils' uname for
+ Python 2.7.18.
+
+Python 2.7.18's `config.sub` attempts to determine whether the OS and
+machine identifier type returned by `uname` is a valid target for
+compilation. On M1 Macs, the system-default `uname` returns a value
+that's eventually parsed as "arm-apple-darwin21.0.1", which passes the
+pattern matching.
+
+However, if the coreutils Homebrew target is installed and its version
+of `uname` is in the path, the result is "arm64-apple-darwin21.0.1".
+This isn't handled by the existing patterns, and fails the build.
+
+This change adds `arm64-*` to the list of patterns, alongside `arm-*`,
+allowing compilation to proceed.
+---
+ config.sub | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/config.sub b/config.sub
+index 40ea5dfe11..6fa6fe879a 100755
+--- a/config.sub
++++ b/config.sub
+@@ -376,7 +376,7 @@ case $basic_machine in
+ 	| alpha-* | alphaev[4-8]-* | alphaev56-* | alphaev6[78]-* \
+ 	| alpha64-* | alpha64ev[4-8]-* | alpha64ev56-* | alpha64ev6[78]-* \
+ 	| alphapca5[67]-* | alpha64pca5[67]-* | arc-* | arceb-* \
+-	| arm-*  | armbe-* | armle-* | armeb-* | armv*-* \
++	| arm-* | arm64-* | armbe-* | armle-* | armeb-* | armv*-* \
+ 	| avr-* | avr32-* \
+ 	| ba-* \
+ 	| be32-* | be64-* \
+-- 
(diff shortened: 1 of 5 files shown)


native-api:
This is supposed to already be covered by https://github.com/pyenv/pyenv/pull/2020, and in a version-independent manner, too. Isn't it?

native-api (commented):
First of all, 3.10 already supports the `arm64` arch (at least, it's supposed to, see https://github.com/pyenv/pyenv/pull/2020#issuecomment-891911842)

Second, I'm not sure if such a primitive patch is reliable.
The core problem was using a too old version of autoconf. So logically speaking, we should patch `config.sub` to a newer version of itself as per https://github.com/pyenv/pyenv/pull/2020#issuecomment-914617428 -- which likely has other relevant changes, too.

chipx86:
That's a fair point.

3.10 does support arm64 in theory, but does not support the coreutils version of `uname`:

```checking build system type... Invalid configuration `arm64-apple-darwin21.0.1': machine `arm64-apple' not recognized
configure: error: /bin/sh ./config.sub arm64-apple-darwin21.0.1 failed
```

All released versions of Python I've tested have suffered from this.

chipx86:
Alright, I've pulled down the latest version of `config.sub` (and `config.guess`) and have verified it addresses the problem. I'll rework this to update those with the latest version of each sometime this weekend (3AM here right now).

I have a couple additional patches I plan to put up. One fixes the `ctypes` module on Python 3.7.10, which is busted (I'm porting over the same patches that are in the Python 2.7.18 patch bundle). 3.6.15 is completely unable to compile, so I'm working on that as well (also needs those same patches, plus a couple other fixes to `configure`/`configure.ac`).

3.6 is near finished, and 3.7 is working without issues (put it through its paces by installing all of my products/dependencies into a new 3.7 virtualenv, ran all of our pretty comprehensive test suites — failed before, checks out now).

native-api:
>  I'll rework this to update those with the latest version of each sometime this weekend (3AM here right now).

Before you delve into that, please consider using the workaround from #2020 in the other affected versions instead.

Since CPython officially only supports Apple's toolchain, we should avoid using a different set of tools since that can introduce incompatibilities (e.g. if the arch triplet is different, precompiled extensions will probably become incompatible).

So in this light, it seems logical to just remove Coreutils from `PATH` in the compilation environment rather than support all their possible differences from Apple's tools.

native-api:
> First of all, 3.10 already supports the `arm64` arch (at least, it's supposed to, see [#2020 (comment)](https://github.com/pyenv/pyenv/pull/2020#issuecomment-891911842))
> 

Shite! They claimed to have fixed that, [but they didn't](https://github.com/python/cpython/pull/25450)! The `config.guess`/`config,sub` update has only happened in https://github.com/python/cpython/pull/25860/files which will go into `3.11`!

So we'll have to keep the workaround for 3.10 as well...

chipx86:
That works for me. Is there a preferred approach for manipulating the `PATH` for this? Should this be done in, say, `plugins/python-build/share/python-build/2.7.18`, or in a patch file? I'm still new to the inner workings of pyenv itself (trying to learn as I go).

Let me know if you have any preliminary thoughts on the other work I mentioned (or if you're aware of any work I'd be duplicating). I'll try to have pull requests up tomorrow, and can make any changes necessary. My primary goal this weekend is to get things working well enough to let me resume development on products for `$DAYJOB` here before the week starts, but I want to go about it right, and give back to pyenv where and how I can.

Thanks for your input on this. Calling it a night here, and hope you have a good day :)

native-api:
> Is there a preferred approach for manipulating the `PATH` for this?

If you're asking about how to apply #2020 to other versions, see its diff.

> Let me know if you have any preliminary thoughts on the other work I mentioned (or if you're aware of any work I'd be duplicating). 

See https://github.com/pyenv/pyenv/issues/1737#issuecomment-794592631

chipx86:
> If you're asking about how to apply #2020 to other versions, see its diff.

Yeah, sorry, I was running on fumes last night, wasn't paying enough attention to that change. I'll port that over in a new pull request and drop this one.

> See [#1737 (comment)](https://github.com/pyenv/pyenv/issues/1737#issuecomment-794592631)

Thanks. I came across the same resolution here, though it only fixes one of the build problems. I'll put up a pull request once I've ironed out the remaining issues, and we can explore the approach further there.

PR #2158, opened 2021-11-13
Add micropython 1.16 and 1.17

scop:

### Prerequisite

### Description

### Tests


plugins/python-build/share/python-build/micropython-1.16
```diff
@@ -0,0 +1,4 @@
+has_tar_xz_support \
+  && { install=install_package; src="https://micropython.org/resources/source/micropython-1.16.tar.xz#48271fb5da97efc22325c3fc1d87ba45b32bab25673206d8d8136c25e4ff29a9"; } \
+  || { install=install_zip; src="https://micropython.org/resources/source/micropython-1.16.zip#427adc00d2a99d102e4ce46d9eb4933d168e863f859bb3ce5b84735f0899c1ed"; }
+$install micropython-1.16 "$src" micropython
```
native-api, line 4: The `a && b || c` construct won't catch a failure in `b` in `-e` mode.
This needs to be rewritten as `if`/`then`/`else`.
(Same for other files.)

native-api (changes requested):

plugins/python-build/share/python-build/micropython-1.16
native-api: Scratch that. That is only relevant if `b` can fail -- and in this case, it can't since you only do static variable assignments there.

native-api (approved):

Issue #2085, opened 2021-10-01
Simplify Pyenv shell setup and Lift session restart requirement after installing Pyenv


native-api:
Seeing that the requirement to restart the session since 2.0 (due to some setup code needing to run at login shell's rather than interactive shell's startup) poses an inconvenience to users, we've been considering ways to lift that requirement in a way that won't break things.

So far, the best idea is:

* `eval "$(pyenv init -)"` adds shims to PATH if they are not already there
* `eval "$(pyenv init --path)"` adds shims to PATH unconditionally and deletes any other shims entries in PATH

---

This will:

* allow a user to run just `export PATH=~/.pyenv/bin:"$PATH"; eval "$(pyenv init -)"` as before after installing Pyenv and be able to use it right away
  * while still allowing to add a PATH entry at a different moment -- e.g. in distros that prepend `~/.local/bin` to PATH towards the end of profile, we need to add a PATH entry after that
* There will be no need to run `eval "$(pyenv init --path)"` in cases where it runs before `eval "$(pyenv init -)"`
  * for Zsh in MacOS at least
  * in CI where a user installs Pyenv themsevles

The drawbacks are:

* in Linux, until the user restarts the login session:
  * they will have to run the above commands (`export PATH=~/.pyenv/bin:"$PATH"; eval "$(pyenv init -)"`) by hand in every terminal to be able to use Pyenv in it
  * processes outside the updated terminal sessions won't be able to find Pyenv (but before 2.0, they were never able to find Pyenv at all -- so that's not a regression)
* If a user wishes to intentionally add shims to PATH multiple times (e.g. in some layered setup), we'll have to add a switch to `--path` to not munge a previous entry and instruct them to use it. Maybe we'll also have to add a switch to `init -` to not add a PATH entry since otherwise it may be impossible to ensure that the PATH entry is only added at the right moments.

Is it still worth it? What do you think?

native-api:
> There may be folks who would try to run pyenv on Jenkins.

Isn't Jenkins an Init service? Then it wouldn't inherit configuration from profile automatically, it needs to somehow apply it on its own.

native-api:
**To give you an idea of whether it'll be more or less complicated with the change:** in Ubuntu's case, the new configuration would be:

```
sed -Ei -e '/^([^#]|$)/ {a \
export PYENV_ROOT="$HOME/.pyenv"
a \
export PATH="$PYENV_ROOT/bin:$PATH"
a \
[...]

PR #2161, opened 2021-11-15
update readme after pyenv changed bashrc script

lgabs:
Fix readme reference when configuring a shell's environment for pyenv, as discussed in [issue 112](https://github.com/pyenv/pyenv-installer/issues/112) of pyenv-installer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ easy to fork and contribute any changes back upstream.
          ' -e ':a' -e '$!{n;ba};}' ~/.profile
          echo 'eval "$(pyenv init --path)"' >>~/.profile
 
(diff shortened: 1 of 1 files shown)


native-api:
The proposed change is incorrect. While it may appear to work, you actually lose the `pyenv` shell function and associated functionality: completion and automatic `rehash` at startup, `shell` and `activate`/`deactivate` subcommands.

The thing that those proposing it are missing is you need to relogin after installation -- as the README section following the shell setup instructions says.

Lifting the restart requirement is being discussed in https://github.com/pyenv/pyenv/issues/2085.

lgabs:
Sorry for incorrect propose, I've thought my pyenv only worked after that change, but I may be due to relogin after installation as well (I'm new to pyenv)

PR #2182, opened 2021-12-04
Fix 3.6.15 build on macOS

J-M0:
Building Python 3.6.15 is broken on macOS due to OpenSSL being improperly configured. This PR fixes that and also adds tests to make sure macOS builds work properly across Python versions.


.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: macos-10.15
+    steps:
+      - uses: actions/checkout@v2
+      # Normally, we would use the superbly maintained...
+      # - uses: actions/setup-python@v2
+      #   with:
+      #    python-version: ${{ matrix.python-version }}
+      # ... but in the repo, we want to test pyenv builds on macOS
+      - run: |
+          brew install autoconf openssl@1.1 pkg-config readline
```
native-api, line 25: Is autoconf needed? Python releases include already-generated configure files.
What about other dependencies as per https://github.com/pyenv/pyenv/wiki#suggested-build-environment?

Newer release lines should be tested in newer MacOS and with OpenSSL 3 but this is good enough for a start I guess.

J-M0: Those are what Homebrew listed as dependencies, so I just copied them in here. 

https://github.com/Homebrew/homebrew-core/blob/b3716bffc17f953c487472d0e5a0e9ae230e7094/Formula/pyenv.rb#L24-L33

If you prefer what’s on the wiki I can easily change it.

native-api: We need to test the suggested build environment so please do.

J-M0: All set

J-M0: Oh, there something I noticed but forgot to mention. It looks like even though the Homebrew versions of sqlite3, xz, and zlib are installed, pyenv is using the native macOS versions. It makes me wonder if the wiki should be more specific about what versions of macOS require those dependencies.

native-api: Sure, it would help if someone looked into this. I know that previous MacOS versions didn't have Zlib -- this is probably the case with other libs, too. Maybe for non-EOL MacOS versions,  some of these are no longer needed.

PR #2183, opened 2021-12-06
Add CPython 3.10.1

richin13:
### Description

https://www.python.org/downloads/release/python-3101/


WayneLambert:
I'm delighted to see this version of cPython added. Is there any plans to push a new release out to Homebrew so that it can be listed as an option on the `penv install --list` following a `brew upgrade pyenv`?

native-api:
@WayneLambert you can install a HEAD version of Pyenv until this change makes it into a release: `brew install pyenv --head`.

WayneLambert:
@native-api Thanks, I didn't know about the `--head` option, so cheers for that.

PR #2189, opened 2021-12-17
Fix Python 3.6.15 compilation on macOS arm64/M1.

chipx86:
### Description

This change ports several established patches to the Python 3.6.15 build, enabling compilation on arm64/Apple M1 architectures:

1. `0001-Detect-arm64-in-configure.patch` — Updates configure to detect arm64 architectures (port of an existing pyenv patch for 2.7.18).

2. `0002-bpo-36231-Support-building-on-macOS-without-usr-incl.patch` — Adds macOS SDK root computation logic for determining include paths (port of existing Python patches introduced in 2.7.17 and 3.7.4).

3. `0003-Fix-macOS-_tkinter-use-of-Tck-Tk-in-Library-Framewor.patch` — Fixes Tcl/Tk support on macOS (port of an existing pyenv patch for 2.7.18).

4. `0004-Port-ctypes-and-system-libffi-patches-for-arm64-macO.patch` — Fixes system `ffi.h`/`libffi` path determination and usage and enables calling of variadic functions, fixing ctypes support (consolidated port of existing pyenv patches for 2.7.18 that iterate on this logic).

5. `0005-BPO-41100-Support-macOS-11-when-building-GH-21113.patch` — Updates Darwin version checks to handle macOS 11's major version bump (port of Python patches introduced in 3.7.0 and 3.9.0).

6. `0006-bpo-41100-fix-_decimal-for-arm64-Mac-OS-GH-21228.patch` — Adds arm64 to the list of allowable architectures for the `decimal` module (port of Python patch introduced in 3.8.10).

GNU coreutils is also disabled through when building on arm64, through `arm64_apple_disable_homebrew_coreutils`, in order to avoid build issues caused by a difference in the results from `uname`.

### Tests

All existing bats unit tests pass.

Performed a clean build of the patched version of Python 3.6.15 on a 2021 M1 MacBook Pro running macOS 12.0.1, a 2017 Intel MacBook Pro running macOS 10.14.6, and CentOS 8 (Intel).

In each case, I verified that `ctypes` was working. I also ran through full installation tests and unit tests for all of our products (comprising dozens of pure Python and native C packages, and many thousands of unit tests).

### DIFF
--- a/plugins/python-build/share/python-build/patches/3.6.15/Python-3.6.15/0001-Detect-arm64-in-configure.patch
+++ b/plugins/python-build/share/python-build/patches/3.6.15/Python-3.6.15/0001-Detect-arm64-in-configure.patch
@@ -1,0 +1,41 @@
+From 002501946bad91a308f37a09204393c172f03c3e Mon Sep 17 00:00:00 2001
+From: Takumi Sueda <puhitaku@gmail.com>
+Date: Sat, 11 Sep 2021 16:50:14 +0900
+Subject: [PATCH 2/6] Detect arm64 in configure
+
+---
+ configure    | 3 +++
+ configure.ac | 3 +++
+ 2 files changed, 6 insertions(+)
+
+diff --git a/configure b/configure
+index e39c16eee2..8dc1fc7595 100755
+--- a/configure
++++ b/configure
+@@ -9230,6 +9230,9 @@ fi
+     	ppc)
+     		MACOSX_DEFAULT_ARCH="ppc64"
+     		;;
++    	arm64)
++    		MACOSX_DEFAULT_ARCH="arm64"
++    		;;
+     	*)
+     		as_fn_error $? "Unexpected output of 'arch' on OSX" "$LINENO" 5
+     		;;
+diff --git a/configure.ac b/configure.ac
+index cf280506bd..34846a7df3 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -2435,6 +2435,9 @@ case $ac_sys_system/$ac_sys_release in
+     	ppc)
+     		MACOSX_DEFAULT_ARCH="ppc64"
+     		;;
++    	arm64) 
++    		MACOSX_DEFAULT_ARCH="arm64" 
++    		;;
+     	*)
+     		AC_MSG_ERROR([Unexpected output of 'arch' on OSX])
+     		;;
+-- 
+2.30.1 (Apple Git-130)
+
--- a/plugins/python-build/share/python-build/patches/3.6.15/Python-3.6.15/0002-bpo-36231-Support-building-on-macOS-without-usr-incl.patch
+++ b/plugins/python-build/share/python-build/patches/3.6.15/Python-3.6.15/0002-bpo-36231-Support-building-on-macOS-without-usr-incl.patch
@@ -1,0 +1,93 @@
+From c7302116573d853d3181133477d9d0e4d4d3abfd Mon Sep 17 00:00:00 2001
+From: Ned Deily <nad@python.org>
+Date: Tue, 18 Jun 2019 16:28:13 -0400
+Subject: [PATCH] bpo-36231:  Support building on macOS without /usr/include
+ (GH-13773) (GH-14208)
+
+---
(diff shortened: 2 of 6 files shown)


plugins/python-build/share/python-build/patches/3.6.15/Python-3.6.15/0001-Detect-arm64-in-configure.patch
```diff
@@ -0,0 +1,41 @@
...
+index cf280506bd..34846a7df3 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -2435,6 +2435,9 @@ case $ac_sys_system/$ac_sys_release in
+     	ppc)
+     		MACOSX_DEFAULT_ARCH="ppc64"
+     		;;
++    	arm64) 
++    		MACOSX_DEFAULT_ARCH="arm64" 
++    		;;
```
native-api, line 35: In the light of https://github.com/pyenv/pyenv/pull/2190/files#r771388183, is this needed with stock Apple's `uname`? It reports `arm` instead of `arm64`.

(same goes for `configure` above)

chipx86: Yeah, this part of `configure` is explicitly calling `/usr/bin/arch` and checking the result. Here's the original code:

```shell
    AC_RUN_IFELSE([AC_LANG_SOURCE([[
    #include <unistd.h>
    int main(int argc, char*argv[])
    {
      if (sizeof(long) == 4) {
    	  return 0;
      } else {
      	  return 1;
      }
    }
    ]])],[ac_osx_32bit=yes],[ac_osx_32bit=no],[ac_osx_32bit=yes])

    if test "${ac_osx_32bit}" = "yes"; then
    	case `/usr/bin/arch` in
    	i386)
    		MACOSX_DEFAULT_ARCH="i386"
    		;;
    	ppc)
    		MACOSX_DEFAULT_ARCH="ppc"
    		;;
    	*)
    		AC_MSG_ERROR([Unexpected output of 'arch' on OSX])
    		;;
    	esac
    else
    	case `/usr/bin/arch` in
    	i386)
    		MACOSX_DEFAULT_ARCH="x86_64"
    		;;
    	ppc)
    		MACOSX_DEFAULT_ARCH="ppc64"
    		;;
    	*)
    		AC_MSG_ERROR([Unexpected output of 'arch' on OSX])
    		;;
    	esac

    fi
```

plugins/python-build/share/python-build/3.6.15
```diff
@@ -3,7 +3,7 @@ prefer_openssl11
...
 install_package "readline-8.1" "https://ftpmirror.gnu.org/readline/readline-8.1.tar.gz#f8ceb4ee131e3232226a17f51b164afc46cd0b9e6cef344be87c65962cb82b02" mac_readline --if has_broken_mac_readline
 if has_tar_xz_support; then
-  install_package "Python-3.6.15" "https://www.python.org/ftp/python/3.6.15/Python-3.6.15.tar.xz#6e28d7cdd6dd513dd190e49bca3972e20fcf455090ccf2ef3f1a227614135d91" standard verify_py36 copy_python_gdb ensurepip
+  install_package "Python-3.6.15" "https://www.python.org/ftp/python/3.6.15/Python-3.6.15.tar.xz#6e28d7cdd6dd513dd190e49bca3972e20fcf455090ccf2ef3f1a227614135d91" arm64_apple_disable_homebrew_coreutils standard verify_py36 copy_python_gdb ensurepip
 else
-  install_package "Python-3.6.15" "https://www.python.org/ftp/python/3.6.15/Python-3.6.15.tgz#54570b7e339e2cfd72b29c7e2fdb47c0b7b18b7412e61de5b463fc087c13b043" standard verify_py36 copy_python_gdb ensurepip
+  install_package "Python-3.6.15" "https://www.python.org/ftp/python/3.6.15/Python-3.6.15.tgz#54570b7e339e2cfd72b29c7e2fdb47c0b7b18b7412e61de5b463fc087c13b043" arm64_apple_disable_homebrew_coreutils standard verify_py36 copy_python_gdb ensurepip
```
native-api, line 8: Same as https://github.com/pyenv/pyenv/pull/2190#discussion_r771388183 (better disciss it there)

PR #2190, opened 2021-12-17
Fix Python 3.7.12 compilation on macOS arm64/M1.

chipx86:
### Description

This change ports pyenv and Python patches to 3.7.12 to enable the `ctypes `and `decimal` modules to compile, and to fix a compilation issue with GNU coreutils.

While Python 3.7.12 itself compiles on arm64/M1, both of these modules fail to compile, due to missing support for locating system libffi and due to architecture gate-keeping. These issues have been fixed in newer releases of Python, and in other pyenv patch bundles.

The following patches are provided:

1. `0001-Port-ctypes-and-system-libffi-patches-for-arm64-macO.patch` — Fixes system `ffi.h`/`libffi` path determination and usage and enables calling of variadic functions, fixing ctypes support (consolidated port of existing pyenv patches for 2.7.18 that iterate on this logic).

2. `0002-bpo-41100-fix-_decimal-for-arm64-Mac-OS-GH-21228.patch` — Adds arm64 to the list of allowable architectures for the `decimal` module (port of Python patch introduced in 3.8.10).

GNU coreutils is also disabled through when building on arm64, through `arm64_apple_disable_homebrew_coreutils`, in order to avoid build issues caused by a difference in the results from `uname`.

### Tests

All existing bats unit tests pass.

Performed a clean build of the patched version of Python 3.7.12 on a 2021 M1 MacBook Pro running macOS 12.0.1, a 2017 Intel MacBook Pro running macOS 10.14.6, and CentOS 8 (Intel).

In each case, I verified that ctypes was working. I also ran through full installation tests and unit tests for all of our products (comprising dozens of pure Python and native C packages, and many thousands of unit tests).

### DIFF
--- a/plugins/python-build/share/python-build/patches/3.7.12/Python-3.7.12/0001-Port-ctypes-and-system-libffi-patches-for-arm64-macO.patch
+++ b/plugins/python-build/share/python-build/patches/3.7.12/Python-3.7.12/0001-Port-ctypes-and-system-libffi-patches-for-arm64-macO.patch
@@ -1,0 +1,312 @@
+From: Christian Hammond <christian@beanbaginc.com>
+Date: Wed, 15 Dec 2021 23:12:36 -0800
+Subject: Port ctypes and system libffi patches for arm64/macOS 10.15+ to Python 3.7.12
+
+This ports the following ctypes and libffi pyenv patches for Python
+2.7.18 to Python 3.7.12:
+
+* `0004-Use-system-libffi-for-Mac-OS-10.15-and-up.patch`
+* `0005-ctypes-use-the-correct-ABI-for-variadic-functions.patch`
+* `0006-ctypes-probe-libffi-for-ffi_closure_alloc-and-ffi_pr.patch`
+
+These patches enable use of system libffi (fixing a broken `ctypes`
+module on arm64 targets) and enable calling variadic functions on arm64.
+They've been combined from patches port from Homebrew to pyenv by Takumi
+Sueda, updated to work on the Python 3.7.12 codebase.
+
+The Homebrew patches are themselves backports of changes in Python 3.9
+and 3.10. That patch can be found at:
+
+https://github.com/Homebrew/formula-patches/blob/master/python/3.8.7.patch
+
+diff --git a/Doc/library/ctypes.rst b/Doc/library/ctypes.rst
+index 715d595b24..7743144978 100644
+--- a/Doc/library/ctypes.rst
++++ b/Doc/library/ctypes.rst
+@@ -1551,6 +1551,13 @@ They are instances of a private class:
+       value usable as argument (integer, string, ctypes instance).  This allows
+       defining adapters that can adapt custom objects as function parameters.
+ 
++   .. attribute:: variadic
++
++      Assign a boolean to specify that the function takes a variable number of
++      arguments.   This does not matter on most platforms, but for Apple arm64
++      platforms variadic functions have a different calling convention than
++      normal functions.
++
+    .. attribute:: errcheck
+ 
+       Assign a Python function or another callable to this attribute. The
+diff --git a/Lib/test/test_unicode.py b/Lib/test/test_unicode.py
+index 4ebd82d3e0..7b73c190b6 100644
+--- a/Lib/test/test_unicode.py
++++ b/Lib/test/test_unicode.py
+@@ -2458,11 +2458,14 @@ class CAPITest(unittest.TestCase):
+     def test_from_format(self):
+         support.import_module('ctypes')
+         from ctypes import (
++            c_char_p,
+             pythonapi, py_object, sizeof,
+             c_int, c_long, c_longlong, c_ssize_t,
+             c_uint, c_ulong, c_ulonglong, c_size_t, c_void_p)
+         name = "PyUnicode_FromFormat"
+         _PyUnicode_FromFormat = getattr(pythonapi, name)
++        _PyUnicode_FromFormat.argtypes = (c_char_p,)
++        _PyUnicode_FromFormat.variadic = True
+         _PyUnicode_FromFormat.restype = py_object
+ 
+         def PyUnicode_FromFormat(format, *args):
+diff --git a/Modules/_ctypes/_ctypes.c b/Modules/_ctypes/_ctypes.c
+index dd0c61fd8a..79137e1dc7 100644
+--- a/Modules/_ctypes/_ctypes.c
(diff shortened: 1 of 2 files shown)


plugins/python-build/share/python-build/3.7.12
```diff
@@ -4,7 +4,7 @@ export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
...
 install_package "readline-8.1" "https://ftpmirror.gnu.org/readline/readline-8.1.tar.gz#f8ceb4ee131e3232226a17f51b164afc46cd0b9e6cef344be87c65962cb82b02" mac_readline --if has_broken_mac_readline
 if has_tar_xz_support; then
-  install_package "Python-3.7.12" "https://www.python.org/ftp/python/3.7.12/Python-3.7.12.tar.xz#f77bf7fb47839f213e5cbf7827281078ea90de7e72b44f10d7ef385ea8c43210" standard verify_py37 copy_python_gdb ensurepip
+  install_package "Python-3.7.12" "https://www.python.org/ftp/python/3.7.12/Python-3.7.12.tar.xz#f77bf7fb47839f213e5cbf7827281078ea90de7e72b44f10d7ef385ea8c43210" arm64_apple_disable_homebrew_coreutils standard verify_py37 copy_python_gdb ensurepip
 else
-  install_package "Python-3.7.12" "https://www.python.org/ftp/python/3.7.12/Python-3.7.12.tgz#33b4daaf831be19219659466d12645f87ecec6eb21d4d9f9711018a7b66cce46" standard verify_py37 copy_python_gdb ensurepip
+  install_package "Python-3.7.12" "https://www.python.org/ftp/python/3.7.12/Python-3.7.12.tgz#33b4daaf831be19219659466d12645f87ecec6eb21d4d9f9711018a7b66cce46" arm64_apple_disable_homebrew_coreutils standard verify_py37 copy_python_gdb ensurepip
```
native-api, line 9: This workaround only "fixes" the case where the user had added `gnubin` from Homebrew Coreutils to their PATH -- which is technically an unsupported environment.

As such, I wasn't too happy to accept it in the first place and only did it with an idea 
that it would only be applied to a few select 3.8 and 3.9 releases that do officially support Apple M1 and were only broken with GNU Coreutils because those releases sport a too old `config,sub`.

As such, do we really need it here, too?

native-api: @pyenv/pyenv-core-maintainers Given what is said above, do you think we can drop this workaround altogether? Do we need to consider it a breaking change if yes?

chipx86: Depends on what configurations you want to support out of the box. I'm okay with whatever you prefer.

I can confirm that 3.6.15 and 3.7.12 contain the old `config.sub`, and will break with `gnubin` without the patch. My previous attempt in #2157 attempted to backport the relevant parts of `config.sub` to fix architecture detection (I wasn't aware of `arm64_apple_disable_homebrew_coreutils` until it came up in that change).

Because I depend on coreutils, my preference would be to ease building with `gnubin` in the path. I can confirm that these do build with and without `gnubin` on M1 and Intel Macs, but I can also see the reason to want to avoid any extra workarounds or configurations. (At least `arm64_apple_disable_homebrew_coreutils` is a consolidated workaround, unlike my previous approach.)

I'll go with whatever approach you prefer. I don't know how common the mix of Python 3.6/3.7, M1, and coreutils may be out there. We've hit it during setup, partly because we do development on Mac but deploy on Linux (coreutils eases compatibility concerns there). If not addressed during build, it might be worth documenting a workaround for anyone who may hit this (we'd plan to do it for our project and internal docs).

For clarification, what's considered a breaking change in this context?

native-api: > Because I depend on coreutils

GNU Coreutils are available on PATH even without adding `gnubin`, with the `g` prefix. Whatever you depend on should be achievable by looking for an executable with `g` first, e.g.: https://github.com/pyenv/pyenv/blob/943c5f995d0ecdc957ca365aa79a9ce3294652bf/libexec/pyenv#L32

chipx86: It's more about compatibility with scripts (first-party, third-party, or customer-provided for debugging) that assume they'll be running on a typical Linux environment. Achieving a closer environment to that without modifying the scripts. Pretty important for our use cases.

native-api: Then I wonder why you run them in an environment different than what they're designed for -- that sounds counterproductive, you're going to lose time from both Linux-aware and MacOS-aware tools breaking due to differences from either standard.

Just to let you know, we've rejected a similar workaround for GNU binutils: https://github.com/pyenv/pyenv/pull/2060#issuecomment-917723332

chipx86: I'm fine removing that workaround. I added it in based on your feedback on #2157, but it's your project, and I'll make the changes you require to get Python 3.7.12 support in in some form.

Regarding the differences in behavior, I totally understand what you mean, but this is an area that we know very well. We're old-school Linux developers who have been in this game a long time. We've built tools that were once standard on most autoconf/automake-based build systems, wrote build architecture for VMware that allowed for wide cross-Linux compatibility (along with writing most of the rest of VMware's Linux products), and helped build specs and software still in use on most Linux desktops. Though, I do agree in principal with what you're saying, and would perhaps issue the same advice to others.

But all that's orthogonal to the patch. If `arm64_apple_disable_homebrew_coreutils` usage is a blocker, I'll remove it, and we'll just document it on our end.

native-api: With such a high-profile background, you'll surely be able to live fine without the kludge in Pyenv!
(That's what I was worried about -- if we break our users' workflow, however unsupported, we'd better provide a workable alternative for them if we'd like them to continue using our product.)

> If arm64_apple_disable_homebrew_coreutils usage is a blocker, I'll remove it,

Yes, please do.
And we can proceed to remove it from elsewhere as well.
Since it only makes a difference in an unsupported setup and we never officially documented this behavior (since it was merely a side effect of a bugfix for specific CPython releases), we don't need to consider it a breaking change. Just notifying authors of the 2.7 and coreutils' PRs should be enough (and the reverting PR will be in the next release notes).

chipx86: Sounds like a plan! I'll get these updated in the next few hours.

chipx86:
Thanks for working with me on this :) Have a good rest of your night/day!

PR #2212, opened 2022-01-08
Revert "Disable coreutils on M1 Apple Silicon with arm64 (#2020)"

native-api:
This reverts commit 90d0d20508a91e7ea1e609e8aa9f9d1a28bb563e.

### Prerequisite

### Description

After further consideration, we've decided to remove this workaround:
* It only has an effect if the user has added `gnubin` from Homebrew Coreutils to PATH which is an unsupported setup
* It was intended to be applied only to a few select 3.8 and 3.9 versions that officially support Apple Silicon and only fail with Homebrew Coreutils in PATH because they have `config.*` from a too old version of Autoconf that doesn't support the Arm64 arch -- but
  * CPython devs [didn't actually fix the problem in 3.10, either, only in 3.11](https://github.com/pyenv/pyenv/pull/2157#issuecomment-968055387), so we'd need to apply it to all 3.10 releases, too
  * users started pushing this workaround into other unrelated branches because they were using the above unsupported setup. See https://github.com/pyenv/pyenv/pull/2190#pullrequestreview-835221952 for discussion.

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -112,13 +112,6 @@ is_mac() {
   [ $# -eq 0 ] || [ "$(osx_version)" "$@" ]
 }
 
-is_arm64_apple() {
-  if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]; then
-    return 0
-  fi
-  return 1
-}
-
 #  9.1  -> 901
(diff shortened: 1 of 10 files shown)


native-api:
@yackrru Be advised that we're removing this workaround, see above for the reasons.

PR #2215, opened 2022-01-14
Update the `$PYENV_ROOT/bin` to `.../shims`

lane-eb:
There is no `bin` folder under `$PYENV_ROOT/`. It should be `$PYENV_ROOT/shims. 
If keep it as `bin`, the `rbenv` will not be set correctly for fish user.

Screen Shot 2022-01-14 at 5 04 37 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ easy to fork and contribute any changes back upstream.
         
         ~~~ fish
         set -Ux PYENV_ROOT $HOME/.pyenv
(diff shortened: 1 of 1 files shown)


native-api:
If you've installed Pyenv system-wide, you don't need to add `bin` to PATH. There's only 1 common use case of this, installing via Homebrew, and shell setup instructions reflect that. `shims` are added to PATH by `eval "$(pyenv init --path)"`.

PR #2218, opened 2022-01-15
Add CPython 3.10.2

SnoopJeDi:
Adds [CPython 3.10.2](https://www.python.org/downloads/release/python-3102/) to pyenv

### DIFF
--- a/plugins/python-build/share/python-build/3.10.2
+++ b/plugins/python-build/share/python-build/3.10.2
@@ -1,0 +1,10 @@
+#require_gcc
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1k.tar.gz#892a0875b9872acd04a9fde79b1f943075d5ea162415de3047c327df33fbaee5" mac_openssl --if has_broken_mac_openssl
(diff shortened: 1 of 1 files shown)


plugins/python-build/share/python-build/3.10.2
```diff
@@ -0,0 +1,10 @@
+#require_gcc
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1k.tar.gz#892a0875b9872acd04a9fde79b1f943075d5ea162415de3047c327df33fbaee5" mac_openssl --if has_broken_mac_openssl
+install_package "readline-8.0" "https://ftpmirror.gnu.org/readline/readline-8.0.tar.gz#e339f51971478d369f8a053a330a190781acb9864cf4c541060f12078948e461" mac_readline --if has_broken_mac_readline
```
native-api, line 5: This could be changed to the latest release, 8.1.2

plugins/python-build/share/python-build/3.10.2
```diff
@@ -0,0 +1,10 @@
+#require_gcc
+prefer_openssl11
```
native-api, line 2: Do we still need to prefer OpenSSL 1.1?
[AFAICS](https://docs.python.org/3/whatsnew/changelog.html), 3.10 supports 3.0.

native-api:
Superceded by https://github.com/pyenv/pyenv/pull/2219

PR #2219, opened 2022-01-15
CPython 3.9.10 and 3.10.2

nedbat:


native-api:
See https://github.com/pyenv/pyenv/pull/2218 on comments about 3.10.2

nedbat:
I'm sorry, I don't know how to decide if we can upgrade the dependencies.

native-api:
You can test if all the tests pass with OpenSSL 3, for both versions (they should).
About Readline, I'm pretty sure we can upgrade it (it's only used in MacOS without Homebrew anyway).

fredrikaverpil:
Ping @tiran - sorry for bringing you into this conversation seemingly randomly 😅 would you say it's wise for pyenv to provide CPython 3.10.2 compiled against openssl 3?

native-api:
We'd like to do the same for 3.9.10+ since [they backported OpenSSL 3 support there, too](https://bugs.python.org/issue38820).

tiran:
Python's OpenSSL 3.0 is preliminary and experimental. I recommend against using Python with OpenSSL 3.0 in production environments. There are known performance regressions and problems. Also OpenSSL 3.0 is new and a lot of code has changed.

native-api:
@tiran Thanks, that's a very valuable info!
Is there any kind of milestone in plans so that we know when to switch?

tiran:
No ETA yet. Somebody needs to put in the work and review CPython's use of OpenSSL APIs. See also https://mail.python.org/archives/list/python-dev@python.org/thread/ATO4DM6QYZGLSGGDZ3TRN5X3QDD5OHOE/

fredrikaverpil:
@tiran thank you for both this input and your invaluable contributions to python. 🙏

SnoopJeDi:
Does `pyenv` have a policy about which versions of these dependencies should be included in a build recipe? I wondered about this when I put #2218 together, but erred on the side of keeping what was used for `3.10.1`. This would be a helpful thing to document in a `CONTRIBUTING.md` file.

fredrikaverpil:
@SnoopJeDi I agree with you. +1 on having some sort of guidance here and I am not aware of one. Ping @anton-petrov ☝️

anton-petrov:
@fredrikaverpil We don't have any policy about the versions of third party libraries that are used for building CPython yet...
But, as I see, there are problems with opensl at the moment, suggest compiling with OpenSSL 1.1.1, not a version 3.
@SnoopJeDi 
> This would be a helpful thing to document in a CONTRIBUTING.md file.
That's a good idea! We need to do that in the near future.

anton-petrov:
@tiran 
> Python's OpenSSL 3.0 is preliminary and experimental. I recommend against using Python with OpenSSL 3.0 in production environments. There are known performance regressions and problems. Also OpenSSL 3.0 is new and a lot of code has changed.

Thanks for this important information!

PR #2253, opened 2022-02-22
pypy 7.3.8 02/20/2022 release

dand-oss:

### Prerequisite
* [x ] Please consider implementing the feature as a hook script or plugin as a first step.
* [x ] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
* [x ] My PR addresses the following pyenv issue (if any)

### Description
- [ x] Here are some details about my PR

### Tests
- [x ] My PR adds the following unit tests (if any)

### DIFF
--- a/plugins/python-build/share/python-build/pypy2.7-7.3.8
+++ b/plugins/python-build/share/python-build/pypy2.7-7.3.8
@@ -1,0 +1,70 @@
+VERSION='7.3.8'
+PYVER='2.7'
+
+# https://www.pypy.org/checksums.html
+aarch64_hash=ca1f8d3146c83002ee97615906b0930e821297dcce3063b5b28933a0690ef298
+linux32_hash=7c84f173bbcd73d0eb10909259d11b5cc253d4c6ea4492e6da8f2532df9b3da5
+linux64_hash=1f2e84fb539ffce233c34769d2f11647955f894be091e85419e05f48011e8940
+osx64_hash=e5c1ff39ad9916ea23e3deb8012fe42367b6b19284cf13b1a1ea2b2f53a43add
+s390x_hash=b4ae4e708ba84602d976ad6ae391ef2eef4b1896d831b8f2b2ec69927dd92014
+win64_hash=806a29a6c5550b1e669d8870683d3379138d3d43eb1e07bdf26d65a0691265f2
+
+### end of manual settings - following lines same for every download
+
+basesrc="pypy${PYVER}-${VERSION}-src"
+basever="pypy${PYVER}-v${VERSION}"
+baseurl="https://downloads.python.org/pypy/${basever}"
+
+# defaults
+cmd='install_package'  # use bz2
+pkg="$(pypy_architecture 2>/dev/null || true)" # assume matches
+url="${baseurl}-${pkg}.tar.bz2" # use bz2
+hash='' # undefined
+
+case "${pkg}" in
+'linux-aarch64' )
+  hash="${aarch64_hash}"
+  url="${baseurl}-aarch64.tar.bz2" # diff url
+  ;;
+'linux' )
+  hash="${linux32_hash}"
+  pkg='linux32'  # package name revised
+  url="${baseurl}-${pkg}.tar.bz2"  # new url
+  ;;
+'linux64' )
+  hash="${linux64_hash}"
+  ;;
+'osx64' )
+  if require_osx_version "10.13"; then
+    hash="${osx64_hash}"
+  else
+    { echo
+      colorize 1 "ERROR"
+      echo ": The binary distribution of PyPy is not available for $(pypy_architecture 2>/dev/null || true), OS X < 10.13."
+      echo "try '${basesrc}' to build from source."
+      echo
+    } >&2
+    exit 1
+  fi
+  ;;
+'s390x' )
+  hash="${s390x_hash}"
+  ;;
+'win64' )
+  hash="${win64_hash}"
+  cmd='install_zip'  # diff command
+  url="${baseurl}-${pkg}.zip"  # zip rather than bz2
+  ;;
+* )
+  { echo
+    colorize 1 "ERROR"
+    echo ": The binary distribution of PyPy is not available for $(pypy_architecture 2>/dev/null || true)."
+    echo "try '${basesrc}' to build from source."
+    echo
+  } >&2
+  exit 1
+  ;;
+esac
+
+# install
+"${cmd}" "${basever}-${pkg}" "${url}#${hash}" 'pypy' "verify_py${PYVER//./}" 'ensurepip'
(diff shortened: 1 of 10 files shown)


plugins/python-build/share/python-build/pypy3.7-7.3.8
```diff
@@ -0,0 +1,44 @@
...
+PYVER='3.7'
+# https://www.pypy.org/checksums.html
+# https://downloads.python.org/pypy
+
+case "$(pypy_architecture 2>/dev/null || true)" in
+"linux" )
+  install_package "pypy${PYVER}-v${VERSION}linux32" "https://downloads.python.org/pypy/pypy${PYVER}-v${VERSION}linux32.tar.bz2#38429ec6ea1aca391821ee4fbda7358ae86de4600146643f2af2fe2c085af839" "pypy" "verify_py${PYVER//./}" ensurepip
+  ;;
+"linux64" )
+  install_package "pypy${PYVER}-v${VERSION}linux64" "https://downloads.python.org/pypy/pypy${PYVER}-v${VERSION}linux64.tar.bz2#409085db79a6d90bfcf4f576dca1538498e65937acfbe03bd4909bdc262ff378" "pypy" "verify_py${PYVER//./}" ensurepip
```
native-api, line 11: dash before "linux" missing

plugins/python-build/share/python-build/pypy2.7-7.3.8-src
```diff
@@ -0,0 +1,10 @@
+VERSION='7.3.8'
+PYVER='2.7'
+# https://www.pypy.org/checksums.html
+# https://downloads.python.org/pypy
+# https://downloads.python.org/pypy
+
+#require_gcc
+prefer_openssl11
+install_package "openssl-1.1.1f" "https://www.openssl.org/source/openssl-1.1.1f.tar.gz#186c6bfe6ecfba7a5b48c47f8a1673d0f3b0e5ba2e25602dd23b629975da3f35" mac_openssl --if has_broken_mac_openssl
+install_package "pypy${PYVER}-v${VERSION}rc1-src" "https://downloads.python.org/pypy/pypy${PYVER}-v${VERSION}rc1-src.tar.bz2#96c9f5a85759cc92000064d3b32ce89748870b35a48e631f713be3f29bf64f3c" "pypy_builder" "verify_py${PYVER//./}" ensurepip
```
native-api, line 10: It's not rc1 anymore, is it?

native-api (commented):
Please double check the identifiers and URLs

plugins/python-build/share/python-build/pypy3.7-7.3.8
dand-oss: Thanks!  On it

dand-oss:
how about a test bash script which checks each url sources from the build scripts?

If I make one, where should we put it?  Any naming convention?

dand-oss:
Ok, refactored and added unit test
added test script which downloads all files and compares checksums
did pyenv install on compatible files

plugins/python-build/share/python-build/test_pypy378_download
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
```
native-api, line 1: A script like this is not useful to keep in the codebase since it does a one-off task.

You can see https://github.com/pyenv/pyenv/pull/1692 for scripts that I used to check URLs in all the scripts if you think you can write one that could have lasting value.

Scripts can be added to `python-build\scripts`.

plugins/python-build/share/python-build/pypy2.7-7.3.8
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+declare VERSION='7.3.8'
```
native-api, line 4: `declare`s are not necessary when defining regular variables IIRC.

native-api (commented):
A commendable effort! Just try to not reinvent the wheel since the existing code has accumulated fixes and workarounds for corner cases over the years.

plugins/python-build/share/python-build/pypy2.7-7.3.8
```diff
@@ -0,0 +1,78 @@
...
+declare PYVER='2.7'
+# https://www.pypy.org/checksums.html
+# https://downloads.python.org/pypy
+
+# associative array for urls
+declare basesrc="pypy${PYVER}-${VERSION}-src"
+declare basever="pypy${PYVER}-v${VERSION}"
+declare baseurl="https://downloads.python.org/pypy/${basever}"
+declare -A urlmap=(
```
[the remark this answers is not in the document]
dand-oss: I made the array so I could get the url and checksum from a unit test script WITHOUT installing - as installing requires matching "architecture".  I'll have to check 3.2 to see how this might be possible.

dand-oss:
undid declare
undid shebang
removed associative arrays
followed appoach used in graal python

plugins/python-build/share/python-build/pypy2.7-7.3.8-src
```diff
@@ -0,0 +1,15 @@
...
+
+# https://www.pypy.org/checksums.html
+hash=0cdad270c62d3ccc53cc87eeb069a6dc46acaf95521b584624bcd6697d94fa1c
+
+### end of manual settings - following lines same for every download
+
+ver="pypy${PYVER}-v${VERSION}-src"
+url="https://downloads.python.org/pypy/${ver}.tar.bz2"
+
+#require_gcc
```
native-api, line 12: No need to copypaste this long-commented-out code :wink:

native-api (commented):
Okay, the code is now good, but could you please move it into scripts?

I understand that you try to eliminate duplication which is normally a good thing -- but in this particular case, it has proven to be the preferrable way so far:
* Small details about the code -- supported architectures & OS versions, naming schemes -- are going to change over time (Pypy has had at least 2 major changes in the versioning scheme so far). A function inside Python-Build would have to get ever more complicated as a result.
* Old releases aren't expected to change -- so the scripts, albeit duplicated, do not really add to maintenance burden.
* Version-specific changes may be required later as quirks are found in specific releases. E.g. [PyPy 2.7 still doesn't support OpenSSL 1.1](https://doc.pypy.org/en/latest/release-pypy2.7-v5.4.1.html?highlight=openssl%203#pypy-5-4-1) so we'd need to prefer 1.0 instead. No-one has complained about it yet (presumably 'cuz PyPy is rarely compiled from source, even less so on Macs and with such old OpenSSL taken from Homebrew) but we'll surely accept a fix from an interested party.
* This makes for very easy support of new releases -- just copypaste the old script and adjust it as necessary if needed.

plugins/python-build/share/python-build/pypy2.7-7.3.8-src
dand-oss: move shared bash code from python-build back into individual scripts
remove commented bash code

burnpanck:
Cool, looking forward for this to make it into a release!

PR #2269, opened 2022-03-08
Version file read improvements

scop:
### Prerequisite
    * See https://github.com/rbenv/rbenv/pull/1393 

### Description

See individual commits for details.

### Tests

### DIFF
--- a/libexec/pyenv-version-file-read
+++ b/libexec/pyenv-version-file-read
@@ -5,7 +5,7 @@ set -e
 
 VERSION_FILE="$1"
 
-if [ -e "$VERSION_FILE" ]; then
+if [ -s "$VERSION_FILE" ]; then
   # Read the first non-whitespace word from the specified version file.
   # Be careful not to load it whole in case there's something crazy in it.
   IFS="${IFS}"$'\r'
--- a/libexec/pyenv-version-file-read
+++ b/libexec/pyenv-version-file-read
@@ -9,14 +9,13 @@ if [ -s "$VERSION_FILE" ]; then
   # Read the first non-whitespace word from the specified version file.
   # Be careful not to load it whole in case there's something crazy in it.
   IFS="${IFS}"$'\r'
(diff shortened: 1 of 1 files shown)


scop:
Unfortunately this doesn't cope with files without a trailing newline. Not sure if there's a concise enough way to accomplish that, leaving open for a bit.

The `-s` change could be applied separately, though.

libexec/pyenv-version-file-read
```diff
@@ -5,18 +5,17 @@ set -e
...
-
-  if [ -n "$versions" ]; then
-    IFS=":"
-    echo "${versions[*]}"
-    exit
-  fi
+  sep=
+  while read -n 1024 -r version _; do
+      [[ -z $version || $version == \#* ]] && continue
+      printf "%s%s" "$sep" "$version"
```
native-api, line 15: You never print a newline at the end?

native-api:
You may take a look at https://mywiki.wooledge.org/BashFAQ/005#Handling_newlines_.28or_lack_thereof.29_at_the_end_of_a_file . It deals with handling a (lack of) trailing newline.

libexec/pyenv-version-file-read
scop: This script is only executed and its output captured so that trailing newlines, if any, are discarded anyway, so it doesn't make a practical difference. It wouldn't hurt either, though, if that's preferred.

scop: This file is only ever read in contexts that discard trailing newlines, if any, anyway, so it makes no practical difference here. We could add one if that's preferred though.

scop: Added in the revised commit 569718495b79d6073e37619866a015d961145f34.

scop:
Ah, right. Revised commit 569718495b79d6073e37619866a015d961145f34 addresses the case without trailing newlines.

native-api:
Thanks! With combined effort, we pushed through and can finally tick this off! :slightly_smiling_face:

PR #2276, opened 2022-03-16
fix(dependence): bump openssl 1.1 to 1.1.1n for CPython 3.7 3.8 3.9

tuzi3040:

### Prerequisite

### Description

### Tests

---

OpenSSL of CPython 3.7 3.8 3.9 should also be bumped to 1.1.1n according to [here](https://github.com/python/cpython/blob/2de452f8bf2f78417e04bcf7919beb502c53a0e2/Misc/NEWS.d/3.9.11.rst).

### DIFF
--- a/plugins/python-build/share/python-build/3.10-dev
+++ b/plugins/python-build/share/python-build/3.10-dev
@@ -1,6 +1,6 @@
 #require_gcc
 prefer_openssl11
 export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
-install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" mac_openssl --if has_broken_mac_openssl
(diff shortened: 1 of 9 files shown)


plugins/python-build/share/python-build/3.10-dev
```diff
@@ -1,6 +1,6 @@
 #require_gcc
 prefer_openssl11
 export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
-install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-1.1.1n" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" mac_openssl --if has_broken_mac_openssl
```
native-api, line 4: Need to update checksums, too.

tuzi3040: Checksum is already updated in 3a95c969e122b1e0791cb8d6cc13063d3ae10857.

native-api: Oh. :blush:

joshfriend:
I was unaware this suggestion applied to older python versions as well, thanks!

Issue #1961, opened 2021-06-01
`eval "$(pyenv init -)"` message persisting even while working on zsh


zachcoleman:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x ] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [ x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [ x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [ x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [MacOS ] Platform information (e.g. Ubuntu Linux 16.04):
- [ amd64] OS architecture (e.g. amd64):
- [1.2.26] pyenv version:
- [ ] Python version:
- [ clang-1200.0.32.21] C Compiler information (e.g. gcc 7.3): 
- [ ] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

After following the README and adding the appropriate code to both .profile and .zprofile:
```shell
export PYENV_ROOT="$HOME/.pyenv"
export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init --path)"
```

The `zsh` startup still gives the message even though the pyenv shims are working:
```shell
# Load pyenv automatically by appending
# the following to your profile:

eval "$(pyenv init -)"

```

I can verify the shims are working like:
```
which python
> /Users/zachcoleman/.pyenv/shims/python
```

This is more of an annoyance than a critical bug. I believe both "profile" files have correct info in them. Please let me know if there is option to suppress this behavior.

native-api:
Please provide the diagnostic info as required by the issue template for us to be able to say anything.

[...]

PR #2296, opened 2022-03-24
CPython 3.9.12

saaketp:
Added CPython 3.9.12


amirsalaar:
Although I have the latest pyenv installed, it doesn't show `3.9.12` in the list of available python versions to me.

 I have installed pyenv through `brew` on my MBP M1.

native-api:
@amirsalaar check that you run the right Pyenv instance. You might have multiple, see [#1961](https://github.com/pyenv/pyenv/issues/1961#issuecomment-852419651).

PR #2326, opened 2022-04-14
Fix to add the pyenv shims to the PATH

marcandreuf:
For some reason, the init command needs the argument '--path' to add the shims path to the PATH var env.

### Prerequisite

### Description

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ easy to fork and contribute any changes back upstream.
          echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.zprofile
          echo 'eval "$(pyenv init --path)"' >> ~/.zprofile
       
-         echo 'eval "$(pyenv init -)"' >> ~/.zshrc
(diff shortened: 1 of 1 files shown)


marcandreuf:
Only updating documentation in the Readme file.

README.md
```diff
@@ -372,13 +372,9 @@ easy to fork and contribute any changes back upstream.
          ~~~ zsh
          echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.zprofile
          echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.zprofile
-         echo 'eval "$(pyenv init --path)"' >> ~/.zprofile
-         
-         echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.profile
```
marcandreuf, line 377: As per the ZSH documentation, the '.profile' file is not used in this shell. It would be used only in special cases where we could expect that the more experienced user will know if they need to add the content in that file. Having this as default configuration for ZSH would be just adding messy files to the system.

PR #2351, opened 2022-05-03
Add `:latest` syntax to documentation for the `install` command

hay:
It seems the `:latest` syntax (as added in commit #1831) is not documented anywhere in this file even though it's very useful so i thought it would be useful to add it.

### Prerequisite

### Description

### Tests

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -228,6 +228,14 @@ Then install the desired versions:
       2.6.8
     * 2.7.6 (set by /home/yyuu/.pyenv/version)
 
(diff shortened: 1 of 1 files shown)


ssbarnea:
Was this feature silently removed or got broken because latest pyenv behaves like this never existed:

```
$ pyenv install 3.10:latest 
python-build: definition not found: 3.10:latest

See all available versions with `pyenv install --list'.

If the version you need is missing, try upgrading pyenv:

  brew update && brew upgrade pyenv

$ pyenv --version
pyenv 2.3.9
```

native-api:
[It's broken.](https://github.com/pyenv/pyenv/issues/2569)

Issue #2137, opened 2021-10-29
Fail loudly if extension modules fail to build (e.g. _ctypes)


ThiefMaster:
It's rather annoying when `pyenv install` succeeds and you only notice later that you got a broken python installation.

Probably this would require parsing the build output, but I think it'd be worth it.

```
building '_ctypes' extension
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -std=c99 -Wextra -Wno-unused-result -Wno-unused-parameter -Wno-missing-field-initializers -Werror=implicit-function-declaration -fvisibility=hidden -I./Include/internal -I/home/adrian/.pyenv/versions/3.10.0/include -fPIC -I./Include -I. -I/
home/adrian/.pyenv/versions/3.10.0/include -I/tmp/python-build.20211029161151.349709/Python-3.10.0/Include -I/tmp/python-build.20211029161151.349709/Python-3.10.0 -c /tmp/python-build.20211029161151.349709/Python-3.10.0/Modules/_ctypes/_ctypes.c -o build/temp.linux-x86_64-3.10/tmp/python-build.20211029161151.349709/
Python-3.10.0/Modules/_ctypes/_ctypes.o -DPy_BUILD_CORE_MODULE
/tmp/python-build.20211029161151.349709/Python-3.10.0/Modules/_ctypes/_ctypes.c:107:10: fatal error: ffi.h: No such file or directory
  107 | #include <ffi.h>
      |          ^~~~~~~
compilation terminated.

Python build finished successfully!
The necessary bits to build these optional modules were not found:
_tkinter
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                  pwd                   time


Failed to build these modules:
_ctypes
```

Especially since `pyenv doctor` didn't complain about anything such a failure is somewhat unexpected...

webknjaz:
> IIRC If `_ctypes` build fails, Pip installation fails at a later stage of the build. Didn't it for you?

[...]

PR #2353, opened 2022-05-05
Add post-install checks for curses, ctypes, lzma, and tkinter

aphedges:

### Prerequisite
  - Related to #2137 but seems different enough to not close that issue

### Description

I got warnings on one machine that `bz2`, `readline`, and `sqlite3` could not be built. When installing the build dependencies according to the wiki, I realized that some warnings were missing. I've added warnings for [`curses`](https://docs.python.org/3/library/curses.html), [`ctypes`](https://docs.python.org/3/library/ctypes.html), and [`tkinter`](https://docs.python.org/3/library/tkinter.html) for all Python 3 versions and warnings for [`lzma`](https://docs.python.org/3/library/lzma.html) for Python 3.3 and later.

The wording for the warning messages might need to be adjusted, but I tried my best to match the style for similar warnings.

### Tests

I tested on Python 3.9.12 and 3.10.4 on CentOS 7 systems both with and without these packages to ensure the warnings work as expected. I also tested Python 3.0.1 and 3.3.7 to ensure I didn't break any older versions with these changes.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1788,9 +1788,12 @@ build_package_verify_py27() {
 build_package_verify_py30() {
   verify_python "${2:-3.0}"
   try_python_module "bz2" "bzip2 lib"
+  try_python_module "curses" "ncurses lib"
+  try_python_module "ctypes" "libffi lib"
   try_python_module "readline" "GNU readline lib"
   verify_python_module "ssl" "OpenSSL lib"
   try_python_module "sqlite3" "SQLite3 lib"
+  try_python_module "tkinter" "Tk toolkit"
   verify_python_module "zlib" "zlib"
 }
 
@@ -1807,6 +1810,7 @@ build_package_verify_py32() {
 # Post-install check for Python 3.3.x
 build_package_verify_py33() {
   build_package_verify_py32 "$1" "${2:-3.3}"
+  try_python_module "lzma" "lzma lib"
 }
 
 # Post-install check for Python 3.4.x


plugins/python-build/bin/python-build
```diff
@@ -1788,9 +1788,12 @@ build_package_verify_py27() {
 build_package_verify_py30() {
   verify_python "${2:-3.0}"
   try_python_module "bz2" "bzip2 lib"
+  try_python_module "curses" "ncurses lib"
+  try_python_module "ctypes" "libffi lib"
   try_python_module "readline" "GNU readline lib"
   verify_python_module "ssl" "OpenSSL lib"
   try_python_module "sqlite3" "SQLite3 lib"
+  try_python_module "tkinter" "Tk toolkit"
```
native-api, line 1796: Not sure about Tkinter. We do not require the corresponding dependency in MacOS in https://github.com/pyenv/pyenv/wiki#suggested-build-environment (but that might be because we could previously use it from XCode SDK).
Then, in environments without an X server, there's no point building a GUI library, pulling all its dependencies.

native-api (commented):
Since these libs are not strictly required for normal Python operation, maybe it'll be more productive to [instead detect when the extensions which _do_ have the corresponding dependencies installed fail to build](https://github.com/pyenv/pyenv/issues/2137)?

Most modules that are checked now are more or less strictly required. `readline` is needed for a meaningful interactive mode which is Python's big selling point. `bz2` is needed for `zipimport` IIRC and `ssl` and `ctypes` for Pip.

native-api:
OTOH, since all these extension modules are part of the standard library, _with no "ifs" and "buts"_ -- a Python without them can very well be considered not fully functional -- so it's perfectly valid to warn the user about that.

I would pass on TKinter still since servers are a huge niche for Pyenv and seeing this pointless, unsilenceable warning would be hugely annoying for their operators.

aphedges:
> OTOH, since all these extension modules are part of the standard library, _with no "ifs" and "buts"_ -- a Python without them can very well be considered not fully functional -- so it's perfectly valid to warn the user about that.

I agree with this train of thought. These are all packages in the standard library, and they can generally be assumed to be functional in any Python installation. I know that I've had problems where Python would mysteriously fail weeks after I install it because I eventually get around to opening an LZMA-compressed zip file or a similar operation. I know that the current `python-build` code considers `ssl` and `zlib` important enough that the build will fail without them present. I think both `ctypes` and `readline` are vital enough that requiring them (instead of simply warning) would make sense. I don't really know where to draw the line, though.

I agree that having Tkinter installed on servers doesn't really make sense. However, likely because it wasn't listed on the wiki, I did have problems a couple of years ago getting Tkinter to work on macOS because I didn't even realize that building the module had failed. It'd be nice if there were an easy way to only warn for Tkinter in some circumstances, but I'm not sure what those should be. Maybe have some environment variable to enable or disable the warning, or maybe only warn on macOS, which likely isn't going to be a server.

native-api:
There is a server version of MacOS. In fact, that's what MacOS CI environments use.

aphedges:
Are you sure? I remember hearing that macOS Server was discontinued. From [About macOS Server 5.7.1 and later - Apple Support](https://support.apple.com/en-us/HT208312):

> As of April 21, 2022, Apple has discontinued macOS Server.

I definitely didn't think about whatever version of macOS runs on CI, though. I'm presuming it's similar to standard macOS.

Regardless, I'm not sure what heuristics to use to figure out whether to warn about missing `tkinter` or not. Detecting the OS doesn't work because it's not needed in CI. Detecting CI doesn't work because there is no general way to detect that. I can see there is a `use_tcltk` function in `python-build`, but it doesn't seem like it has a useful return value, and it probably wouldn't help if Tcl/Tk isn't installed on the system at all.

I'm thinking that an environment variable to enable or disable the warning might be what is needed, given the inability to decide what the user wants automatically. I can implement it, but I'd like to know whether you'd want the warning to default be enabled or not.

native-api:
I don't think anyone would bother with an envvar that affects so little...

Is TKinter such a big deal anyway? I've tried building apps on it. Ended up with a two-month-salary worth of damages. It's horrible to use, has no tooling, no documentation, looks like sh1t and underneath that are serious (as in causing routine crashes) bugs that no-one has bothered to fix for _decades._ I've yet to see a _single_ 3rd-party app built on it, everyone uses Qt or wxWidgets. If that's not evidence that it's is a dead-end library that no-one cares about, I dunno what is.

native-api:
Anyway, @pyenv/pyenv-core-maintainers , what's your opinion on this? I'm probably biased here...

aphedges:
I personally rarely use Tkinter and have never written anything with it, but I do need to use it maybe once or twice a year to run some existing program or another. I don't have any strong opinions about it as a framework besides the fact that it's in the standard library so some packages assume it's present.

yyuu:
To be quite honest, I cannot provide any clear guidance and/or draw line within standard libraries. My assessment below is all about my personal guess and I might be wrong. tl;dr - let’s remove verification for TkInter. How do you guys think…? 🤔 

With regard to other standard libraries we are verifying right now, SSL is vital to maintain things functional since it’s 2022 now and most of websites are on HTTPS; we need SSL before everything. Compression libraries are other parts of foundation to make external connections (to get further libraries installed…) so we wanted to ensure it to be there.

One question I have is about “sqlite3”; it has been included as a standard library, however, I have question on the popularity of the library amongst applications. Nowadays, I think it’d be more intuitive if Python treats it as a 3rd party library can be installed from PyPI, IMHO. So far we have been verifying it because it has been a part of standard library, however, it might be excessive if the library is not popular enough.

Back to the original question on drawing the line, I am fine with LZMA since it’s a part of standard library and is a foundational compression functionality. (although usage of LZMA would be lesser popular compared to GZip…) CFFI is bit tricky but I recognize it as foundational part of standard library. Although TkInter is more like to “sqlite3” case, moreover it would have even lesser popularity compare to it. I actually don’t have any statistics on popularity hence this is all about my guess, but adding verification to TkInter would be more troublesome than its benefit.

In case there are some others want to perform extra verifications for TkInter, I think it’d be better idea to start from pyenv’s “after_install” hook to perform the verification instead of changing the verification behaviour for all Pythons. I don’t think I have ever used TkInter for the last 15 years. My case might NOT be as similar as others though.

aphedges:
This discussion made me realize that having concrete statistics about standard library usage would be helpful. I spent an hour analyzing the top 500 packages on PyPI according to the list at <https://hugovk.github.io/top-pypi-packages/>. The least popular of these was downloaded over 4 million times over a recent 30 day period, so these are clearly all used often. Using a variety of heuristics, I have some vague numbers about the usage of various packages:

- `ctypes` is used by 82 of the packages.
- `tkinter` is used by 9 of the packages, including `ipython`, `matplotlib`, and `nltk`. I can confirm that I've used a GUI for the latter two multiple times before, but that was mostly when I starting to learn Python and didn't know much about using a CLI.
- `sqlite3` is used by 21 of the packages, including `Django`, `SQLAlchemy`, `ansible`, `mypy`, and `pandas`.

All the rest of the extension modules are used by at least one package, with `curses` used by 11 and `zlib used by 57. I can provide numbers for all of them, along with my methodology, if you want. There are probably some false positives, but  a manual inspection shows at least some of these are legitimate.

I'd like to point out that I'm not saying we need a build failure for any of these checks. The code I've added just prints a warning if importing the module fails. It doesn't seem to have any performance implications, so I don't see any reason to not do it except to avoid showing warnings that the user might not care about. I just feel that it would be more clear to users, especially beginners, if they know upfront that something is missing from their compiled standard library that they would expect to have if they installed a binary from Python.org or Anaconda.

native-api:
> tkinter is used by 9 of the packages, including ipython, matplotlib, and nltk

I can attest that IPython has no GUI functionality itself, it rather has a magic command to run TKinter's event loop in background (and for a number of other GUI libraries as well) so that a user is able to run a GUI and use the console without freezing either.

Matplotlib does have GUI functionality and allows one to select from a number of GUI toolkits as well as save information in a number of image formats instead of displaying it.

Looking through NTLK docs, TKinter does appear to be the only backend it can use but it's only needed if you use the `ntlk.draw` visualization submodule. `ntlk.download` uses TKinter only if it's available, and [this actually causes problems in noninteractive environments](https://support.enthought.com/hc/en-us/articles/204470000-NLTK-Natural-Language-Toolkit-Download-function-hangs).

native-api:
> However, likely because it wasn't listed on the wiki

Then a solution can be to just list it in the suggested environment :wink: That's fine by me.

native-api:
> Then a solution can be to just list it in the suggested environment wink

Done.

aphedges:
Sorry for the late response. I haven't had the time to follow up on this recently, but I did look into `$DISPLAY`. I was able to confirm it was set on my macOS and Windows installations but not my terminal-only Linux installation. I just didn't get around to implementing your suggested change. Thank you very much for implementing it yourself!

I agree that adding `tcl-tk` to the Homebrew installation instructions was a good idea. Even so, this check will still be helpful because <https://github.com/python/cpython/commit/b36d222110d0d6d84dc8e973ca87d976c2423f5d> changed the Tkinter installation logic to require `export PKG_CONFIG_PATH="$(brew --prefix)/opt/tcl-tk/lib/pkgconfig"` with Homebrew on macOS. I was able to verify this on my macOS installation with `$(pyenv prefix 3.11.0a7)/bin/python -c 'import tkinter'`. In addition, I was able to confirm the requirement for `$DISPLAY` on my macOS and Linux systems:

```console
# On macOS
$ echo $DISPLAY
/private/tmp/com.apple.launchd.v2IIgZLwTR/org.xquartz:0
$ python -c 'import tkinter'
$ python -m tkinter
# On Linux over SSH
$ echo $DISPLAY

$ python -c 'import tkinter'
$ python -m tkinter
  File "/nas/home/ahedges/.pyenv/versions/3.7.13/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/nas/home/ahedges/.pyenv/versions/3.7.13/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/nas/home/ahedges/.pyenv/versions/3.7.13/lib/python3.7/tkinter/__main__.py", line 7, in <module>
    main()
  File "/nas/home/ahedges/.pyenv/versions/3.7.13/lib/python3.7/tkinter/__init__.py", line 3988, in _test
    root = Tk()
  File "/nas/home/ahedges/.pyenv/versions/3.7.13/lib/python3.7/tkinter/__init__.py", line 2023, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable
```

PR #2404, opened 2022-07-01
describe how to use versions alongside system

niccokunzmann:
### Prerequisite

### Description

The order of the versions for `pyenv global` have a meaning.  #2406 shows this.
This documents what to do to keep the system default python
version as the version to be used for all commands for the major version but at the same time expose the other installed Python versions with their major and minor versions to PATH.
I.e. python3 -> system
python3.10 -> system
python3.9 -> pyenv
python3.8 -> pyenv
...

To use with tox.

I realize that this is something to be discussed in the forum. If that needs to happen first, please close the issue.
Although, this would have saved my some minutes of searching and trying out.

### Tests

none, documentation

### DIFF
--- a/libexec/pyenv-global
+++ b/libexec/pyenv-global
@@ -16,7 +16,12 @@
 # Example: To enable the python2.7 and python3.7 shims to find their
 #          respective executables you could set both versions with:
 #
-# 'pyenv global 3.7.0 2.7.15'
+#     $ pyenv global 3.7.0 2.7.15
+#
+# Example: To enable all shims and the system installed version,
+#          write "system" first.
+#
+#     $ pyenv global system `pyenv versions --bare`
 #
 
 
--- a/libexec/pyenv-local
+++ b/libexec/pyenv-local
@@ -23,8 +23,13 @@
 # Example: To enable the python2.7 and python3.7 shims to find their
 #          respective executables you could set both versions with:
 #
-# 'pyenv local 3.7.0 2.7.15'
-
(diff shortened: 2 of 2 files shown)


native-api (commented):
Please report the problem that this is supposed to solve, preferrably as an issue with the necessary diagnostic info.

Pip3 segfaults usually have nothing to do with version order so the proposed change seems irrelevant so far.

niccokunzmann:
@native-api, I described more clearly what this documentation solves for me. Also, I opened the issue #2406 which this does not solve.

native-api (changes requested):
In the case you've reported, `pip3` corresponds to 3.6.15 -- because all Python 3 installations provide the `pip3` executable and 3.6.15 is the first in the list.

Indeed, if you wish `pip3` to correspond to the system version, you need to specify `system` before other py3 installations -- but that won't fix the segmentation fault with 3.6.15's Pip.

---

So the proposed text is incorrect as written -- it's not a requirement write `system` first as it says.

---

It will however be okay if it's rewritten to instead explain this peculiarity with `pip3`. (That is, if multiple selected installations provide an executable, the first one in the list that does wins.)

niccokunzmann:
That sounds like a good proposal to me. I will change it when I find the time.

native-api:
It's probably also better to place it into https://github.com/pyenv/pyenv#understanding-python-version-selection because this mechanic is not specific to `pyenv global`.

Issue #2097, opened 2021-10-06
Pyenv versions listing should use version sort


Rotonen:
With the introduction of 3.10, the sorting of versions is no longer intuitively correct.

```
$ pyenv versions
* system
  3.10.0
  3.9.7
  ```

This seems to happen as a side effect of glob expansion sorting on a byte for byte basis on most filesystems.

https://github.com/pyenv/pyenv/blob/fab0082bd5cdda07f0bfdd69a9c676bc2d2906b3/libexec/pyenv-versions#L135

GNU `sort` would have a `-V` `--version-sort` [[1](https://man7.org/linux/man-pages/man1/sort.1.html)] flag, but that's not portable as at least `sort` on macOS does not have that [[2](https://ss64.com/osx/sort.html)].

Columnar sort with a separator [[2](https://stackoverflow.com/a/39454220/1214697)]could be a way to approach the problem. This could be added as a pipe to the end of that for loop.

This is the path the `--definitions` command of `python-build` has taken.

https://github.com/pyenv/pyenv/blob/fab0082bd5cdda07f0bfdd69a9c676bc2d2906b3/plugins/python-build/bin/python-build#L1825-L1830

https://github.com/pyenv/pyenv/blob/fab0082bd5cdda07f0bfdd69a9c676bc2d2906b3/plugins/python-build/bin/python-build#L1832-L1835

[1] https://man7.org/linux/man-pages/man1/sort.1.html
[2] https://ss64.com/osx/sort.html
[3] https://stackoverflow.com/a/39454220/1214697

native-api:
> This seems to happen as a side effect of glob expansion sorting on a byte for byte basis on most filesystems.

This has nothing to do with filesystem order. [Bash guarantees the sort to be alphabetical.](https://serverfault.com/questions/122737/in-bash-are-wildcard-expansions-guaranteed-to-be-in-order)

native-api:
> This seems to happen as a side effect of glob expansion sorting on a byte for byte basis on most filesystems.

This has nothing to do with filesystem order. [Bash guarantees the sort to be alphabetical.](https://serverfault.com/questions/122737/in-bash-are-wildcard-expansions-guaranteed-to-be-in-order)

If you want to reorder it, you'll have to support arbitrary filenames. (E.g. read entries to an array, sort them like in https://stackoverflow.com/questions/7442417/how-to-sort-an-array-in-bash but use `\0` as a separator and `sort -z`)

[...]

PR #2405, opened 2022-07-01
Use version sort in `pyenv versions`

fofoni:

Considered; implementation is too simple for a plugin.

Is this recommendation still valid? Looks like rbenv's had version sort for months: rbenv/rbenv@28cd6f12

  - Closes https://github.com/pyenv/pyenv/issues/2097

### Description
Check if the system `sort` command supports version sort and, if so, use it.

I used `sort --version-sort` as discussed in #2097. `rbenv`'s solution does not depend on `--version-sort`, and would require a little bit more refactoring, but I'd be happy to port it to pyenv if you think it's worth the effort.

### Tests
Should I add any?

### DIFF
--- a/test/versions.bats
+++ b/test/versions.bats
@@ -175,12 +175,12 @@ OUT
   create_version "1.53.0"
   create_version "1.218.0"
 
-  correct_versions_file="$(mktemp)"
+  correct_versions_file="$(mktemp "${PYENV_TEST_DIR}/tmp.XXXXXXXXXX")"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.9.0" >>"$correct_versions_file"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.53.0" >>"$correct_versions_file"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.218.0" >>"$correct_versions_file"
 
-  wrong_versions_file="$(mktemp)"
+  wrong_versions_file="$(mktemp "${PYENV_TEST_DIR}/tmp.XXXXXXXXXX")"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.218.0" >>"$wrong_versions_file"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.53.0" >>"$wrong_versions_file"
   printf '%s/versions/%s\n' "${PYENV_ROOT}" "1.9.0" >>"$wrong_versions_file"
--- a/test/versions.bats
+++ b/test/versions.bats
@@ -17,6 +17,14 @@ stub_system_python() {
   touch "$stub" && chmod +x "$stub"
 }
 
+create_executable() {
+  local name="$1"
+  local bin="${PYENV_TEST_DIR}/bin"
+  mkdir -p "$bin"
+  sed -Ee '1s/^ +//' > "${bin}/$name"
+  chmod +x "${bin}/$name"
+}
+
 @test "no versions installed" {
   stub_system_python
   assert [ ! -d "${PYENV_ROOT}/versions" ]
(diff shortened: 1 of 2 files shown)


fofoni:
Oops I see that I'm relying on the existence of at least some `sort` command; I'll fix it.

native-api:
`sort` is a standard UNIX command and is used elsewhere, too. You can rely on its presence.

fofoni:
> `sort` is a standard UNIX command and is used elsewhere, too. You can rely on its presence.

apparently it's missing inside the CI environment, because it was breaking some bats tests

fofoni:
BTW how do I run the test suite locally? Instead of pushing only to find I've broken more tests.

native-api:
> BTW how do I run the test suite locally? Instead of pushing only to find I've broken more tests.

`make test`

fofoni:
Looks like it's good now!

native-api (commented):
As new functionality, this probably warrants a test.

fofoni:
> As new functionality, this probably warrants a test.

The problem is the CI environment [doesn't have `sort`](https://github.com/pyenv/pyenv/runs/7155941041#step:6:187). I could implement a test like "if system has sort, assert sorted by version; else, assert sorted lexicographically" -- would it be ok?

native-api:
> apparently it's missing inside the CI environment

I don't think so. Github's CI enviorments are supposed to emulate Ubuntu and h

> The problem is the CI environment [doesn't have `sort`](https://github.com/pyenv/pyenv/runs/7155941041#step:6:187).

It does have `sort` as it's a part of `coreutils`. The reason it's not found there is the [`path_without`](https://github.com/pyenv/pyenv/blob/304515f2cdd11db151b7a5733d11934d3990a67e/test/test_helper.bash#L117-L138) test function. It doesn't include `sort` into the list of executables it ensures to be present on PATH -- `sort` needs to be added to that list.

native-api:
> I could implement a test like "if system has sort, assert sorted by version; else, assert sorted lexicographically"

The tests need to run in systems with both "modern" and "old" `sort` -- so you need to mock it for the test.

Since the new logic only fires with "modern" sort, it's sufficient to mock up a "modern" `sort` for the new test: the logic for "old" sort is already tested in existing tests which have test data where lexicographical sorting makes no difference.

There is an existing function that mocks an executable, use it as an example. The logic for the mockup can be to just unconditionally output the correct result for the specific set of versions that you make in the test.

fofoni:
Ok, the test passes, but I'm not sure if that's what you meant.

If you drop the `--version-sort` in https://github.com/fofoni/pyenv/blob/c70edbec5386d617f5ace823c97d90a85654c51a/libexec/pyenv-versions#L134 then the test will fail. Is that it?

fofoni:
Also, I didn't add `sort` to the whitelist inside `path_without`, since it's not required anymore to make the tests pass.

test/versions.bats
```diff
@@ -161,3 +169,34 @@ OUT
   run pyenv-versions --bare
   assert_success ".venv"
 }
+
+@test "uses version sort when available" {
+  create_version "1.9.0"
+  create_version "1.53.0"
+  create_version "1.218.0"
+
+  correct_versions_file="$(mktemp)"
```
native-api, line 178: Any temporary files need to be created under `$PYENV_TEST_DIR` so that they are deleted after the test. Does this ensure that?

native-api (commented):
The test is fine.

Thinking about it though, other tests would fire the code branch that correspond to the system's `sort`. So we do also need a test for an old `sort`, likewise mocking it -- or in a modern system, this case would not be tested :neutral_face:

test/versions.bats
fofoni: It doesn't -- I thought about that after I pushed the commit, I'll fix it.

fofoni:
Well, the mktemp problem is fixed.

Now, just to be sure: with the implementation we have right now, if the system doesn't have a modern `sort` binary, then no `sort` binary is used at all. In this case, what happens is the old behavior of getting the order (guaranteed by bash as you mentioned [here](https://github.com/pyenv/pyenv/issues/2097#issuecomment-937620720)) from the glob `$versions_dir/*`.

As you also already said,

> Since the new logic only fires with "modern" sort, it's sufficient to mock up a "modern" sort for the new test: the logic for "old" sort is already tested in existing tests which have test data where version sorting makes no difference.

this old behavior is already checked in tests like "[multiple versions](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L64)", "[indicates current version](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L79)", "[globally selected version](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L103)", and others in the `versions.bats` file.

I understand that maybe explicitly testing the order with another test dedicated for this might be desired; I'd just like to make sure that's what you're asking for.

native-api:
> this old behavior is already checked in tests like "[multiple versions](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L64)", "[indicates current version](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L79)", "[globally selected version](https://github.com/fofoni/pyenv/blob/master/test/versions.bats#L103)", and others in the `versions.bats` file.

The problem is, in those tests, we don't mock anything. So with a new `sort`, those tests would fire the new execution path rather than the old one.

fofoni:
Oh ok, that makes sense

native-api:
Good job. Thank you!

PR #2412, opened 2022-07-14
Fixed relative path traversal due to using version string in path

comrumino:
### Description
- It looks like rbenv addressed the [path traversal vulnerability some time ago.](https://github.com/rbenv/rbenv/commit/370c26a6c9ee0511972ea04904fcc89014a22987)
- There are some additional change available in rbenv that simplify the read logic. Even so, my changes are functionally the same.

### DIFF
--- a/libexec/pyenv-version-file-read
+++ b/libexec/pyenv-version-file-read
@@ -11,9 +11,16 @@ if [ -s "$VERSION_FILE" ]; then
   IFS="${IFS}"$'\r'
   sep=
   while read -n 1024 -r version _ || [[ $version ]]; do
-      [[ -z $version || $version == \#* ]] && continue
-      printf "%s%s" "$sep" "$version"
-      sep=:
+    if [[ -z $version || $version == \#* ]]; then
+      # Skip empty lines and comments
+      continue
+    elif [ "$version" = ".." ] || [[ $version == */* ]]; then
+      # The version string is used to construct a path and we skip dubious values.
+      # This prevents issues such as path traversal (CVE-2022-35861).
+      continue
+    fi
+    printf "%s%s" "$sep" "$version"
+    sep=:
   done <"$VERSION_FILE"
   [[ $sep ]] && { echo; exit; }
 fi
--- a/test/version-file-read.bats
+++ b/test/version-file-read.bats
@@ -82,3 +82,15 @@ IN
   run pyenv-version-file-read my-version
   assert_success "3.9.3:3.8.9:2.7.16"
 }
+
+@test "skips relative path traversal" {
+  cat > my-version <<IN
+3.9.3
+3.8.9
+  ..
+./*
+2.7.16
+IN
+  run pyenv-version-file-read my-version
+  assert_success "3.9.3:3.8.9:2.7.16"
+}


libexec/pyenv-version-file-read
```diff
@@ -11,9 +11,15 @@ if [ -s "$VERSION_FILE" ]; then
   IFS="${IFS}"$'\r'
   sep=
   while read -n 1024 -r version _ || [[ $version ]]; do
-      [[ -z $version || $version == \#* ]] && continue
-      printf "%s%s" "$sep" "$version"
-      sep=:
+    [[ -z $version || $version == \#* ]] && continue
+    # The version string is used to construct a path. Non-zero exit when
+    # invalid characters are found. This prevents issues such as CVE-2022-35861.
+    if [[ ! "$version" =~ "^[a-z0-9][-.0-9a-z]+$" ]]; then
```
native-api, line 17: The Linux kernel allows any characters in a directory entry except `\0` and `/` (which is a path separator). So this looks overzealous to me. The Rbenv commit does it right.

native-api (commented):
Do we really need to disallow path separators altogether?

AFAICS, the vulnerability rather lies in the fact that the resulting path can be outside of `$PYENV_ROOT/versions`. So it's technically sufficient to normalize the path and check if it lies within `$PYENV_ROOT/versions`.

That use case it not officially supported though -- so I wouldn't mind just disallowing path traversal constructs altogether like Rbenv did.

native-api:
In the e-maiol, you mentioned having a CVE number assigned. Did you get it? If yes, it'll probably be worth mentioning.

comrumino:
> In the e-mail, you mentioned having a CVE number assigned. Did you get it? If yes, it'll probably be worth mentioning.

I have CVE-2022-35861 reserved. 

> disallowing path traversal constructs altogether like Rbenv did.

Either approach makes sense to me. My original approach was more of an "allow" approach where rbenv took a "deny" approach.

comrumino:
I am done amending my commit. Please review the newest commit.

The current CVE form fields are below. Let me know if you would like any revisions.

### Suggested description
pyenv 1.2.24 through 2.3.2
Allows local users to gain privileges via a .python-version file in the current working directory.
An attacker can craft a Python version string in .python-version
to execute shims under their control. (Shims are executables that
pass a command along to a specific version of pyenv. The version
string is used to construct the path to the command, and there is no
validation of whether the version specified is a valid version.)

### VulnerabilityType Other
Uncontrolled Search Path Element

### Vendor of Product
pyenv

### Affected Product Code Base
pyenv - 1.2.24-2.3.2

### Affected Component
.python-version,shims/*

### Attack Type
Context-dependent

### Impact Code execution
true

### Attack Vectors
Crafted .python-version file in the current directory allows attacker to specify shims under their control

### Reference
https://github.com/pyenv/pyenv
https://github.com/pyenv/pyenv/pull/2412

### Discoverer
James Stronz

libexec/pyenv-version-file-read
```diff
@@ -6,16 +6,18 @@ set -e
...
+  # Read the first word from the specified version file. Avoid reading it whole.
   IFS="${IFS}"$'\r'
-  sep=
-  while read -n 1024 -r version _ || [[ $version ]]; do
-      [[ -z $version || $version == \#* ]] && continue
-      printf "%s%s" "$sep" "$version"
-      sep=:
-  done <"$VERSION_FILE"
-  [[ $sep ]] && { echo; exit; }
+  read -n 1024 -d "" -r version _ <"$VERSION_FILE" || :
```
native-api, line 11: The removal of the loop will disallow parsing multiple versions in the file. Rbenv doesn't support that apparently.

Tests are also missing.

native-api:
I tried to merge Rbenv codebase like @yyuu did before -- since there may be other useful changes besides this -- but was unsuccessful so far (some files didn't merge because of different names, perhaps they weren't properly renamed with Git when forking or merging before).

I'll continue trying, so whichever effort completes first will get into the codebase.

libexec/pyenv-version-file-read
comrumino: I amended my work again, this time I took the time to add some new unit tests as well :)

PR #2421, opened 2022-07-26
Allow pypy2 pypy3 patching

brogon:
### Description
Update to modifications after PR https://github.com/pyenv/pyenv/pull/2419 which didn't work due to:
 -  "extglob" mode in BASH doesn't support RegEx-like character classes like `[:digits:]`, you need to
    use `[0-9]` (unless there is another option for *that*...)

 -  BASH `shopt` is *really* strange - it's used by both the parser *and* the runtime: you need to "wrap" 
    `apply_python_patch()` with `shopt -s extglob / shopt -u extglob`, but then *also* call `shopt -s extglob`
    and `shopt -u extglob` *again* within the function.    
    Otherwise it parses the glob pattern, but applies it "literally" (non-extglob-wise) - which won't match 
    anything reasonable. :grin:

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1639,11 +1639,17 @@ build_package_auto_tcltk() {
 
 # extglob must be set at parse time, not at runtime
 # https://stackoverflow.com/questions/49283740/bash-script-throws-syntax-errors-when-the-extglob-option-is-set-inside-a-subsh
+#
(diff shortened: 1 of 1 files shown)


native-api:
Dammit!
I knew that warranted a test...

native-api:
My apologies for pushing untested code in the previous PR :frowning_face:

PR #2423, opened 2022-07-27
Add Fig as an autocomplete method

brendanfalk:

### Prerequisite

### Description

[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including pyenv, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ This project was forked from [rbenv](https://github.com/rbenv/rbenv) and
 * [Upgrading](#upgrading)
 * [Uninstalling pyenv](#uninstalling-pyenv)
 * [Advanced Configuration](#advanced-configuration)
+  * [Autocomplete](#autocomplete)
   * [Using Pyenv without shims](#using-pyenv-without-shims)
   * [Environment variables](#environment-variables)
 * **[Development](#development)**
@@ -562,6 +563,15 @@ that may shadow Pyenv's shims.
   in these distributions because the system's Pip places executables for
   modules installed by a non-root user into those per-user `bin` directories.
 
+### Autocomplete
+
+<a href="https://fig.io/" target="_blank"><img src="https://fig.io/badges/Logo.svg" width="15" height="15"/></a> [Fig](https://fig.io/) provides IDE-style autocompletions for pyenv. It works in bash, zsh, and fish.
+
+To install, run:
+
+```shell
+brew install fig
+```
 
 ### Using Pyenv without shims
 


native-api:
@pyenv/pyenv-core-maintainers Any opinions?

AFAICS, the tool is completely independent from Pyenv and its bundled completions, it maintains its own database (I dunno how flexible and that probably means that it's not environment-aware). At the very least, that means that it will be somewhat disconnected from the specific Pyenv version installed and its plugins.
It also only supports MacOS.

So atm I'm not sure if we can even recommended it, not to say endorse...

brendanfalk:
Hey @native-api! So yes, you're totally correct, Fig maintains completions for public CLI tools in its [withfig/autocomplete](https://github.com/withfig/autocomplete) repo, separate to the CLI itself. Here is the completion spec we have for pyenv currently: https://github.com/withfig/autocomplete/blob/master/src/pyenv.ts

1. We actually are fully aware of your environment and can actually do very powerful suggestions. e.g.
pyenv_demo

3. We can change the completions we give users based on the version of pyenv installed. See https://fig.io/docs/handbook/versioning-completion-specs. Admittedly, we haven't set this up for pyenv just yet, but we'd absolutely love to work with you on this (it's quite easy to do, and we'd be happy to help)
4. We are macOS only right now but are literally < 1 month away from launching support for Window and Linux:
* Linux: https://github.com/withfig/fig/issues/34
* Windows: https://github.com/withfig/fig/issues/35

Finally, we have 100k+ people who use Fig, and pyenv is one of the top 100 CLI tools Fig users use for autocomplete (whereas python, pip, etc are all much higher). We'd generally love to work with you to support pyenv as much as we can and make the experience absolutely great!

Happy to answer any more questions.

native-api:
Okay. After consulting team members, here's the deal.

A separate, prominent advertisement section recommending Fig as the default way of action is too much -- especially since we cannot vouch for Fig's completions' accuracy and up-to-dateness for whatever Pyenv version the user has installed.
But we can briefly mention Fig under an appropriate installation section -- e.g. at the end of https://github.com/pyenv/pyenv#set-up-your-shell-environment-for-pyenv , -- with smth like this:

> In MacOS, you might also want to install [Fig](https://fig.io/) which provides alternative shell completions with an IDE-like popup interface.

Would this okay with you? Is the suggested wording accurate -- e.g. are any other steps needed to ensure that Fig won't conflict with the suggested shell setup?

brendanfalk:
@native-api My apologies. I clearly just completely missed your November comment.

I have created a new PR (https://github.com/pyenv/pyenv/pull/2574) with your suggestion above. Let me know how this looks and thanks again for your help

Issue #2336, opened 2022-04-23
'pyenv uninstall' multiple versions


adamchainz:
After upgrading python versions, I want to uninstall the old ones. Instinctively I try uninstall a bunch with:

```
$ pyenv uninstall 3.11.0a6 3.10.3 3.7.11 3.9.10
```

but this doesn't work:

```
Usage: pyenv uninstall [-f|--force] <version>

   -f  Attempt to remove the specified version without prompting
       for confirmation. If the version does not exist, do not
       display an error message.

See `pyenv versions` for a complete list of installed versions.
```

I'd like `pyenv uninstall` to accept multiple versions, so I don't need to run multiple commands.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): macOS
- [x] OS architecture (e.g. amd64): x86
- [x] pyenv version: 2.2.5
- [x] Python version: 3.10.4
- [x] C Compiler information (e.g. gcc 7.3): clang
- [n/a] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`. E.g. if you have a problem with installing Python, run `env PYENV_DEBUG=1 pyenv install -v <version> 2>&1 | tee trace.log` (note the `-v` option to `pyenv install`).

hardikpnsp:
This looked simple enough for a first-time contributor to `pyenv`, I gave it a try. Hoping that this is helpful.

adamchainz:
[...]

PR #2432, opened 2022-08-06
Add support for multiple versions in `pyenv uninstall`

hardikpnsp:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2336

### Description
- [x] adds support for uninstalling multiple python version at once using `pyenv uninstall`
- [x] updates docs (and man pages entries?)

### Tests
- [x] replaces unit test `too many arguments for pyenv-uninstall` with `more than one argument for pyenv-uninstall`
- [x] adds test for `pyenv-uninstall hooks` for the multiple version scenario

### DIFF
--- a/plugins/python-build/bin/pyenv-uninstall
+++ b/plugins/python-build/bin/pyenv-uninstall
@@ -2,7 +2,7 @@
 #
 # Summary: Uninstall Python versions
 #
-# Usage: pyenv uninstall [-f|--force] <version1> <version2> <..>
+# Usage: pyenv uninstall [-f|--force] <version> ...
 #
 #    -f  Attempt to remove the specified version without prompting
 #        for confirmation. If the version does not exist, do not
--- a/plugins/python-build/bin/pyenv-uninstall
+++ b/plugins/python-build/bin/pyenv-uninstall
@@ -35,6 +35,16 @@ fi
 
 [ "$#" -gt 0 ] || usage 1 >&2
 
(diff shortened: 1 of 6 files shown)


plugins/python-build/bin/pyenv-uninstall
```diff
@@ -59,29 +52,43 @@ IFS=$'\n' scripts=(`pyenv-hooks uninstall`)
 IFS="$OLDIFS"
 for script in "${scripts[@]}"; do source "$script"; done
 
+uninstall-python() {
+  local DEFINITION="$1"
+  case "$DEFINITION" in
+  "" | -* )
+    usage 1 >&2
+    ;;
+  esac
```
native-api, line 61: Shouldn't we validate the command line before starting any work?

hardikpnsp: I agree, we should. will fix this in a subsequent commit. Also, do you think it would be a good idea to do the directory check `[ ! -d $PREFIX]` for all versions beforehand too? (when not running in --force mode). This would make the command behavior more atomic. but on the other hand, I don't see much issue with the uninstall command not being atomic in interactive mode.

native-api:
Thank you for your efforts! Supporting multiple versions in commands that currently only support one is a feature that has been requested for some time! (Though it's not the most requested one so we weren't focusing on it)
This is a start!

Issue #2105, opened 2021-10-11
Add Cinder compiler (Python 3.8.5+cinder) support


alexdoesh:
I'd like to add Facebook's [Cinder](https://github.com/facebookincubator/cinder/) compiler support to pyenv, what's the best way to make it happen soon? They have no proper distribution yet, but it's possible to get the source and compile it (with or without docker image, I have playbooks for both). Can somebody point me, what would be the best approach here, I'll wrap it up and submit PR.

native-api:
We now have a contribution guide which gives more details about authoring installation scripts.

PR #2433, opened 2022-08-07
Add Cinder 3.8

filips123:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2105

### Description
- [x] Adds script for installing Facebook's [Cinder](https://github.com/facebookincubator/cinder/)
- [x] Based on `python3.8-dev` script, adjusted for Cinder's repository
- [x] Cinder does not have any versioned releases currently, so this just uses the repository
- [x] Cinder 3.10 is currently not finished and I couldn't manage to successfully compile it
- [x] According to the README, macOS is *probably* not supported; should I mark this somehow?

### Tests

No tests are added, but I managed to compile this on Debian 11 and run basic Numpy and Flask scripts.

### DIFF
--- a/plugins/python-build/share/python-build/cinder-3.8-dev
+++ b/plugins/python-build/share/python-build/cinder-3.8-dev
@@ -1,0 +1,5 @@
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1k.tar.gz#892a0875b9872acd04a9fde79b1f943075d5ea162415de3047c327df33fbaee5" mac_openssl --if has_broken_mac_openssl
+install_package "readline-8.0" "https://ftpmirror.gnu.org/readline/readline-8.0.tar.gz#e339f51971478d369f8a053a330a190781acb9864cf4c541060f12078948e461" mac_readline --if has_broken_mac_readline
+install_git "Cinder-3.8-dev" "https://github.com/facebookincubator/cinder" "cinder/3.8" standard verify_py38 copy_python_gdb ensurepip
--- a/plugins/python-build/share/python-build/cinder-3.8-dev
+++ b/plugins/python-build/share/python-build/cinder-3.8-dev
@@ -1,14 +1,14 @@
 require_distro Fedora 32 &>/dev/null || \
-  echo >&2 <<!
+  cat >&2 <<!
 Warning! The Cinder compiler only officially supports
 Facebook's Docker images which are Fedora 32 - based.
 It is likely to fail to build on a system
 with a different GCC and/or Glibc version.
 !
 
-[[ $(gcc --dumpversion) == 10 ]] || \
+[[ $(gcc -dumpversion) == 10 ]] || \
 { command -v "gcc-10" && export CC=gcc-10; } || \
-echo >&2 <<!
+cat >&2 <<!
 Warning! GCC 10 is not found on PATH.
 The build will likely fail.
 !
--- a/plugins/python-build/share/python-build/cinder-3.8-dev
+++ b/plugins/python-build/share/python-build/cinder-3.8-dev
@@ -2,15 +2,15 @@ require_distro Fedora 32 &>/dev/null || \
   cat >&2 <<!
 Warning! The Cinder compiler only officially supports
 Facebook's Docker images which are Fedora 32 - based.
-It is likely to fail to build on a system
+It may fail to build on a system
 with a different GCC and/or Glibc version.
 !
 
-[[ $(gcc -dumpversion) == 10 ]] || \
-{ command -v "gcc-10" && export CC=gcc-10; } || \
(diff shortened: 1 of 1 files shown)


native-api:
Build is failing in our stock environments. Please take a look at that.

filips123:
On Ubuntu 20.04 from GH Actions, it seems GCC is too old to support C++20. On 18.04, there are some other failures that are still probably related to the too old compiler. macOS build failures are kinda expected because Cinder doesn't support it.

It also doesn't work in Ubuntu 22.04 but works in Debian 11, so apparently, there are some differences in compilers on Ubuntu and Debian that cause build failures. I'm not really sure why this happens... Maybe Cinder is really too experimental and unstable to be used outside the provided Docker images.

native-api:
Compilers provide more-or-less standard C/C++ environments... if Cinder doesn't support either of the 2 major compilers -- what _does_ it support, anyway?

It's fine if it only supports some bleeding-edge versions and we test for those in the installation script and give some meaningful error messages if those are missing... but so far, it's unclear what we should even test for!

native-api:
> It's fine if it only supports some bleeding-edge versions and we test for those in the installation script and give some meaningful error messages if those are missing...

That said, usually, the software's `configure` scripts does tests like this and bails out if they fail, before any confusing compilation errors start piling up... It's strange that Cinder's doesn't -- _if_ they want third parties to use it, that is.

filips123:
> It's fine if it only supports some bleeding-edge versions and we test for those in the installation script and give some meaningful error messages if those are missing... but so far, it's unclear what we should even test for!

Yeah, I agree.

> if they want third parties to use it, that is.

Well, the thing is that they don't really want to (compile Cinder outside their environement):

> Cinder should build just like CPython; configure and make -j. However as most development and usage of Cinder occurs in the highly specific context of Meta we do not exercise it much in other environments. As such, the most reliable way to build and run Cinder is to re-use the Docker-based setup from our GitHub CI workflow.

I thought there would still be fewer problems when I created this PR. I currently don't have much time to deal with them, so I guess it can be closed until someone tries to fix them or Cinder supports more platforms.

native-api:
We do have a stock function, `require_distro`, to test for Linux distro and version and warn if it doesn't check.

If you say that it runs in some vanilla distros, that should be sufficient.

(Its message wording would need to be altered to also fit source distributions though.)

filips123:
I did some [tests](https://github.com/filips123/pyenv-cinder-testing/actions/runs/2869593711) on various distros to check what Cinder supports.

It appears Cinder only supports GCC 10:

* Builds on Ubuntu 16.04-20.04 (GCC 5.4.0-9.4.0) all failed because of too old GCC versions that don't support C++20 or other required features. The same with Debian 9-10 (GCC 6.3.0-8.3.0) and CentOS 7 (GCC 4.8.5).

* Builds on Debian 11 (GCC 10.2.1) and Fedora 32-33 (GCC 10.3.1) work fine. Fedora 32 is also what is used to build Cinder officially in Docker container.

* What's more interesting is that newer GCC versions also do not seem to work. Builds Ubuntu 22.04 (GCC 11.2.0), Fedora 34-37 (GCC 11.3.1-12.1.1), Arch (GCC 12.1.1), and Alpine (GCC 11.2.1 with musl) all fail (with the same three errors, one on GCC 11, another on GCC 12, and third on Alpine).

* (I also tried tests on OpenSUSE, but they failed because of some unrelated error.)

native-api:
Could you try building with GCC 10 (`CC=gcc-10 pyenv install <...>`) in environments where it's available/can be installed?

If it (at least somewhat) works, we can set the above envvar if GCC 10 is available (and that envvar is not already set) and warn otherwise.
Since Fedora 32 is the only officially supported ewnvironment, it's still worth checking the distro against it regardless.

native-api:
I also just realized that you can just silence `require_distro`'s output and check its exit code if you need a custom warning message :slightly_smiling_face:

filips123:
Adding `CC=gcc-10` on Ubuntu 22.04 still [doesn't work](https://github.com/filips123/pyenv-cinder-testing/runs/7882340687?check_suite_focus=true). So maybe it's related to glibc version?

In case this can be fixed, I think the install script could do something like this:

* If distro is not Fedora 32, emit warning about distro.
* If normal `gcc` is already 10 (for example, on Debian 11 and Fedora 32/33), just build it.
* If it's not, but `gcc-10` exists, set it as `CC`.
* If it still doesn't exist, emit warning about GCC version.
* Build Cinder.

(Or similarly for glibc if that's what's important)

native-api:
> Adding `CC=gcc-10` on Ubuntu 22.04 still [doesn't work](https://github.com/filips123/pyenv-cinder-testing/runs/7882340687?check_suite_focus=true). So maybe it's related to glibc version?

Can't say because the bvuild was run without `-v` so I can't see the actuall error.

The approach LGTM.

filips123:
Here are logs with verbose build: https://github.com/filips123/pyenv-cinder-testing/runs/7887409227?check_suite_focus=true#step:11:977

Issue #2436, opened 2022-08-15
Multiple entries of PYENV_ROOT prefix in .bashrc and .zshrc, causing delays when opening the terminal


arunpdl:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [ ] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [ ] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [ ] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [ ] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [ ] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [ ] Platform information (e.g. Ubuntu Linux 16.04): Manjaro Linux (21.3.6)
- [ ] OS architecture (e.g. amd64): amd64
- [ ] pyenv version: 2.3.1
- [ ] Python version: 3.10.5
- [ ] C Compiler information (e.g. gcc 7.3): 12.1.1
- [ ] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`. E.g. if you have a problem with installing Python, run `env PYENV_DEBUG=1 pyenv install -v <version> 2>&1 | tee trace.log` (note the `-v` option to `pyenv install`).

There seems to be multiple entries of prefix being created on my ~/.bashrc and ~/.zshrc.
`
export PYENV_ROOT="$HOME/.pyenv"
command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init -)"

export PYENV_ROOT="$HOME/.pyenv"
command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init -)"

[...]

PR #2437, opened 2022-08-15
Update README for clarity.

janithpet:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2436

### Description
Similar to #2431; updated instructions to be clear that the commands should be executed, rather than copied.


native-api (commented):
That's overboard IMO. I was thinking about a single note in the Zsh section like your PR and just mentioning to "run" it.

Let's wait for a reply from the affected party first. Confusions like this are rare, maybe the explanation was enough.

Issue #2440, opened 2022-08-23
Add MicroPython 1.19.1


liminspace:
It would be nice to add supporting MicroPython 1.19.1.

PR #2443, opened 2022-08-28
Add micropython 1.18 and 1.19.1

dmitriy-serdyuk:

### Prerequisite
  * N/A
  * N/A
  - Closes https://github.com/pyenv/pyenv/issues/2440

### Description
This commit will add fresh versions of microptyhon: 1.18 and 1.19.1.

### Tests
  * N/A

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -899,7 +899,7 @@ build_package_micropython() {
     "$MAKE" $MAKE_OPTS
     cd ../ports/unix
     "$MAKE" $MAKE_OPTS axtls
-    "$MAKE" $MAKE_OPTS CFLAGS_EXTRA="-DMICROPY_PY_SYS_PATH_DEFAULT='\"${PREFIX_PATH}/lib/micropython\"' $CFLAGS_EXTRA"
+    "$MAKE" $MAKE_OPTS CFLAGS_EXTRA="-DMICROPY_PY_SYS_PATH_DEFAULT='\".frozen:${PREFIX_PATH}/lib/micropython\"' $CFLAGS_EXTRA"
     "$MAKE" install $MAKE_INSTALL_OPTS PREFIX="${PREFIX_PATH}"
     ln -fs micropython "${PREFIX_PATH}/bin/python"
     mkdir -p "${PREFIX_PATH}/lib/micropython"
--- a/plugins/python-build/share/python-build/micropython-1.18
+++ b/plugins/python-build/share/python-build/micropython-1.18
@@ -1,0 +1,4 @@
+has_tar_xz_support \
+  && { install=install_package; src="https://micropython.org/resources/source/micropython-1.18.tar.xz#96fc71b42ed331c64e1adc5a830ec4f29f2975c23e8751109c03f32b80fa3eb4"; } \
+  || { install=install_zip; src="https://micropython.org/resources/source/micropython-1.18.zip#90fa8049cf275310638b9e9c77121f6042f7250b814ef622f9522befde009f57"; }
+$install micropython-1.18 "$src" micropython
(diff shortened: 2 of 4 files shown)


tomkins (changes requested):
Frozen built in modules (eg. upip) won't work with this, due to: https://github.com/pyenv/pyenv/blob/a8afc611467d2a62f8d119205943395c1b5111de/plugins/python-build/bin/python-build#L902 and https://github.com/micropython/micropython/pull/8105 (and other commits/PRs)

pyenv MicroPython 1.19.1:

```
$ micropython
MicroPython v1.19.1 on 2022-08-29; darwin [GCC 4.2.1] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import sys
>>> sys.path
['', '/Users/tomkins/.pyenv/versions/micropython-1.19.1/lib/micropython']
>>>

$ micropython -m upip
ImportError: no module named 'upip'
```

Homebrew MicroPython 1.19.1:

```
$ /opt/homebrew/bin/micropython
MicroPython v1.19.1 on 2022-06-17; darwin [GCC 4.2.1] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import sys
>>> sys.path
['', '.frozen', '/Users/tomkins/.micropython/lib', '/usr/lib/micropython']
>>>

$ /opt/homebrew/bin/micropython -m upip
upip - Simple PyPI package manager for MicroPython
Usage: micropython -m upip install [-p <path>] <package>... | -r <requirements.txt>
import upip; upip.install(package_or_list, [<path>])

If <path> isn't given, packages will be installed to sys.path[1], or
sys.path[2] if the former is .frozen (path can be set from MICROPYPATH
environment variable if supported).
Default install path: /Users/tomkins/.micropython/lib

Note: only MicroPython packages (usually, named micropython-*) are supported
for installation, upip does not support arbitrary code in setup.py.
```

After tweaking `DMICROPY_PY_SYS_PATH_DEFAULT` in `python-build` to add `.frozen` to the path - upip worked fine.

dmitriy-serdyuk:
Thanks! Added this change: now upip works. Just in case, I also tested older versions -- everything works.

plugins/python-build/bin/python-build
```diff
@@ -899,7 +899,7 @@ build_package_micropython() {
     "$MAKE" $MAKE_OPTS
     cd ../ports/unix
     "$MAKE" $MAKE_OPTS axtls
-    "$MAKE" $MAKE_OPTS CFLAGS_EXTRA="-DMICROPY_PY_SYS_PATH_DEFAULT='\"${PREFIX_PATH}/lib/micropython\"' $CFLAGS_EXTRA"
+    "$MAKE" $MAKE_OPTS CFLAGS_EXTRA="-DMICROPY_PY_SYS_PATH_DEFAULT='\".frozen:${PREFIX_PATH}/lib/micropython\"' $CFLAGS_EXTRA"
```
native-api, line 902: Won't this break previous versions?

If yes, the easiest way would probably be to set some kind of flag and do the corresponding thing depending on it. One way is to introduce another `build_*` function that would be setting it.

dmitriy-serdyuk: I checked, it doesn't break previous versions:

```
$ micropython
MicroPython v1.17 on 2022-08-30; linux version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import sys
>>> sys.path
['', '.frozen', '/home/serdyuk/.pyenv/versions/micropython-1.17/lib/micropython']
>>> import upip
>>>
```

dmitriy-serdyuk:
Updated the test: now it passes on a local ubuntu machine.

native-api:
The builds seem okay, the tests fail because MicroPython doesn't support `--version`. Is there an alternative?
E.g. does `-c 'import sys;print(sys.version)'` work for all releases?

PR #2449, opened 2022-09-05
docs: update Fish PATH update

gregorias:
### Prerequisite

### Description

The original instructions were added 2 years ago in https://github.com/pyenv/pyenv/commit/0f2d659732094f6bef11361dbc6261132edc63e2. Since then, [Fish released](https://github.com/fish-shell/fish-shell/releases/tag/3.2.0) [a more declarative instruction for this operation](https://fishshell.com/docs/current/cmds/fish_add_path.html), which is now the recommended way of manipulating `PATH` in Fish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,14 +324,21 @@ See [Advanced configuration](#advanced-configuration) for details and more confi
 
   - For **Fish shell**:
 
-    Execute this interactively:
+    If your Fish is older than 3.2.0, execute this interactively:
 
     ~~~ fish
     set -Ux PYENV_ROOT $HOME/.pyenv
     set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
     ~~~
 
-    And add this to `~/.config/fish/config.fish`:
+    Otherwise, execute the snippet below:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -328,7 +328,7 @@ See [Advanced configuration](#advanced-configuration) for details and more confi
 
     ~~~ fish
     set -Ux PYENV_ROOT $HOME/.pyenv
-    set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
+    fish_add_path $PYENV_ROOT/bin
```
native-api, line 331: `fish_add_path` is new in Fish 3.2.0. There are still non-EOL environments that provide earlier versions -- e.g. [Focal](https://packages.ubuntu.com/focal/fish).

So atm this can probably only be offered as an option "if you have Fish 3.2.0+".

native-api (changes requested):

README.md
gregorias: I've adjusted the instructions to offer this choice. PTAL

PR #2452, opened 2022-09-05
Clean up defining DEFINITION_CANDIDATES

cavcrosby:
### Description
- I've noticed some of my builds failing that use pyenv and it appears to be due to a syntax error around defining DEFINITION_CANDIDATES (`/opt/hostedtoolcache/pyenv/pyenv.d/install/latest.bash: line 5: syntax error near unexpected token (`). This PR aims to fix this issue and the workflows that ran on my pyenv fork seem to support that this issue is now fixed https://github.com/cavcrosby/pyenv/actions.

### DIFF
--- a/pyenv.d/install/latest.bash
+++ b/pyenv.d/install/latest.bash
@@ -1,13 +1,16 @@
 DEFINITION_PREFIX="${DEFINITION%%:*}"
 DEFINITION_TYPE="${DEFINITION_PREFIX%%-*}" # TODO: support non-CPython versions
 if [[ "${DEFINITION}" != "${DEFINITION_PREFIX}" ]]; then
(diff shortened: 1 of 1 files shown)


native-api (commented):
The error has been fixed in https://github.com/pyenv/pyenv/commit/23559ee6d10c5c702f924803b81eaa54a776b360

This construct looks like an unnecessary complication. What does it achieve, exactly, that the array initializer doesn't?

cavcrosby:
Good to hear!

Yea, I believe at this point I'd have to agree that your changes in https://github.com/pyenv/pyenv/commit/23559ee6d10c5c702f924803b81eaa54a776b360 should suffice. Hence I will close this PR.

PR #2456, opened 2022-09-07
Add CPython security releases 3.7.14, 3.8.14 and 3.9.14

edgarrmondragon:

### Prerequisite

### Description

- https://www.python.org/downloads/release/python-3714
- https://www.python.org/downloads/release/python-3814
- https://www.python.org/downloads/release/python-3914

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/3.7.14
+++ b/plugins/python-build/share/python-build/3.7.14
@@ -1,0 +1,9 @@
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-1.1.1n" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" mac_openssl --if has_broken_mac_openssl
+install_package "readline-8.1" "https://ftpmirror.gnu.org/readline/readline-8.1.tar.gz#f8ceb4ee131e3232226a17f51b164afc46cd0b9e6cef344be87c65962cb82b02" mac_readline --if has_broken_mac_readline
+if has_tar_xz_support; then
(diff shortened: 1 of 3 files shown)


native-api:
Do the patches we provide for the last releases in those release lines no longer apply?

edgarrmondragon:
@native-api do you mean something like what's in `pyenv/plugins/python-build/share/python-build/patches/3.7.13/Python-3.7.13/`?

I'm not familiar with the patches or how they are applied. Should I be looking at the CPython commit history between tags to get those patches?

native-api:
See https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md#authoring-installation-scripts .

The easiest way is probably to just try to apply those patches to the new codebase. If a patch doesn't apply, then if it's a backport, the probem is most likely fixed in the new release; if it's a user-provided extended support patch, it needs to be reworked which is the responsibility of a party interested in that extended support.

ReillyBrogan:
@edgarrmondragon Do this (more or less):

```
git clone https://github.com/python/cpython.git && cd cpython
git checkout v3.7.14
git am /path/to/pyenv/plugins/python-build/share/python-build/patches/3.7.13/Python-3.7.13/0001-Port-ctypes-and-system-libffi-patches-for-arm64-macO.patch
git am /path/to/pyenv/plugins/python-build/share/python-build/patches/3.7.13/Python-3.7.13/0002-bpo-41100-fix-_decimal-for-arm64-Mac-OS-GH-21228.patch
git am /path/to/pyenv/plugins/python-build/share/python-build/patches/3.7.13/Python-3.7.13/0003-bpo-42351-Avoid-error-when-opening-header-with-non-U.patch
```

If those apply without errors then most likely those patches are still needed in 3.7.14 (IE they were not fixed between [3.7.13 and 3.7.14](https://github.com/python/cpython/compare/v3.7.13...v3.7.14)) and you should create the `pyenv/plugins/python-build/share/python-build/patches/3.7.14/Python-3.7.14/` and copy the three patches to it. Then test that you can still install and build 3.7.14 with your local branch (and otherwise spot test it).

Luckily it seems like 3.8.14 and 3.9.14 don't have any patches that need to be checked and so those _should_ be fine.

Disclosure: I'm not involved with this project and my advice may be incorrect. I was just wondering why 3.9.14 wasn't yet available and so found this PR and researched what needed to be done.

PR #2458, opened 2022-09-12
Add CPython 3.11.0rc2

ViktorHaag:
3.11.0rc2 was released today (September 12th)

https://www.python.org/downloads/release/python-3110rc2/

PR #2459, opened 2022-09-12
Add CPython 3.11.0rc2

ViktorHaag:
3.11.0rc2 was released today (September 12th)

https://www.python.org/downloads/release/python-3110rc2/

(Fixes earlier PR #2458 based on feedback from @saaketp)


smrde:
Can you please add 3.10.7 as well? It was released on Sept. 6, 2022.
https://www.python.org/downloads/release/python-3107/

ViktorHaag:
> Can you please add 3.10.7 as well? It was released on Sept. 6, 2022.
> https://www.python.org/downloads/release/python-3107/

I believe that it has already been added with PR #2454 .

PR #2463, opened 2022-09-14
Add patches for 3.7.14 needed for Apple Silicon

samdoran:
### Prerequisite
  - None

### Description
Mostly the same patches from 3.7.13 are needed for 3.7.14 on Apple Silicon.

### Tests


native-api:
Could you also port patches [for the recently-added 3.8.14 and 3.9.14](https://github.com/pyenv/pyenv/pull/2456)?

samdoran:
I'd be happy to, but I don't think those patches are needed for 3.8.14 and 3.9.14. I looked through the CPython source and the changes are already present in those versions and the need for a `VARIADIC` flag was solved in another way (based on reading the comments in the source).

Also, I've installed both 3.8.14 and 3.9.14 and `ctypes` compiled fine.

samdoran:
Thank you for merging!

Issue #1715, opened 2020-10-11
python-build: option to disable Homebrew library usage


uranusjr:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): macOS Mojave
- [x] OS architecture (e.g. amd64): x86_64
- [x] pyenv version: 1.2.21
- [x] Python version: none
- [x] C Compiler information (e.g. gcc 7.3): clang-1200.0.32.2
- [x] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`
  * *(not relevent)*

Homebrew installs each version of a library under different subdirectories (named after the version), and only links the lib into place. [It also automatically removes old versions when installing new ones.](https://docs.brew.sh/FAQ#how-can-i-keep-old-versions-of-a-formula-when-upgrading) This makes it problematic to link against unversioned libraries (say readline), since the .so linked against at compile time may disappear at any time, causing the functionality to be broken once Homebrew packages are upgraded.

[...]

Issue #1686, opened 2020-09-03
Stop looking for Homebrew openssl?


brandonfriess-stripe:
I'm curious if this project would have any interest in moving away from using openssl from homebrew by default? Similar to what ruby-build has already [done](https://github.com/rbenv/ruby-build/pull/1375).

While the dynamic linking to brew installed openssl versions at `brew --prefix openssl@1.1` is nice, it will eventually break python installs whenever openssl bumps their formula version and a user brew upgrades it (or another package that declares a dependency on openssl updates it).

It appears like most of the code in python-build would support this change pretty easily. I replicated the changes from the ruby-build PR linked above and was able to install python versions successfully. Admittedly, I haven't tested this extensively.

Thoughts?

tomy0000000:
I've recently also encountered an simlilar issue when homebrew bumped OpenSSL to version 1.1, the old pyenv-compiled python failed to resolve the correct path.

For those who came across to this problem, you need to reinstall **all** compiled python with pyenv which build a new python  with the correct openssl path.

However, I think using homebrew OpenSSL is still a good idea, reasons as follow:
* You shouldn't be using system OpenSSL, as you wouldn't use system builtin Python, either.
* Managing a correct version of OpenSSL is probably beyond the scope of pyenv.

A good way to prevent link breaks from homebrew upgrading is to bind to a _correct_ directory.

```bash
$ brew --prefix openssl@1.1
/usr/local/opt/openssl@1.1
$ readlink /usr/local/opt/openssl@1.1
../Cellar/openssl@1.1/1.1.1g
```

If you look deeper, you'll find

```bash
$ ls -l /usr/local/opt | grep openssl
lrwxr-xr-x  1 tomy0000000  admin  28  9  8 00:09 openssl -> ../Cellar/openssl@1.1/1.1.1g
lrwxr-xr-x  1 tomy0000000  admin  28  9  8 00:09 openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1g
```

Therefore, a possible solution might be
1. Use exists method to verify that homebrew has installed `openssl`, `openssl@1.1`, or any future release
2. Use `/usr/local/opt/openssl` as OpenSSL path to build python

This method guaranteed that whenever there's a new version of OpenSSL, as long as homebrew point the symlink to the latest version, the old python is able to resolve with that symlink.

brandonfriess-stripe:
I think it's problematic using Homebrew openssl for a few reasons (the below is taken from ruby-build):

[...]

PR #2464, opened 2022-09-14
Add ability to easily skip all use of Homebrew

samdoran:
### Prerequisite
  - None

### Description

This has been discussed and turned down before ( #1686, #1715) and the [suggestion](https://github.com/pyenv/pyenv/issues/1715#issuecomment-929531904) was to make `brew` a noop. That's what I do but it would be nice if there was a simpler way for `pyenv` to handle this. Having to make `brew` a noop, then fix it for normal use is not hard but it gets tiresome (I compile Python a lot 🤓).

I came up with a solution using an environment variable `PYENV_DISABLE_HOMEBREW`. I wanted to open a PR and discuss adding this to `pyenv`.

If it's something that would be considered, I'd be happy to work on documentation and tests.

### Tests
- [ ] None yet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,15 @@
-.PHONY: test test-build
+.PHONY: test test-build test-unit test-plugin
 
 # Do not pass in user flags to build tests.
 unexport PYTHON_CFLAGS
 unexport PYTHON_CONFIGURE_OPTS
 
-test: bats
+test: test-unit test-plugin
+
+test-unit: bats
 	PATH="./bats/bin:$$PATH" test/run
+	
+test-plugin: bats
 	cd plugins/python-build && $(PWD)/bats/bin/bats $${CI:+--tap} test
 
 PYTHON_BUILD_ROOT := $(CURDIR)/plugins/python-build
--- a/plugins/python-build/README.md
+++ b/plugins/python-build/README.md
@@ -113,6 +113,7 @@ You can set certain environment variables to control the build process.
   checksum of the file to the mirror URL.
 * `PYTHON_BUILD_SKIP_MIRROR`, if set, forces python-build to download packages from
   their original source URLs instead of using a mirror.
+* `PYTHON_BUILD_SKIP_HOMEBREW`, if set, will not search for libraries installed by Homebrew on macOS.
 * `PYTHON_BUILD_ROOT` overrides the default location from where build definitions
   in `share/python-build/` are looked up.
 * `PYTHON_BUILD_DEFINITIONS` can be a list of colon-separated paths that get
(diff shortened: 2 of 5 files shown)


samdoran:
cc @sivel

native-api:
Where are you going to take deps from in MacOS if not from Homebrew?

Pyenv can compile OpenSSL and Readline from source and take Zlib from XCode SDK. What about others?

native-api:
As an alternative solution, there's a 3rd-party plugin, https://github.com/sprout42/pyenv-fix-version, that recompiles installed Pythons if the underlying deps have been upgraded.

native-api:
> Where are you going to take deps from in MacOS if not from Homebrew?
> 
> Pyenv can compile OpenSSL and Readline from source and take Zlib from XCode SDK. What about others?

AFAICS, according to https://news.ycombinator.com/item?id=9407119, there's smth called "pkgsrc" which is stable release instead of rolling release.

native-api:
The breakage you're describing is what you get when you're dealing with rolling-release stuff. You _have_ to sync all your dependent software whenever you do an upgrade or you risk this.

So it's either
* switch to a stable-release product (and miss out on new versions of libs that might be needed for new versions of Python), or
* introduce extra complexity and sync all your stuff (which is Pythons and their Pip-installed extension modules) to the rolling-release ecosystem at every upgrade

sivel:
> Where are you going to take deps from in MacOS if not from Homebrew?
> 
> Pyenv can compile OpenSSL and Readline from source and take Zlib from XCode SDK. What about others?

What does it matter where they come from? If a user wants more control over handling deps, give them a way to do so.  It's not the default behavior, nor should you be concerned about offering instructions to users that want a way to prevent using homebrew.

Simply, the desire is to give the user the option to handle the deps themselves, in whatever way they desire.  Just because homebrew is installed, and I am running macos, doesn't mean I want pyenv to magically assume that I want to use homebrew for the deps.

native-api:
> What does it matter where they come from? If a user wants more control over handling deps, give them a way to do so. It's not the default behavior, nor should you be concerned about offering instructions to users that want a way to prevent using homebrew.
> 
> Simply, the desire is to give the user the option to handle the deps themselves, in whatever way they desire. Just because homebrew is installed, and I am running macos, doesn't mean I want pyenv to magically assume that I want to use homebrew for the deps.

Okay, that makes sense. The normal software build behavior is to take whatever is available on the configured/default search paths -- and let the user arrange them as needed. But the Homebrew search behavior is not a part of that.

My primary concern was that there's simply nothing else realistically available in MacOS to get the dependencies from that is any better than Homebrew for your use case (which would make this use case rather unrealistic thus probably not worth supporting). But now that we see there is, that's not a problem.

---

The envvars would need to be prefixed with `PYTHON_BUILD_` like other envs that affect the build (see https://github.com/pyenv/pyenv/tree/master/plugins/python-build#readme).

samdoran:
I'll change the var name and add docs.

samdoran:
Thank you so much!

PR #2480, opened 2022-10-11
Add CPython 3.10.8

MuXodious:

### Prerequisite

### Description
- [x] Released on 11/10/2022
- [x] https://www.python.org/downloads/release/python-3108/

### Tests


tobixx:
Hi, may I ask if there is a reason why not also 3.9.15, 3.8.15, 3.7.15 was added ?

MuXodious:
> Hi, may I ask if there is a reason why not also 3.9.15, 3.8.15, 3.7.15 was added ?

Nope, there are no spesific reasons not to add any of those versions. I personally needed 3.10.8 and then created this commit. You can add them yourself by following [CONTRIBUTING.md](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md). I can try adding them myself too, if nobody else does, once I have spare time. ✌️

adamchainz:
I've opened #2482 to add the three missed releases.

NEXT-JP:
Seems like the comment from @tobixx underscores a question I've always had; Why hasn't the process of adding new releases of python versions and creating a new PyEnv release been automated. It always seems to take a few weeks for the latest Python version to be available in the latest PyEnv release.

PR #2482, opened 2022-10-12
Add CPython 3.7.15, 3.8.15, and 3.9.15

adamchainz:
### Prerequisite
n/a

### Description

Add new versions announced yesterday: [3.7.15](https://www.python.org/downloads/release/python-3715/), [3.8.15](https://www.python.org/downloads/release/python-3815/), and [3.9.15](https://www.python.org/downloads/release/python-3915/).

3.10.8, also released yesterday, was merged in #2480.

### Tests

Computed SHA's with `curl <url> | gsha256sum -`.


adamchainz:
Ah, I missed the patches. Thanks @native-api ! 🤘

PR #2487, opened 2022-10-17
Auto-resolve prefixes to the latest version

native-api:

### Prerequisite

### Description

When given a prefix rather than full name, Pyenv resolves it to the latest release in the corresponding version line.

It must be a full prefix -- the searched prefix is `<prefix>[-.]`

Other flavors may be sorted incorrectly atm.

### Tests


BrandonLWhite:
Awesome feature.  This has been the missing piece in pyenv.  Thank you for implementing this!

PR #2507, opened 2022-10-31
Fix mismatched quote in `latest` subcommand

neersighted:
As per title, there is a mismatched backtick and single quote. I have chosen the single quote as the more common convention.


native-api:
Those are [traditional UNIX quotes](https://unix.stackexchange.com/questions/85303/where-does-this-convention-for-quotation-marks-come-from). We use them elsewhere in the codebase, too.

neersighted:
Interesting, I don't think I've ever encountered this with **single** quotes. Fair enough, however.

PR #2511, opened 2022-11-03
GitHub Workflows security hardening

sashashura:

### Prerequisite

### Description

This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### Tests


native-api:
The flip side is we'll need to constantly figure out what each permission controls -- thus which permissions we need for each operation.
Whle the documentation at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions, https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token and https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#administration is extensive, as usual, it lacks fine details.

E.g.:
* Why did you grant any permissions to the job if it doesn't use the Github REST API, at all?
* What permissions do we need to e.g. create a PR or a release?
* Do we need any permissions for operations that stem from the workflow definition -- e.g. to run a dependent action or pass data to it?

---

How probable and severe are the risks anyway? E.g.: 
* We have workflow start protection for new contributors to filter out bad agents
* Actions' token doesn't have permissions for Administration so even injecting malicious code can't hijack the project or the organization.

Let's compose a basic [risk heat map](https://www.balbix.com/insights/cyber-risk-heat-map/) so that we can decide which risks we defend against and which we don't.
(Since this requires a colored table, a Google spreadsheet would probably be the most convenient.)

native-api:
> After the workflows run with minimal privileges it requires modification only if you add a job or step that requires new permission.

That's the problem. We do currently plan to add new workflows:
* Detect new Python releases, create PRs and probably merge them, too.
* Create releases recently after a new CPython release comes and bump Homebrew formula

native-api:
So we can merge this, it won't hurt anything now, but whenever we start tinkering with workflow, it'll probably start causing problems and will be removed since it's wasted time and nerves for questionable gain...

native-api:
@pyenv/pyenv-core-committers , any thoughts on this?

native-api:
Okay, in it goes.

@sashashura , thanks for adding explanations why the permissons are needed!

PR #2518, opened 2022-11-08
Add simple `.editorconfig` file

aphedges:

### Prerequisite

### Description

I saw 036fd63bbd9fea17f3363a511a264e7eaf47ce88, and I was inspired to create a very rudimentary and uncontroversial [`.editorconfig`](https://editorconfig.org) file. Although it isn't a strict check, it should limit newline problems in the future. More can be added as desired, but I don't want to impose a style that isn't already present.

### Tests

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+# Makefiles always use tabs for indentation
+[Makefile]
+indent_style = tab
+indent_size = unset # Allow user-defined tab width


native-api (commented):
How prevalent is this format among common editors?

aphedges:
I'd say it's decently common. There are lists of editors that directly support it (https://editorconfig.org/#pre-installed) and that have plugins for it (https://editorconfig.org/#download), but I'm not sure how much these various editors are used or how often the plugins are installed.

`.editorconfig` files are definitely used, though. For example, the CPython repository has its own [`.editorconfig`](https://github.com/python/cpython/blob/57be5459593bbd09583317ebdafc4d58ae51dbf4/.editorconfig).

Ultimately, style needs to be enforced by CI/code review, but `.editorconfig` at least makes it easier to write in that style.

Issue #2494, opened 2022-10-24
python 3.11.0 fails to build on macOS (OpenSSL build failure)


minddrive:
Platform: macOS 12.6
OS architecture: x86_64 (Intel)
pyenv version: 2.3.5-3-g0726e02e
Python version: 3.11.0
C Compiler information: Apple clang version 14.0.0 (clang-1400.0.29.102)

Building the new release of Python 3.11.0 fails due to a failure for the build of OpenSSL 1.1.1q:

```
Installing openssl-1.1.1q...

BUILD FAILED (OS X 12.6 using python-build 2.3.5-3-g0726e02e)

Inspect or clean up the working tree at /var/folders/d7/04qpwbhs5mg_m5rycyyk_3x40000gp/T/python-build.20221024151009.558
Results logged to /var/folders/d7/04qpwbhs5mg_m5rycyyk_3x40000gp/T/python-build.20221024151009.558.log

Last 10 log lines:
                 apps/libapps.a -lssl test/libtestutil.a -lcrypto
ld: warning: directory not found for option '-L/Users/relayer/.pyenv/versions/3.11.0/lib'
clang  -Iinclude -arch x86_64 -O3 -Wall -D_REENTRANT -DZLIB -DZLIB_SHARED -DNDEBUG -I/Users/relayer/.pyenv/versions/3.11.0/include  -MMD -MF test/v3ext.d.tmp -MT test/v3ext.o -c -o test/v3ext.o test/v3ext.c
test/v3ext.c:201:24: error: implicitly declaring library function 'memcmp' with type 'int (const void *, const void *, unsigned long)' [-Werror,-Wimplicit-function-declaration]
        if (!TEST_true(memcmp(ip1->data, ip2->data, ip1->length) <= 0))
                       ^
test/v3ext.c:201:24: note: include the header <string.h> or explicitly provide a declaration for 'memcmp'
1 error generated.
make[1]: *** [test/v3ext.o] Error 1
make: *** [all] Error 2
```

This is actually a long known issue: https://github.com/openssl/openssl/issues/18720 .  The fix isn't planned until 1.1.1r (likely several months away), but there's a simple patch (a single include line) that can fix this.  Might it be possible to get this added to correct the problem?  (If I knew of the process, I could offer a PR, if needed.)

minddrive:
[...]

PR #2520, opened 2022-11-08
Patch openssl 1.1.1q on macos for Python 3.8.15 and 3.9.15

twangboy:

### Prerequisite
  - Additional fixes for: https://github.com/pyenv/pyenv/issues/2494

### Description
Python updated openssl to version 1.1.1q in 3.8.15, 3.9.15 and 3.11. Python 3.10.7 is at 1.1.1o.
The following PR added a patch for building openssl 1.1.1q on 3.11 but not for 3.8.15 and 3.9.15.
https://github.com/pyenv/pyenv/pull/2500
This applies the openssl 1.1.1q patches to the 3.8.15 and 3.9.15 builds of python.

### Tests
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 /src/*.o
 /bats/
 /default-packages
+.idea
+*.un~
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@
 /src/*.o
 /bats/
 /default-packages
+.idea
+*.un~
```
native-api, line 12: What's this?

amureki: Extending project's `.gitignore` with personal OS/tooling-related files and paths is usually not recommended practice.
Especially in this exact case, it is totally unrelated to the patch you are providing.

I suggest you configuring global `.gitignore` file for your workstation:
https://adamj.eu/tech/2022/11/06/git-set-up-a-global-ignore-file/

native-api: I think IDEA and Vim are used widely enough to warrant adding.

It's okay to address additional concerns in a PR (as long as it's within reason, or course), but then each concern needs to be in its own commit (see the [contribution giude](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md)).

PR #2526, opened 2022-11-15
Add --no-push-path option

### LINKED ISSUE #2389: v2.3.0 (specifically #2310) breaks interactive (non-login) shell usage by prepending shims to PATH
If I understand correctly, #2310 largely undoes what #1898 did.
Neither clearly specified a migration path (being breaking changes), causing issues for users...

The issue is that now, in interactive non-login shell `pyenv init -` modifies `PATH` (to add shims directory), which is undesirable and improper shell usage (what #1898 implemented correctly).
This is a breaking change. For example, it breaks `pipenv` shell if used with `pipenv shell --fancy` (which assumes _correct_ shell configuration) - by prepending shims directory which breaks virtualenv.

With #2310, is there a way to implement split configuration, i.e. have pyenv init not touch `PATH`? What is the correct configuration, what did I miss?

P.S. I do not mean to sound offensive, although I am a bit mad.

isaacl:
Add option to skip PATH shim push to front.  This is useful when one uses pyenv --path in a login script, and wants to init pyenv functions in each shell (i.e. zshrc).  However, in certain cases, like when using pipenv, one needs zshrc to NOT push the pyenv shims to the front of the path.

### Prerequisite

### Description
-  Add option --no-path-update to skip PATH var manipulation.
-  Change `refute_line` to look for substrings matches
    This fixes, for example, the --no_hash test, which was automatically passing 

### Tests
- [x] My PR adds the following unit tests (if any):
Tests for the new --no-push-path option.

Fixes #2389

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -151,10 +151,16 @@ function print_path() {
   case "$shell" in
     fish )
       echo 'while set index (contains -i -- '\'"${PYENV_ROOT}/shims"\'' $PATH)'
-      if [ -n "$no_push_path" ]; then echo 'if test -z "$index"'; fi
-      echo 'set -eg PATH[$index]; end; set -e index'
-      echo 'set -gx PATH '\'"${PYENV_ROOT}/shims"\'' $PATH'
-      if [ -n "$no_push_path" ]; then echo 'end'; fi
+      if [ -n "$no_push_path" ]; then
+        echo 'set found_pyenv_shim 1; break; end';
+        echo 'if test -z "$found_pyenv_shim"'
+        echo 'set -gx PATH '\'"${PYENV_ROOT}/shims"\'' $PATH'
+        echo 'else; set -e found_pyenv_shim; end'
+      else
+        echo 'set -eg PATH[$index]; end;'
+        echo 'set -gx PATH '\'"${PYENV_ROOT}/shims"\'' $PATH'
+      fi
+      echo 'set -e index'
       ;;
     * )
       # Some distros (notably Debian-based) set Bash's SSH_SOURCE_BASHRC compilation option
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -154,11 +154,11 @@ function print_path() {
     case "$shell" in
       fish )
           echo 'if not contains -- "'"${PYENV_ROOT}/shims"'" $PATH'
-          echo '  set -gx PATH "'"${PYENV_ROOT}/shims"'" $PATH'
+          echo '  set -gx PATH '\'"${PYENV_ROOT}/shims"\''" $PATH'
           echo 'end'
           ;;
       * )
-          echo 'if [[ ":$PATH:" != *":'"${PYENV_ROOT}"/shims':"* ]]; then'
+          echo 'if [[ ":$PATH:" != *'\':"${PYENV_ROOT}"/shims:\''* ]]; then'
           echo '  export PATH="'"${PYENV_ROOT}"'/shims:${PATH}"'
           echo 'fi'
           ;;
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -150,17 +150,17 @@ function init_dirs() {
 function print_path() {
   case "$shell" in
     fish )
-      echo 'while set index (contains -i -- '\'"${PYENV_ROOT}/shims"\'' $PATH)'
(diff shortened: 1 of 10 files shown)


native-api:
With `--no-path-update`, what does `eval "$(pyenv init --path)"` do, exactly? It seems to me that nothing :slightly_smiling_face: 

Another user has had the same probem as you in https://github.com/pyenv/pyenv/issues/2389 , and the way we were intend to do to fix that is to add a switch to not _push_ shims to the start of PATH if they're already there.

isaacl:
I want the shims to not get pushed to the front when in a pipenv shell... Would your solution work for this?

> With --no-path-update, what does eval "$(pyenv init --path --no-path-update)" do, exactly?

That does nothing, indeed.  I could make it an error...

native-api:
> I want the shims to not get pushed to the front when in a pipenv shell... Would your solution work for this?

Yes. As for the name, my best idea is "`--no-push-path`".

isaacl:
OK, I rewrote the patch based on --no-push-path.   I have not extensively tested this, however.

isaacl:
Note the bash check is not perfect -- I think in theory the subshell could produce a differently quoted path, which would then be detected as 'different'.  Open to patches or ideas. 

I've already fixed this for myself locally by patching the output of pyenv:

```bash
eval "$(pyenv init - | sed -n '/PYENV_SHELL/,$p')"
```

I'm just trying to be a good citizen and produce something useful for others...

libexec/pyenv-init
```diff
@@ -143,9 +150,17 @@ function init_dirs() {
...
+      if [ -n "$no_push_path" ]; then
+        echo 'set found_pyenv_shim 1; break; end';
+        echo 'if test -z "$found_pyenv_shim"'
+        echo 'set -gx PATH '\'"${PYENV_ROOT}/shims"\'' $PATH'
+        echo 'else; set -e found_pyenv_shim; end'
+      else
+        echo 'set -eg PATH[$pyenv_index]; end;'
+        echo 'set -gx PATH '\'"${PYENV_ROOT}/shims"\'' $PATH'
+      fi
+      echo 'set -e pyenv_index'
```
native-api, line 163: This looks overcomplicated. It's probably better to split this into two chunks for either case.

libexec/pyenv-init
```diff
@@ -155,12 +170,20 @@ function print_path() {
...
       echo 'fi; done; '
       echo 'echo "${paths[*]}"'\'')"'
-      echo 'export PATH="'"${PYENV_ROOT}"'/shims:${PATH}"'
+      # if no_push_path is unset, always update path.
+      # else, check for differences between the PATH variable and our proposed path var (pyenv_new_path)
+      # we check for differences by using bash string substitution to remove all chars but ":"
+      # and checking for equality -- i.e. ensuring old and new paths have same number of components.
+      # If they are equal, it means pyenv shims were NOT found, and therefore need to be added.
+      if [ -z "$no_push_path" -o "x${PATH//[^:]/}" = "x${pyenv_new_path//[^:]/}" ]; then
+        echo 'export PATH="'"${PYENV_ROOT}"'/shims:${pyenv_new_path}"'
```
native-api, line 184: Same here. You can check `$PATH` for an entry with `[[ :$PATH: == *":<entry>:"* ]]` .

isaacl: I agree it's complicated.

I suppose I could do your suggested contains check earlier (if no_push_path is set).  I wasn't sure if we could use `[[` `]]` in this file (since I didn't see them elsewhere).

libexec/pyenv-init
isaacl: Yes, it's complicated.  I'm not sure what splitting into chunks accomplishes, however -- don't both branches need the same while loop?

native-api: The `--no-push` version doesn't need a loop at all, only a one-time condition check.

isaacl (commented):
Please take another look!  simplified bash code.

isaacl:
Sorry for the test failures.  I'm running tests locally now and verified they passed.

I rewrote the `--no-push-path` tests to check for correct path manipulation logic rather than looking for specific bash lines.

native-api:
I've eliminated redundant printed logic and replaced unit tests with integration tests; `refute_line` changes are now unneeded.

native-api:
Thank you for your efforts! :rocket:

Issue #1675, opened 2020-08-10
pyenv local mechanism not working in Jupyter terminal


mwiebusch78:
I am using pyenv 1.2.20 on Mac OS X 10.15.5 and I have two Python versions (3.6.11 and 3.8.5) installed via pyenv. The global python version is 3.8.5 and I have a project folder myproject configured with pyenv local to use version 3.6.11 instead. To enable automatic switching between python versions I'm executing `eval "$(pyenv init -)"` both in my `.bashrc` and `.bash_profile.` All this works fine when I use the Mac terminal emulator. `python --version` gives 3.6.11 when I'm inside myproject and 3.8.5 anywhere else.

However, if I start a jupyterlab server (outside of myproject) and use the terminal emulator from jupyterlab the python version I see is always 3.8.5, even if I'm inside myproject. Even if I manually run `eval "$(pyenv init -)"` in the jupyterlab terminal the switching still does not work.

I have noticed that the $PATH variable I see in the Mac OS terminal is different from the one in the jupyterlab terminal. The former is

```
/Users/wiebuschm/.local/opt/google-cloud-sdk/bin:/usr/local/Cellar/pyenv-virtualenv/1.1.5/shims:/Users/wiebuschm/.pyenv/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```

while the latter is

```
/usr/local/Cellar/pyenv-virtualenv/1.1.5/shims:/Users/wiebuschm/.pyenv/shims:/Users/wiebuschm/.pyenv/versions/3.8.5/bin:/usr/local/Cellar/pyenv/1.2.20/libexec:/Users/wiebuschm/.local/opt/google-cloud-sdk/bin:/usr/local/Cellar/pyenv-virtualenv/1.1.5/shims:/Users/wiebuschm/.pyenv/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```

However, I don't know what causes this difference and even if I manually set PATH in the jupyterlab terminal to be the same as in the Mac OS terminal it still won't automatically switch the Python version when I cd into myproject.

thealtruist:
@mwiebusch78 I had the same problem but I just discovered a workaround: 

1. Manually set the PYENV_VERSION environment variable
`PYENV_VERSION=3.6.11`
2. Create the virtual environment like normal using venv
`python -m venv venv36`

native-api:
[...]

PR #2531, opened 2022-11-20
Don't set `PYENV_VERSION` in the environment of the invoked commmand

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/1675

### Description

See https://github.com/pyenv/pyenv/issues/1675#issuecomment-1263585074 for the issue description and https://github.com/pyenv/pyenv/issues/1675#issuecomment-1263652478 for explanation.

> The reason is that Pyenv sets the `PYENV_VERSION` envvar in the environment of the executed command.

That was added in https://github.com/rbenv/rbenv/commit/8ee2f2657a088851d0aa75736c7b0305a10522f1 to ensure the selected version is used when PATH is not changed by `pyenv-exec` (e.g. for the system version).

The envvar ensures that the called subcommands are working with the correct version. AFAICS, it's not _strictly_ needed but it steamlines things -- e.g. avoids a repeated call to `pyenv-version-name`.
But when invoking the command itself, the changed PATH ensures that the right executables are getting the priority.

However, we only add one PATH entry now, for the leading version. If multiple alt versions are selected, we need to add them all to ensure correct selection with just PATH.
If `system` occurs in the middle of the list, we'll probably have to add the following entries to the _tail_ of PATH.
And probably remove `shims` from PATH, for good measure.

Team, do you think we should fix the PATH manipulation now, or wait until users complain, at which point we'll be able to obtain details on the intended behavior?

### Tests

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -24,7 +24,7 @@ jobs:
       #    python-version: ${{ matrix.python-version }}
       # ... but in the repo, we want to test pyenv builds on Ubuntu
       - run: |
-          sudo apt-get install -yq make build-essential libssl-dev zlib1g-dev \
+          sudo apt update && sudo apt install -yq make build-essential libssl-dev zlib1g-dev \
           libbz2-dev libreadline-dev libsqlite3-dev curl \
           libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
       # https://github.com/pyenv/pyenv#installation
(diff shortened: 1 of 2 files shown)


anton-petrov (commented):
I don't think we need to make changes to path handling because of a very specific case. I also use jupiter, and I don't see a global problem in this case - it's always better to use a native terminal instead of its web version. For cases when, for example, jupiter hub is installed and many users connectedt to the server, pyenv is not commonly used, and only one version of python is installed. For such environment it is better to install Anaconda/Miniconda or compile CPython manually. But it's already devops territory, so... Еhus, I propose to think about the best solution and do not change anything for now.

PR #2535, opened 2022-11-28
Fix typo in README.md

weensy:
Fix markdown because there was a space at the beginning of the `git clone` command


weensy:
Previously displayed readme:
image

PR #2553, opened 2022-12-08
Add Pypy 7.3.10

dand-oss:
Add pypy 3.7.10 release
fix mac package name in pypy 3.7.9 release
remove windows support for pypy

### DIFF
--- a/.github/workflows/modified_scripts_build.yml
+++ b/.github/workflows/modified_scripts_build.yml
@@ -32,26 +32,35 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: |
-          brew install openssl openssl@1.1 readline sqlite3 xz zlib
-      - run: |
+          #envvars
           export PYENV_ROOT="$GITHUB_WORKSPACE"
           echo "PYENV_ROOT=$PYENV_ROOT" >> $GITHUB_ENV
           echo "$PYENV_ROOT/shims:$PYENV_ROOT/bin" >> $GITHUB_PATH
       - run: |
+          #prerequisites
+          brew install openssl openssl@1.1 readline sqlite3 xz zlib
+          if [[ "${{ matrix.python-version }}" =~ pypy.*-(src|dev) ]]; then
+            pyenv install 2.7.18
+            PYENV_VERSION=2.7.18 pip install curses genc pycparser
+          fi
+      - run: |
+          #build
           pyenv install -v ${{ matrix.python-version }}
           pyenv global ${{ matrix.python-version }}
       # Micropython doesn't support --version
-      - run: >
+      - run: |
+          #print version
           if [[ "${{ matrix.python-version }}" == "micropython-"* ]]; then
             python -c 'import sys; print(sys.version)'
           else
-            python --version;
+            python --version
             python -m pip --version
           fi
       # Micropython doesn't support sys.executable, os.path, older versions even os
       - env:
           EXPECTED_PYTHON: ${{ matrix.python-version }}
         run: |
+          #check
           if [[ "${{ matrix.python-version }}" == "micropython-"* ]]; then
             [[ $(pyenv which python) == "${{ env.PYENV_ROOT }}/versions/${{ matrix.python-version }}/bin/python" ]] || exit 1
             python -c 'import sys; assert sys.implementation.name == "micropython"'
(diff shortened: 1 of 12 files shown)


dand-oss:
Passes all tests other than for "src" builds

as for building pypy from source
the github build images for this would need many more requirements
such as cpython, cffi

and pypy build is very complex multi-stage

**https://doc.pypy.org/en/latest/build.html**

For building PyPy, we recommend installing a pre-built PyPy first (see [Downloading and Installing PyPy](https://doc.pypy.org/en/latest/install.html)). It is possible to build PyPy with CPython, but it will take a lot longer to run – depending on your architecture, between two and three times as long.

Even when using PyPy to build PyPy, translation is time-consuming – 20 minutes on a fast machine – and RAM-hungry. You will need at least 3 GB of memory on a 32-bit machine and 6GB on a 64-bit machine.

dand-oss:
ok the source builds now work
pyenv-install sets up the reqs if not present

mac-os fails to build pypy

   File "/Users/runner/work/pyenv/pyenv/versions/2.7.18/lib/python2.7/ctypes/__init__.py", line 366, in __init__
    self._handle = _dlopen(self._name, mode)
[translation:ERROR] OSError: dlopen(/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libexpat.tbd, 0x0006): tried: '/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libexpat.tbd' (not a mach-o file), '/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libexpat.1.tbd' (not a mach-o file)

ctypes loading libexapt

not gonna fix that

dand-oss:
>            PYENV_VERSION=2.7.18 pip install curses genc pycparser

I think curses is built in now, no package anymore.  That source build dep check in pyenv-install must be very old.

native-api:
> I think curses is built in now, no package anymore. That source build dep check in pyenv-install must be very old.

Okay. That said, `genc` and `pycparser` are not mentioned in https://doc.pypy.org/en/latest/build.html . But `cffi` is mentioned. Maybe that documentation is also too old? :wink:

native-api:
You also said that using PyPy as bootstrap would be much faster.
Which PyPy versio is it's better to use?

The problem with 2.7.18 is it's EOL can can break at any time in a newer environment.

dand-oss:
wow, u have good eye for detail...

pypy build COULD be out of date, only 2 guys doing most of the work, cfbolz and mattipicus.

Pypy was first done 2007 if I recall, so the rpythhon translator must still need 2.x

we could try pypy 2.7, they keep releasing it.  I have not built pypy with a pypy binary release.

cffi might be needed for shared libs in the build.  You are right, the build is "succeeding" on linux without it.

I have the pypy repo checked out.  I will look for genc and pyparser.

dand-oss:
appears to me genc and pyparser are provided by "rpython", the internal runtime for pypy.

occurs to me that pypy project wrote and provides cffi, so pypy build may bootstrapy cffi itself now.
https://foss.heptapod.net/pypy/cffi
armin rigo still working on it regularly, last commit 5 days ago

Try no deps at all, see where we get!

native-api:
Checked it locally. `pycparser` is required, others are not.

Linux seems fine (got a build "killed", probably ran out of memory).
In MacOS, it can't find libexpat for some reason even though I've installed it and set envvars. That seems to be PyPy's problem.

So I'll rebase it to clarify the history and merge.

dand-oss:
sounds like this is as good as we can do

thanks for the yaml work, I am unfamliar with github actions, if learn would be better contributor.

native-api:
I didn't yet check pypy2.7 as bootstrap, doing that now.

dand-oss:
> The problem with 2.7.18 is it's EOL and can break at any time in a newer environment. Or just when PyPI drops support...

they can't
https://twitter.com/pypyproject/status/1160209907079176192
There is no plan to deprecate the 2.7 version of PyPy, basically for as long as PyPy exists. RPython is based on Python 2, which is not really realistic to change, so we need PyPy2 to bootstrap. Documented here: [http://doc.pypy.org/en/latest/faq.html#how-long-will-pypy-support-python2](http://doc.pypy.org/en/latest/faq.html#how-long-will-pypy-support-python2)

dand-oss:
+            # brew install expat
+            # sudo apt install -yq libexpat1-dev
from    Remove no-longer-needed PyPy prerequisites
we can remove dead code before merge...

native-api:
> * ```
>              # brew install expat
>       ```
> 
>     * ```
>              # sudo apt install -yq libexpat1-dev
>       ```
> 
> 
> from Remove no-longer-needed PyPy prerequisites we can remove dead code before merge...

Already

dand-oss:
for mac issue

not a mach-o file error in Python - mismatching architecture
https://developer.apple.com/forums/thread/717724

For a Python project, I'm reading a .so file with the ctypes.CDLL and ctypes.cdll.LoadLibrary(...),

ans:

Apple platforms draw an important distinction between architecture and platform. Just because something is the right architecture doesn’t necessarily mean it’ll work on the target platform. In this case there’s an obvious difference, ELF versus Mach-O, but within Mach-O there are subtle differences. I explain this in more detail in my [An Apple Library Primer](https://developer.apple.com/forums/thread/715385)  post.

dand-oss:
so mac python is elf (which one?), and github mac image is mac-o
"architecture" but both are x86

native-api:
The "not a mach-o file error" doesn't happen with PyPy bootstrapper.
So it looks like we're done here when the checks complete.

dand-oss:
that means the pyenv mac python binary is incompatible with this architecture

considering systemic problem

Did we install cpython wrong (platfrom x86 but elf on mac o image)
or is pyenv idea of mac builds broken (need both, as msg above)

native-api:
> Did we install cpython wrong (platfrom x86 but elf on mac o image)
> or is pyenv idea of mac builds broken (need both, as msg above)

```
% file /Users/admin/.pyenv/versions/2.7.18/bin/python
/Users/admin/.pyenv/versions/2.7.18/bin/python: Mach-O 64-bit executable x86_64
```

The error message said that it tried to `_dlopen` a `.tbd` which is not a real library AFAICS.
So it seems to be a flaw in the library locator.

Issue #2294, opened 2022-03-24
`--enable-shared` should be default


rickmark:
# Summary

Because `pyenv` is used by developers, and because adding shared requires removal and re-installation (a timely and potentially breaking process) shared library (which is only a small addition to the static lib) should be on by default.

Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [X] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [X] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [X] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [X] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [X] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

native-api:
We don't override the defaults in order to provide users with as close to vanilla experience as practical, in order to minimize the possible surprises. Do you have any evidence that shows that `--enable-shared` is the configuration that users expect by default? AFAICS from https://github.com/python/cpython/blob/12c0012cf97d21bc637056983ede0eaf4c0d9c33/configure.ac#L1224-L1237, CPython only builds with `--enable-shared` by default in Cygwin -- hardly the most common use case.

If you need custom configuration switches in every installation, you can put them into the `PYTHON_CONFIGURE_OPTS` envvar.

native-api:
Good, but that seems to be Homebrew's local decision -- and outside their main platform to boot.

Is there anything closer to the roots? E.g. previously, [a user reported that SQLite extensions are enabled by default in the official Python Docker image](https://github.com/pyenv/pyenv/issues/1702). That was good enough for me :slightly_smiling_face:

rickmark:
Maybe this should be kicked up to python itself?  It seems silly that if you install their package from the internet you get the Dylib / framework but when building from source you do not?

[...]

PR #2554, opened 2022-12-08
--enable-shared by default on non-mac systems

anton-petrov:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2294

### Description

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -806,6 +806,9 @@ build_package_standard_build() {
       use_homebrew_zlib || true
     fi
     use_dsymutil || true
+    if ! is_mac; then
+      build_package_enable_shared || true
+    fi
   fi
 
   ( if [ "${CFLAGS+defined}" ] || [ "${!PACKAGE_CFLAGS+defined}" ]; then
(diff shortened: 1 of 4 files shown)


plugins/python-build/bin/python-build
```diff
@@ -806,6 +806,9 @@ build_package_standard_build() {
       use_homebrew_zlib || true
     fi
     use_dsymutil || true
+    if ! is_mac; then
+      build_package_enable_shared || true
+    fi
```
native-api, line 811: What's the rationale for "non-Mac"?

I only found that it's incompatible with `--enable-framwework` so we can probably only not apply it if that has been passed.

native-api: https://github.com/python/cpython/blob/cd67c1bb30eccd0c6fd1386405df225aed4c91a9/configure.ac#L3049

anton-petrov:
Maybe we should experiment with flags and look at how CPython is built in the homebrew:
https://github.com/Homebrew/homebrew-core/blob/51fdfef54d478e1fbfa66724d621c06a44be7a3c/Formula/python%403.11.rb#L139

PR #2555, opened 2022-12-10
fix actions to test pypy source build

dand-oss:
request to test pypy src build


dand-oss:
included in another PR

PR #2561, opened 2022-12-14
Fix non-bash output while detecting shell

ianchen-tw:

### Prerequisite

### Description
`pyenv-init --detect-shell` will output non-bash script result while configuration files are not supported, since placeholder values are used for explanation.

### Tests
Test for the new function.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -22,7 +22,6 @@ fi
 mode="help"
 no_rehash=""
 no_push_path=""
-no_config_place_holder=""
 for args in "$@"
 do
   if [ "$args" = "-" ]; then
@@ -37,7 +36,6 @@ do
 
   if [ "$args" = "--detect-shell" ]; then
     mode="detect-shell"
-    no_config_place_holder=1
     shift
   fi
 
@@ -45,7 +43,7 @@ do
     no_push_path=1
     shift
   fi
-  
+
   if [ "$args" = "--no-rehash" ]; then
     no_rehash=1
     shift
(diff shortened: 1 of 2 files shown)


native-api (commented):
> pyenv-init --detect-shell will output non-bash script result while configuration files are not supported, since placeholder values are used for explanation.

Maybe my English is not good enough, but I've no idea what this means or how it explains the purpose and necessity of the added function.

Perhaps if you could update the Pyenv-Installer PR to see how it is supposed to be used?

native-api:
> Perhaps if you could update the Pyenv-Installer PR to show how it is supposed to be used?

Okay, you already did. Let's see then...

libexec/pyenv-init
```diff
@@ -112,8 +114,15 @@ function detect_profile() {
     rc='~/.profile'
     ;;
   * )
-    profile='your shell'\''s login startup file'
-    rc='your shell'\''s interactive startup file'
+    # if no_config_place_holder is set,
+    # do not filled in any placehodler for config file
+    if [ -n "$no_config_place_holder" ]; then
```
native-api, line 119: This should not be a global variable.

Either it should be an optional argument to the function, or an exit code should indicate whether shell detection was successful.

native-api (changes requested):

libexec/pyenv-init
ianchen-tw: b502c9a9dffb6be593e189459e3cbacb8b15d960

PR #2575, opened 2022-12-29
Add `pyenv rehash` to the steps needed to run a newly-installed version.

j8spencer:
The documentation as written will not run a newly-installed version since the shims need to be rehashed first. This change makes that clear.

### Prerequisite

### Description

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,7 @@ E.g. to install and then switch to the latest 3.10 release:
 
 ```sh
 pyenv install 3.10
+pyenv rehash
 pyenv global 3.10
 ```
 


native-api:
A manual rehash is not needed because [`pyenv install` does it automatically after the installation succeeds](https://github.com/pyenv/pyenv/blob/f897c502022cf3132574afc084fddf9252644d72/plugins/python-build/bin/pyenv-install#L281-L283).

j8spencer:
It was required for me on a fresh install. After `pyenv install 3.11.1` and `pyenv global 3.11.1`, python/python3/pip/pip3 still invoked the system-installed versions. After 'pyenv rehash`, I got the new 3.11.1 for all those commands.

native-api:
That's not normal. If report it and provide the necessary diagnostic info, we can look into it.

j8spencer:
What diagnostic info do you need? Is there a script I can run to capture
what you'd like?

native-api:
Everything is listed in the issue template.

Issue #2569, opened 2022-12-23
2.3.9 Regression in pyenv-latest


willgleich:
pyenv 2.3.9

`pyenv install 3.6:latest`
> python-build: definition not found: 3.6:latest
   See all available versions with `pyenv install --list'.

After installing pyenv with `curl https://pyenv.run | bash` the latest release will not install python versions using the latest tag. This is showing up on an aarach64 ubuntu 20.04 image.

Workaround: Downgrade to pyenv v2.3.8 and everything works as expected 
```
export PYENV_GIT_TAG=v2.3.8
curl https://pyenv.run | bash
```

### Prerequisite
* [X] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [X] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [X] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [X] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [X] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [X] Platform information (e.g. Ubuntu Linux 16.04): Ubuntu 20.04
- [X] OS architecture (e.g. amd64): aarach64
- [X] pyenv version: 2.3.9
- [X] Python version: 3.8.10
- [X] C Compiler information (e.g. gcc 7.3): gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1)
- [X] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`. E.g. if you have a problem with installing Python, run `env PYENV_DEBUG=1 pyenv install -v <version> 2>&1 | tee trace.log` (note the `-v` option to `pyenv install`).
  * https://gist.github.com/willgleich/3235f0a282b139761b8dbf2dc2c9af8c

[...]

PR #2568, opened 2022-12-22
Allow multiple versions for pyenv-install

rockandska:
Make sure you have checked all steps below.

### Prerequisite
* [X] Please consider implementing the feature as a hook script or plugin as a first step.
  * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins](https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a plugin if possible.
* [X] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv will be imported to pyenv too, eventually.
  * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with rbenv.
* [X] My PR addresses the following pyenv issue (if any)
  - Closes https://github.com/pyenv/pyenv/issues/919

### Description
As stated in #919, many of the users (as me) expect to have pyenv to install all versions specified in `.python-version` and have `pyenv install` works the same way as `pyenv local` works

### Tests
- [X] My PR adds the following unit tests (if any)
  - `install proper multi versions`
  - `install resolves a prefix with multi versions`
  - `install pyenv local multi versions by default`
  - `multi arguments for pyenv-install`
- [X] My PR remove the following unit tests (if any)
  - `too many arguments for pyenv-install`
    - replaced by  `multi arguments for pyenv-install` since user could specify multi versions now

PR #2599, opened 2023-01-22
Fix `:latest` after #2568

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2569

### Description

#2568 inadvertently broke `:latest` due to changed logic.
Tests passed because there was no test for `:latest`.

### Tests

Test for `:latest`.
Note that other tests disable hooks. We enable only Pyenv built-in ones for this particular test.


rockandska:
sorry to forgot about this @native-api

native-api:
That's okay @rockandska . As I said, the reason this slipped through is because we had no unit test for `:latest`.

PR #2612, opened 2023-02-06
Fix link resolving in pyenv-versions

laggardkernel:
Resolves a bug introduced in #2609, `resolve_link` is not defined when `pyenv-realpath.dylib` is built. Here, the public API should be `realpath` but not `resolve_link`.

### DIFF
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -112,7 +112,7 @@ print_version() {
   if [[ -L "$path" ]]; then
     # Only resolve the link itself for printing, do not resolve further.
     # Doing otherwise would misinform the user of what the link contains.
-    version_repr="$version --> $(resolve_link "$path")"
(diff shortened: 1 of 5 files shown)


libexec/pyenv-versions
```diff
@@ -112,7 +112,7 @@ print_version() {
   if [[ -L "$path" ]]; then
     # Only resolve the link itself for printing, do not resolve further.
     # Doing otherwise would misinform the user of what the link contains.
-    version_repr="$version --> $(resolve_link "$path")"
```
native-api, line 115: Using `resolve_link` here is intentional. As the comment says, `pyenv versions` should only display the link's contents, not the final path.

native-api: I guess we can use `readlink` here -- since it's supposed to be always present and we don't need `-f` here.

PR #2626, opened 2023-02-20
Add "Building for maximum performance" instructions for Apple silicon

GabrielBianconi:
This is a documentation-only PR, targeted at Apple silicon users trying to follow "Building for maximum performance".

The `-march=native` flag does not work under the default `clang` version at the time of writing (clang 14):

https://stackoverflow.com/questions/65966969/why-does-march-native-not-work-on-apple-m1

Additionally, `--enable-optimizations --with-lto` is only supported by recent Python minor versions on macOS:

https://github.com/python/cpython/issues/86401

In the issue above, they mention 3.10+, with the intent to backport. I haven't checked if it's been backported to specific prior micro versions.

Finally, the installation does not use a Homebrew-updated `clang` by default.

### DIFF
--- a/plugins/python-build/README.md
+++ b/plugins/python-build/README.md
@@ -175,6 +175,15 @@ common for performance improvement from this to be in the ballpark of 30%.
 env PYTHON_CONFIGURE_OPTS='--enable-optimizations --with-lto' PYTHON_CFLAGS='-march=native -mtune=native' pyenv install 3.6.0
 ```
 
+If you are using a Mac computer with Apple silicon (e.g. M1), you must install
+`clang` 15+ before running the command above, and choose a recent minor version
+of Python. If you use Homebrew to update `clang` (`brew install llvm`), you must
+make sure the correct version is used:
+
+```sh
+env PATH="/opt/homebrew/opt/llvm/bin:$PATH" PYTHON_CONFIGURE_OPTS='--enable-optimizations --with-lto' PYTHON_CFLAGS='-march=native -mtune=native' pyenv install 3.10
+```
+
 You can also customize the task used for profile guided optimization by setting
 the `PROFILE_TASK` environment variable, for instance, `PROFILE_TASK='-m
 test.regrtest --pgo -j0'` will run much faster than the default task.


plugins/python-build/README.md
```diff
@@ -175,6 +175,15 @@ common for performance improvement from this to be in the ballpark of 30%.
 env PYTHON_CONFIGURE_OPTS='--enable-optimizations --with-lto' PYTHON_CFLAGS='-march=native -mtune=native' pyenv install 3.6.0
 ```
 
+If you are using a Mac computer with Apple silicon (e.g. M1), you need to install
+`clang` 15+ before running the command above and choose a recent minor version
+of Python. If you use Homebrew to install `clang` (`brew install llvm`), override
+the compiler to use:
+
+```sh
+env PATH="/opt/homebrew/opt/llvm/bin:$PATH" PYTHON_CONFIGURE_OPTS='--enable-optimizations --with-lto' PYTHON_CFLAGS='-march=native -mtune=native' pyenv install 3.10
```
native-api, line 184: Did you test this command? [We have many reports of the build breaking when Homebrew's LLVM is on PATH](https://www.google.com/search?q=Homebrew+LLVM+on+PATH+site%253Agithub.com%252Fpyenv%252Fpyenv):

GabrielBianconi: I did, and seemed to have worked correctly. I installed by 3.10 (3.10.10) and 3.11 (3.11.2) using the exact command above. I'm using macOS 12.5.1 on an M1.

native-api: Besides, is the result compatible with PyPI-hosted wheels? E.g. GCC-built Python has a different architecture tag.
Apple has been using `arm` instead of the standard `arm64` the last time I checked.

native-api: If it didn't build them from source, the architectures must be compatible.

This is not an officially supported setup though so if it goes in, it must be with a warning that this has not been extensively tested and might break of have subtle incompatibilities, use at your own risk...

I don't think the README should be advertising unsupported setups. The wiki would be fine.

Plus, this is only temporarily relevant, until Apple ships CLang 15.

native-api, line 184: `-march`? What happens to automatically deducing architecture from `arch`? Will this work when compiling for either `arm64` or `amd64`?

GabrielBianconi: I didn't build from source. Please feel free to move this information to the wiki or add the necessary disclaimers. I spent quite some time solving it so I figured I'd share the solution for others.

GabrielBianconi: I didn't build from source. Please feel free to move this information to the wiki or add the necessary disclaimers. I spent quite some time solving it so I figured I'd share the solution for others. :)

GabrielBianconi: This is from the original command in the README (L175, just above my change). I copied the command as is and just added Homebrew's LLVM.

native-api: The original command is for Linux where there typically isn't an issue of two native architectures like in MacOS.

GabrielBianconi: I can try another command if you'd like?

native-api: Here: https://github.com/pyenv/pyenv/wiki/Build-for-maximum-performance-in-MacOS

The wiki has been set up to be user-editable so you can tune that text to your heart's content.

GabrielBianconi: Looks great, please feel free to close this PR.

PR #2629, opened 2023-02-22
Support for "BusyBox version" of "head"

schuellerf:
I wanted to try some things within [Venus OS](https://github.com/victronenergy/venus) without killing the python installation there (seemingly a use case for pyenv :-) )

I used the pyenv source package `v2.3.13.tar.gz` and this failed with
```
# head -1
head: invalid option -- '1'
BusyBox v1.31.1 (2023-02-09 11:42:03 UTC) multi-call binary.

Usage: head [OPTIONS] [FILE]...
```

So I was tempted to patch pyenv to use `head -n1` instead of `head -1`
In the systems I know, both are valid.

If there are versions of `head` out there which *don't* support `head -n1` I could also implement some check first and use the appropriate version then, which would be an overkill if it's not necessary.

Any thoughts on this?

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -29,7 +29,7 @@ if enable -f "${BASH_SOURCE%/*}"/../libexec/pyenv-realpath.dylib realpath 2>/dev
 else
   [ -z "$PYENV_NATIVE_EXT" ] || abort "failed to load \`realpath' builtin"
 
-  READLINK=$(type -P greadlink readlink | head -1)
+  READLINK=$(type -P greadlink readlink | head -n1)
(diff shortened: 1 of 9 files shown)


schuellerf:
... probably it's not necessary to patch this in _all places_. I could as well narrow this down.

native-api (commented):
I don't think there are versions of `head` that don't support `-n1` [as it's a part of the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/head.html). In fact, `-1` has evolved as a shorthand of `-n1`.

schuellerf:
... not sure if my change caused the check to fail
Also no clue what to do about it (other than re-run :-) )

Issue #2141, opened 2021-11-01
Command `pyenv install 3.6.8` fails on Arch ('error: failed to install pip via get-pip.py')


el-iot:
I cannot seem to install python `3.6.8` (I have not tried other versions) on Arch Linux. The last 10 or-so log lines of the command `pyenv install 3.6.8` are:
```
rm -f /home/el/.pyenv/versions/3.6.8/share/man/man1/python3.1
(cd /home/el/.pyenv/versions/3.6.8/share/man/man1; ln -s python3.6.1 python3.1)
if test "xupgrade" != "xno"  ; then \
	case upgrade in \
		upgrade) ensurepip="--upgrade" ;; \
		install|*) ensurepip="" ;; \
	esac; \
	 ./python -E -m ensurepip \
		$ensurepip --root=/ ; \
fi
Looking in links: /tmp/tmp470fs0f0
Collecting setuptools
Collecting pip
Installing collected packages: setuptools, pip
Successfully installed pip-18.1 setuptools-40.6.2
```
Paradoxically, the logs indicate that `pip` could not be installed with `get-pip.py`, yet the final log lines indicate that pip was successfully installed.
Platform / OS:
```
$ uname -rm
5.14.14-arch1-1 x86_64
```
Pyenv:
```
$ pyenv --version
pyenv 2.2.0
```
Python:
```
$ python --version
Python 3.9.7
```
C Compiler information (e.g. gcc 7.3): 
```
$ gcc --version
gcc (GCC) 11.1.0
<copyright statement>
```
Please reproduce the problem with debug tracing enabled and attach the resulting output as a gist
[python-build.20211101142508.2192272.log](https://github.com/pyenv/pyenv/files/7451162/python-build.20211101142508.2192272.log)

Let me know if I have missed anything! I don't often make issues on Github and any feedback would be good :+1:

native-api:
Please attach the console output, not the log file.

vakabus:
I think I might have the same problem. The console output is this:

```
vasek@dingo ~ [1]> pyenv install 3.6.12
Downloading Python-3.6.12.tar.xz...
-> https://www.python.org/ftp/python/3.6.12/Python-3.6.12.tar.xz
Installing Python-3.6.12...


BUILD FAILED (Arch rolling using python-build 20180424)

Inspect or clean up the working tree at /tmp/python-build.20211103110330.314292
Results logged to /tmp/python-build.20211103110330.314292.log

Last 10 log lines:
[...]

PR #2630, opened 2023-02-24
bpo-27987 for v3.5.10 and v3.6.15: align by 16bytes on 64bit platforms

chaimleib:
### Description
This resolves an error encountered on x86_64 for python versions 3.5.10 and 3.6.15 where `./python -c 'import ctypes'` causes an immediate segfault. Also, when pyenv install runs `./python -E -m ensurepip --upgrade --root=/`, this imports ctypes and also causes a segfault. 

The segfault causes `pyenv install 3.5.10` and the same for 3.6.15 to fail and not install.

This bug was resolved in [bpo-27987](https://bugs.python.org/issue27987) for the v2.7 branch and v3.7 branch and later. (Note that in the conversation, although one of the patches was reverted, the revert was reverted at the end.) This PR backports those patches for 3.5.10 and 3.6.15, which were left unfixed. These patches would likely not be accepted upstream because 3.5 and 3.6 are EOL.

Despite these versions being EOL, I was interested for old times sake. Also, the test harness for chaimleib/intervaltree lists these versions, among others, as test targets. 

### Tests
```
set -e
for v in 3.5.10 3.6.15; do
  pyenv install $v
  pyenv global $v
  python --version
  python -c 'import ctypes'
done
```

### DIFF
--- a/plugins/python-build/share/python-build/patches/3.5.10/Python-3.5.10/0001-bpo-27987-pymalloc-align-by-16bytes-on-64bit-platfor.patch
+++ b/plugins/python-build/share/python-build/patches/3.5.10/Python-3.5.10/0001-bpo-27987-pymalloc-align-by-16bytes-on-64bit-platfor.patch
@@ -1,0 +1,43 @@
+From 112cceb514e79b4c9805f9e5ea037a205be428da Mon Sep 17 00:00:00 2001
+From: Inada Naoki <songofacandy@gmail.com>
+Date: Tue, 14 May 2019 18:51:15 +0900
+Subject: [PATCH 1/2] bpo-27987: pymalloc: align by 16bytes on 64bit platform
+ (GH-12850)
+
+---
+ .../2019-04-16-11-52-21.bpo-27987.n2_DcQ.rst                | 3 +++
+ Objects/obmalloc.c                                          | 6 ++++++
+ 2 files changed, 9 insertions(+)
+ create mode 100644 Misc/NEWS.d/next/Core and Builtins/2019-04-16-11-52-21.bpo-27987.n2_DcQ.rst
+
+diff --git a/Misc/NEWS.d/next/Core and Builtins/2019-04-16-11-52-21.bpo-27987.n2_DcQ.rst b/Misc/NEWS.d/next/Core and Builtins/2019-04-16-11-52-21.bpo-27987.n2_DcQ.rst
+new file mode 100644
+index 0000000000..b0f32a5c6c
+--- /dev/null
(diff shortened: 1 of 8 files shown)


native-api:
This looks like a solution for a old issue -- https://github.com/pyenv/pyenv/issues/2141 !

native-api:
The errors are because 3.5.10 is incompatible with MacOS 11+ (3.6.15 has downstream patches that make it compatible). Going to fix that? [We're fine either way sine this is not an additional breakage](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md#adding-version-specific-fixespatches).

plugins/python-build/share/python-build/3.5.10
```diff
@@ -1,6 +1,6 @@
 prefer_openssl11
-install_package "openssl-1.1.0j" "https://www.openssl.org/source/old/1.1.0/openssl-1.1.0j.tar.gz#31bec6c203ce1a8e93d5994f4ed304c63ccf07676118b6634edded12ad1b3246" mac_openssl --if has_broken_mac_openssl
-install_package "readline-8.0" "https://ftpmirror.gnu.org/readline/readline-8.0.tar.gz#e339f51971478d369f8a053a330a190781acb9864cf4c541060f12078948e461" mac_readline --if has_broken_mac_readline
+install_package "openssl-1.1.1k" "https://www.openssl.org/source/openssl-1.1.1k.tar.gz#892a0875b9872acd04a9fde79b1f943075d5ea162415de3047c327df33fbaee5" mac_openssl --if has_broken_mac_openssl
+install_package "readline-8.1" "https://ftpmirror.gnu.org/readline/readline-8.1.tar.gz#f8ceb4ee131e3232226a17f51b164afc46cd0b9e6cef344be87c65962cb82b02" mac_readline --if has_broken_mac_readline
```
native-api, line 3: This has no effect: these are only used in MacOS without the corresponding Homebrew packages.

Look at other patches for 3.6.15 and their commit history. We've had an issue with the specific error message that you're getting.

native-api: Moreover, this changes version lines, so unless you're sure 3.5 supports these new version lines, better leave them alone.

chaimleib: Aha. I read https://github.com/pyenv/pyenv/issues/1737#issuecomment-781647584 and I noticed that for other versions of Python, updating dependencies seemed to work, so I thought I'd give this a shot.

But thinking more about this, somehow v3.5.10 used to build correctly and find the `sendfile` function before. Updating dependencies shouldn't change that.

So, I reverted the change.

native-api:
This doesn't look normal... Did it just... work? :laughing: 

Well done! And thank you!

PR #2652, opened 2023-03-22
Update README from pyenv install -l to -list

brandanblevins:
A simple update to the READEME with the correct command.


native-api:
`-l` is the correct command:

```
$ pyenv install --help
Usage: pyenv install [-f] [-kvp] <version>...
<...>
  -l/--list          List all available versions
<...>
```

PR #2655, opened 2023-03-25
Try locate `readlink` first in pyenv-hooks, fix #2654

### LINKED ISSUE #2654: Slow execution on WSL due to excessive PATH traversal for `greadlink`
Too many issues will kill our team's development velocity, drastically.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Debian 11 on WSL2
- [x] OS architecture (e.g. amd64): amd64
- [x] pyenv version: 2.3.16
- [x] Python version: not relavant
- [x] C Compiler information (e.g. gcc 7.3): not relevant

In `pyenv-hooks`, `greadlink` or `readlink` will be searched with `type -P`.

https://github.com/pyenv/pyenv/blob/9fad1f46c50526df1fa3a5fd8188e785bb530b9c/libexec/pyenv-hooks#L29-L33

On common Linux systems, `greadlink` does not exist. So bash will traverse the whole `PATH` for it, then fails and fall back to `readlink` and succeeds. This process will repeat multiple times shown by `PYENV_DEBUG=1`.

This behavior leads to multiple PATH traversals on Linux, and causes heavy IO on WSL because it includes Windows PATH (which is really slow to access) by default. This adds about 1-2 seconds latency to every `pyenv exec` on my machine.

~~Swapping `greadlink` and `readlink` might be the easiest workaround for this problem.~~ The easiest workaround will be first search for `readlink` then `greadlink`.

Harry-Chen:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2654

### Description

See #2654 for details.

### Tests

### DIFF
--- a/libexec/pyenv-hooks
+++ b/libexec/pyenv-hooks
@@ -26,7 +26,7 @@ if ! enable -f "${BASH_SOURCE%/*}"/pyenv-realpath.dylib realpath 2>/dev/null; th
     echo "pyenv: failed to load \`realpath' builtin" >&2
     exit 1
   fi
-READLINK=$(type -P greadlink readlink | head -n1)
+READLINK=$(type -P readlink)
 if [ -z "$READLINK" ]; then
   echo "pyenv: cannot find readlink - are you missing GNU coreutils?" >&2
   exit 1
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -29,7 +29,7 @@ if enable -f "${BASH_SOURCE%/*}"/../libexec/pyenv-realpath.dylib realpath 2>/dev
 else
   [ -z "$PYENV_NATIVE_EXT" ] || abort "failed to load \`realpath' builtin"
 
-  READLINK=$(type -P greadlink readlink | head -n1)
+  READLINK=$(type -P readlink)
   [ -n "$READLINK" ] || abort "cannot find readlink - are you missing GNU coreutils?"
 
   resolve_link() {
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -34,7 +34,7 @@ if ! enable -f "${BASH_SOURCE%/*}"/pyenv-realpath.dylib realpath 2>/dev/null; th
     exit 1
   fi
 
-  READLINK=$(type -P greadlink readlink | head -n1)
+  READLINK=$(type -P readlink)
   if [ -z "$READLINK" ]; then
     echo "pyenv: cannot find readlink - are you missing GNU coreutils?" >&2
     exit 1
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -56,9 +56,14 @@ lib() {
 }
 lib "$1"
 
+READLINK=$(type -P readlink)
+if [ -z "$READLINK" ]; then
+  echo "pyenv: cannot find readlink - are you missing GNU coreutils?" >&2
+  exit 1
+fi
 
 resolve_link() {
-  $(type -P greadlink readlink | head -n1) "$1"
+  $READLINK "$1"
 }
 
 abs_dirname() {
--- a/plugins/python-build/test/test_helper.bash
+++ b/plugins/python-build/test/test_helper.bash
@@ -151,7 +151,7 @@ path_without() {
       if [ "$found" != "${PYENV_ROOT}/shims" ]; then
         alt="${PYENV_TEST_DIR}/$(echo "${found#/}" | tr '/' '-')"
         mkdir -p "$alt"
-        for util in bash head cut readlink greadlink; do
+        for util in bash head cut readlink; do
           if [ -x "${found}/$util" ]; then
             ln -s "${found}/$util" "${alt}/$util"
           fi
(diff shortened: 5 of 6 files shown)


libexec/pyenv-hooks
```diff
@@ -26,7 +26,10 @@ if ! enable -f "${BASH_SOURCE%/*}"/pyenv-realpath.dylib realpath 2>/dev/null; th
     echo "pyenv: failed to load \`realpath' builtin" >&2
     exit 1
   fi
-READLINK=$(type -P greadlink readlink | head -n1)
```
native-api, line 29: This specific order is here for a reason: in MacOS, we do prefer GNU readlink from Homebrew Coreutils to the stock BSD readlink.

But maybe this reason no longer applies?

native-api: ...and if we don't really need the extra features of GNU readlink, we can probably drop the choosing code altogether.

Harry-Chen: > This specific order is here for a reason: in MacOS, we do prefer GNU readlink from Homebrew Coreutils to the stock BSD readlink.

If so, maybe detecting the system first would be a better choice?

> if we don't really need the extra features of GNU readlink, we can probably drop the choosing code altogether.

`$READLINK` is called without any options in `pyenv` (or so-called readlink mode). I do not know the exact difference between GNU and BSD versions of readlink, but seems in readlink mode they do basically the same thing. Maybe using either version is fine.

I searched the code and found multiple occurrences of such `readlink` detection. Then I realized the reason why Python builds incredibly slowly in WSL -- still due to failed `greadlink` search in the whole PATH before executing every command. So I believe this is a serious problem for WSL users and should be solved in some way.

native-api: To my best knowledge, the only feature that we would need from GNU readlink is `-f`. But, since the presence of it was never a given, we have workarounds. `-f` is (sloppily) relied upon in one place in Pyenv-Virtualenv, but nowhere in Pyenv proper.

I suggest removing the chooser and see if any of the tests fail.

native-api: We do require GNU readlink to run Bats (our test runner), there's a comment in the corresponding place. So make sure to leave the selector there. That code only runs when running tests, it doesn't affect normal use of Pyenv.

Harry-Chen: I searched for `resolve_link` and `readlink`, seems no `-f` is used:

```
plugins/python-build/test/pyenv_ext.bats
91:resolve_link() {
243:  [[ "$(resolve_link "${INSTALL_ROOT}/bin/python")" == "python3.4" ]]
244:  [[ "$(resolve_link "${INSTALL_ROOT}/bin/python-config")" == "python3.4-config" ]]

plugins/python-build/bin/python-build
65:resolve_link() {
80:    path="$(resolve_link "$name" || true)"

libexec/pyenv
35:  resolve_link() {
50:      path="$(resolve_link "$name" || true)"

libexec/pyenv-hooks
35:resolve_link() {
47:    path="$(resolve_link "$name" || true)"

libexec/pyenv-versions
43:  resolve_link() {
56:      path="$(resolve_link "$name" || true)"
```

I will try to remove all `greadlink` names first, and wait for the CI result to fix any problem.

Harry-Chen: Sorry to bother, but please approve the workflows @native-api

native-api: Btw you can experiment in your fork. Just enable Github Actions in it.

Harry-Chen:
Only one test failed:

```text
not ok 183 finds versions where present
# (from function `assert_equal' in file test/test_helper.bash, line 67,
#  from function `assert_output' in file test/test_helper.bash, line 76,
#  in test file test/whence.bats, line 20)
#   `assert_output <<OUT' failed
# expected: 2.7
# 3.4
# actual:   /home/runner/work/pyenv/pyenv/test/../libexec/pyenv-latest: line 39: printf: write error: Broken pipe
# 2.7
# 3.4
```

Seems this is not a functional error, but I have no permission to re-run this workflow in `pyenv` repo. So I ran it again in my fork [with no failure](https://github.com/Harry-Chen/pyenv/actions/runs/4524590349/jobs/7968527704). Maybe it is safe to conclude that all `greadlink` can be safely removed?

anton-petrov:
Tested on macOS without `greadlink`: all work fine as for me, so, PR ready for merge.

PR #2670, opened 2023-04-05
Add CPython 3.10.11

mirekdlugosz:

### Prerequisite
NA

### Description
- Add CPython 3.10.11 released earlier today: https://www.python.org/downloads/release/python-31011/

### Tests
NA


hoanguyen-sportsbet:
this below function does not make any sense - as the result - it'll download the same package and it's causing an issue if we don't have xz util installed 
**if has_tar_xz_support**; then
    install_package "Python-3.10.11" "**https://www.python.org/ftp/python/3.10.11/Python-3.10.11.tar.xz**#3c3bc3048303721c904a03eb8326b631e921f11cc3be2988456a42f115daf04c" standard verify_py310 copy_python_gdb ensurepip
**else**
    install_package "Python-3.10.11" "**https://www.python.org/ftp/python/3.10.11/Python-3.10.11.tar.xz**#3c3bc3048303721c904a03eb8326b631e921f11cc3be2988456a42f115daf04c" standard verify_py310 copy_python_gdb ensurepip
fi

mirekdlugosz:
@hoanguyen-sportsbet you are right. Opened https://github.com/pyenv/pyenv/pull/2677

PR #2676, opened 2023-04-10
Nicer git upgrade command

noamcohen97:
### Description
Use `git -C` instead of changing the user's working directory and switching it back at the end

### DIFF
--- a/plugins/python-build/bin/pyenv-install
+++ b/plugins/python-build/bin/pyenv-install
@@ -271,7 +271,7 @@ for DEFINITION in "${DEFINITIONS[@]}"; do
         echo "  brew update && brew upgrade pyenv"
       elif [ -d "${here}/.git" ]; then
         printf ":\n\n"
-        echo "  cd ${here} && git pull && cd -"
+        echo "  git -C \"$(realpath "${here}")\" pull"
       else
         printf ".\n"
       fi


plugins/python-build/bin/pyenv-install
```diff
@@ -271,7 +271,7 @@ for DEFINITION in "${DEFINITIONS[@]}"; do
         echo "  brew update && brew upgrade pyenv"
       elif [ -d "${here}/.git" ]; then
         printf ":\n\n"
-        echo "  cd ${here} && git pull && cd -"
+        echo "  git -C \"$(realpath "${here}")\" pull"
```
native-api, line 274: Why the need for `realpath`?

native-api: We don't even load our `realpath` fallback here (it's not preinstalled in e.g. OSX).

native-api (commented):
I have doubts about this. This is not a script, it's an advice for the user.

So it needs to be as basic as possible.

Using any fancy switches and stuff can make an inexperienced user think they are required (there were precendents)!

plugins/python-build/bin/pyenv-install
noamcohen97: `realpath` is just to make the path look nicer, without `..` and stuff.
I can get rid of it :)

noamcohen97:
> So it needs to be as basic as possible.

Currently, the advice uses three commands to run, whereas I believe that one command would suffice and will be simpler. The `-C` switch is a common switch for commands running in the current working directory. 

Additionally, I noticed that whenever git pull fails, the user's environment gets stuck in the `pyenv` directory. 
My modification ensures that the command leaves the user's environment intact in the event of a git pull failure.

Finally, I have added quotes around the path in case it contains a space, which would prevent any errors that might occur as a result of unescaped characters.

native-api:
> Additionally, I noticed that whenever git pull fails, the user's environment gets stuck in the `pyenv` directory.
> My modification ensures that the command leaves the user's environment intact in the event of a git pull failure.

That's a fair point -- need to change the 2nd `&&` into `;`

native-api:
> > So it needs to be as basic as possible.
> 
> Currently, the advice uses three commands to run, whereas I believe that one command would suffice and will be simpler. The `-C` switch is a common switch for commands running in the current working directory.
> 
> Additionally, I noticed that whenever git pull fails, the user's environment gets stuck in the `pyenv` directory. My modification ensures that the command leaves the user's environment intact in the event of a git pull failure.
> 
> Finally, I have added quotes around the path in case it contains a space, which would prevent any errors that might occur as a result of unescaped characters.

@pyenv/pyenv-core-maintainers , any second opinions?

native-api:
> > Additionally, I noticed that whenever git pull fails, the user's environment gets stuck in the `pyenv` directory.
> > My modification ensures that the command leaves the user's environment intact in the event of a git pull failure.
> 
> That's a fair point -- need to change the 2nd `&&` into `;`

On a second thought, even this is okay. Since the commands are for manual use, it's better to break down loudly and clearly point the user that something went wrong and they cannot continue.

PR #2723, opened 2023-06-13
add removal of pyenv fish

Jeroendevr:
### Prerequisite
* [N/A ] My PR addresses the following pyenv issue (if any)

### Description

Some small doc improvement. 
Environment variables work a bit differently in Fish. Hereby explicitly how to remove the pyenv_dir in Fish. 

If this is accepted, I'll make contribution upstream to rbenv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,10 +557,18 @@ uninstall from the system.
   its root directory. This will **delete all Python versions** that were
   installed under the `` $(pyenv root)/versions/ `` directory:
 
-    ```sh
-    rm -rf $(pyenv root)
-    ```
+    - For **shell**
 
+      ```sh
+      rm -rf $(pyenv root)
+      ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -557,10 +557,18 @@ uninstall from the system.
...
+    - For **shell**
+
+      ```sh
+      rm -rf $(pyenv root)
+      ```
+
+    - For **Fish shell**:
 
+      ~~~ fish
+      rm -rf $PYENV_ROOT
```
native-api, line 569: It should probably be `(pyenv root)`

native-api:
According to https://fishshell.com/docs/current/fish_for_bash_users.html#command-substitutions, Fish actually supports the `$(command)` command substitution syntax as well. So there's no need for 2 separate sections.

PR #2739, opened 2023-06-28
Add Cinder 3.10 and Cinder configure patches

filips123:

### Prerequisite

### Description
- [x] Adds script to install Cinder 3.10, similarly to Cinder 3.8 (#2433). All distro and GCC warnings have been kept because Cinder still doesn't officially support building outside their Docker container
- [x] Also adds patches for Cinder configure scripts (both for 3.8 and 3.10) to disable `-Werror`. There are some warnings (`Wdeprecated-declarations`, `Wmaybe-uninitialized`, and others) that occur on newer (GCC 11+) versions and cause the build to fail. I'm not sure why those warnings/error occur on those newer versions, but from a quick search, it appears that at least some of them might be false-positive because of bugs in GCC. When `-Werror` is removed, the build works fine (it compiles and runs a simple Numpy test, not sure if everything is perfectly supported though). There is also existing issue about that (facebookincubator/cinder#114).

Supported distros:
- Ubuntu 22.04+
- Debian 11+
- Fedora 32+
- Arch

Basically, distros that have at least GCC 10 / support for C++20 should work. Alpine is not supported because it uses MUSL. Older distro versions are not supported because of of too old GCC version. I haven't tested macOS, but it probably doesn't work.

Test runs:
- Cinder 3.10: https://github.com/filips123/pyenv-cinder-testing/actions/runs/5400882249/jobs/9809934704
- Cinder 3.8: https://github.com/filips123/pyenv-cinder-testing/actions/runs/5400895786/jobs/9809967403

### Tests

### DIFF
--- a/plugins/python-build/share/python-build/cinder-3.10-dev
+++ b/plugins/python-build/share/python-build/cinder-3.10-dev
@@ -1,28 +1,40 @@
-require_distro Fedora 32 &>/dev/null || \
-{ echo
-  colorize 1 "WARNING"
-  cat >&2 <<!
-: The Cinder compiler only officially supports
-Facebook's Docker images which are Fedora 32 - based.
-It may fail to build on a system
-with a different GCC and/or Glibc version.
-!
-  echo
-}
+if [ "$(expr substr "$(uname -s)" 1 5)" != "Linux" ]; then
+  {
+    echo
+    colorize 1 "ERROR"
+    echo ": Cinder currently only supports Linux."
+    echo
+    return 1
+  } >&2
+fi
 
-[[ $(${CC:-gcc} -dumpversion 2>/dev/null) == 10 ]] || \
-{ command -v "gcc-10" >/dev/null && \
-  export CC="gcc-10" && \
-  echo "python-build: setting the compiler to \`gcc-10'"; } || \
-{
-  echo
-  colorize 1 WARNING
-  cat >&2 <<!
-: GCC 10 is not found on PATH.
-The build may fail.
-!
-  echo
-}
+if [[ $(${CC:-gcc} -dumpversion 2>/dev/null) != 10 ]] &&
+  (command -v "gcc-10" >/dev/null) &&
+  (command -v "g++-10" >/dev/null); then
+  echo "python-build: setting the compiler to GCC 10"
+  export CC="gcc-10"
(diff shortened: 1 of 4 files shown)


anton-petrov:
> Please be aware that Cinder is only built or tested on Linux x64; anything else (including macOS) probably won't work.
Remove macOS builds.

> c++: error: unrecognized command line option ‘-std=c++20’; did you mean ‘-std=c++2a’?
Ubuntu 20.04 build failed because of incorrect command line option.

Cinder is very specific Python implementation, so, I'm not sure if it's necessary to add it. 

But, if you want, pls, modify your PR for all cases, so that Cinder is installed only in Linux of a certain version and an error would be printed for other OS and platforms.

filips123:
> Ubuntu 20.04 build failed because of incorrect command line option.

It's probably because Ubuntu 20.04 uses GCC 9.4.0, which does not fully support C++20 that is needed by Cinder. That flag comes from Cinder build system.

> Cinder is very specific Python implementation, so, I'm not sure if it's necessary to add it. 

Well, yeah, it is a pretty specific implementation, but my previous PR for Cinder 3.8 was accepted, and pyenv already includes some "less standard" implementations, so I think it can still be useful to support it.

> But, if you want, pls, modify your PR for all cases, so that Cinder is installed only in Linux of a certain version and an error would be printed for other OS and platforms.

In [the previous PR](pyenv/pyenv#2433), it was suggested to just display [a warning](https://github.com/filips123/pyenv/blob/add-cinder-3.10/plugins/python-build/share/python-build/cinder-3.10-dev#L1-L25) on distros other than Fedora 32 with GCC other than 10, as that's what Cinder officially supports. This can probably be useful in case Cinder adds support for macOS in the future, so the check won't block the build there, and because then I didn't know which exact requirements Cinder has for building on Linux.

However, if needed, I can update those checks to also stop the build if not on Linux and using GCC less than 10.

anton-petrov:
> However, if needed, I can update those checks to also stop the build if not on Linux and using GCC less than 10.
Ok

filips123:
I added errors when not on Linux and not using at least GCC 10. I also added a few improvements to also support building on older distros (such as Ubuntu 20.04) as long as GCC 10 is installed (`gcc-10` and `g++-10`).

PR #2744, opened 2023-07-23
README update: Added UNIX reference near Automatic installer

VictorieeMan:
It's nice if the UNIX easy installation is easy to find by ctrl+f search, especially in such a large README file

Other ways of marking it would work as well; but having something near that says UNIX, macOS or Linux is nice I think.

Great project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,9 @@ As far as Pyenv is concerned, version names are simply directories under
 ## Installation
 
 ### Getting Pyenv
-#### Homebrew in macOS
+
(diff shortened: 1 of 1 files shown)


native-api:
The formulation as suggested is incorrect.

Both installer and Git checkout are for UNIX/MacOS; Homebrew is for MacOS and now for Linux, too.

Maybe a subsection "UNIX/MacOS", and then separately "Windows"?

VictorieeMan:
Good correction! I like that subsection idea.

PR #2749, opened 2023-07-28
Add PowerShell support

fblanchetNaN:

### Prerequisite
  - N/A

### Description

I added support for PowerShell (on Linux), it could probably be implemented for rbenv also, however I never used ruby and don't feel confident to implement it.

I know it's maybe a patch too important for the core code, if so just close this (draft) PR.

### Tests
- [ ] My PR adds the following unit tests (if any): WIP

I don't know how BATS works, I need help / hints to understand what should I do -- where is the tested shell specified?

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,16 @@ See [Advanced configuration](#advanced-configuration) for details and more confi
     pyenv init - | source
     ~~~
 
+  - For **Microsoft PowerShell**:
+
+    Add the commands to `$profile.CurrentUserAllHosts` by running the following in your terminal:
+
+    ~~~ pwsh
+    echo '$Env:PYENV_ROOT="$HOME/.pyenv"' >> $profile.CurrentUserAllHosts
+    echo '$Env:PATH="$Env:PYENV_ROOT/bin:$Env:PATH"' >> $profile.CurrentUserAllHosts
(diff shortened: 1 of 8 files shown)


native-api:
> I don't know how BATS works, I need help / hints to understand what should I do -- where is the tested shell specified?

Tests can be run with `make test`. It shows the commands it runs so you can run them by hand or run `bats` with other arguments when diagnosing tests (you'll probably need `-f`) etc.

Look at the existing tests for the files you've changed to get an idea of what they do. They're basically Bash scripts with some extra syntax that denotes tests and setup/teardown code that's preprocessed and then executed by Bats.

We're using a rather old Bats version, 1.2.0, the docs are at https://github.com/bats-core/bats-core/blob/v1.2.0/README.md . Apart from functions defined by Bats, we use additional ones defined in `test_helper.bash` and test files themselves.

Most tests are unit tests that mock any external interactions; IIRC in the affected files, there're only 1-2 integration tests for Fish.

native-api:
> I know it's maybe a patch too important for the core code, if so just close this (draft) PR.

@pyenv/pyenv-core-maintainers IMO supporting PowerShell is fine. However, due to unfamiliarity with it, we may need to request help if more stuff interacting with it will be needed in the future or problems are reported.

native-api (changes requested):
The test failures look like bona fide bugs in the code

PR #2754, opened 2023-08-13
Add micropython 1.20.0

scop:
The unix port no longer uses axtls. Earlier versions do, but only 1.9.x
requires it to be separately built; newer ones have the dependency in
order.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -925,6 +925,7 @@ build_package_ironpython_builder() {
 }
 
 build_package_micropython() {
+  local package_name="$1"
   if [ "${MAKEOPTS+defined}" ]; then
     MAKE_OPTS="$MAKEOPTS"
   elif [ -z "${MAKE_OPTS+defined}" ]; then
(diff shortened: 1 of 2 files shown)


plugins/python-build/bin/python-build
```diff
@@ -933,7 +933,7 @@ build_package_micropython() {
   { cd mpy-cross
     "$MAKE" $MAKE_OPTS
     cd ../ports/unix
-    "$MAKE" $MAKE_OPTS axtls
+    grep -qwF axtls Makefile && "$MAKE" $MAKE_OPTS axtls
```
scop, line 936: This is ugly, I don't know if there's a standard way of handling similar things in pyenv. On the other hand I wonder if there was a way to avoid doing this explicitly in the first place.

This might also break earlier version builds; tested with 1.15 that it doesn't.

native-api: If the build procedure has changed, the standard way is to make another build_* function (in this case, build_package_micropython) and call the correct function for corresponding versions. We already did that for another flavour.

scop: Turns out that the explicit `axtls` make seems to have been unnecessary since 1.10. Duplicating the function seems a bit overkill, as we can use the package name for good enough results IMO. Modified the PR as such.

native-api:
Superceded by https://github.com/pyenv/pyenv/pull/2869

Issue #2417, opened 2022-07-25
`pyenv` does not install `pip` for versions `3.10.*`


EmilLaursen:
I have an problem with symptions similar to #2348, but the proposed solution does fix the issue.

System python is `3.10.5`. Have used pyenv for a while with `3.9.*` versions, no problem. When I use pyenv to install any `3.10.*` version, it appears it does not install an accompanying `pip`. As a consequence, pip points to the system `pip`, and the system `site-packages`.
```console
pyenv global 3.10.4

❯ which pip
/home/eso/.pyenv/shims/pip

❯ pyenv which pip
/home/eso/.local/bin/pip

❯ pip --version
pip 22.1.2 from /home/eso/.local/lib/python3.10/site-packages/pip (python 3.10)
```
Looking in the `bin` folder show no `pip`:
```console
ls  ~/.pyenv/versions/3.10.4/bin
2to3       idle   idle3.10  pydoc3     python   python3.10         python3.10-gdb.py  python-config
2to3-3.10  idle3  pydoc     pydoc3.10  python3  python3.10-config  python3-config
```
However, for any `3.9.*` install, pip is present in `bin`:
```console
ls  ~/.pyenv/versions/3.9.13/bin
2to3         idle     normalizer  pip3.9       pydoc     python            python3.9-gdb.py
2to3-3.9     idle3    pip         pkginfo      pydoc3    python3           python3-config
doesitcache  idle3.9  pip3        poetry       pydoc3.9  python3.9         python-config
flake8       keyring  pip3.10     pycodestyle  pyflakes  python3.9-config  virtualenv
```
And the same sequence of commands shows the shim points to the correct `pip`:
```console
❯ pyenv global 3.9.13

❯ which pip
/home/eso/.pyenv/shims/pip

❯ pyenv which pip
/home/eso/.pyenv/versions/3.9.13/bin/pip

❯ pip --version
pip 22.1.2 from /home/eso/.pyenv/versions/3.9.13/lib/python3.9/site-packages/pip (python 3.9)
```
I have tried `3.10.3`-`3.10.5` - with the same result. And `3.9.10`-`3.9.13` work correctly. My `.zshrc` has the following:
```zsh
export PYENV_ROOT="$HOME/.pyenv"
command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init -)"
```

### Prerequisite
[...]

PR #2764, opened 2023-08-25
use -I with ensurepip

xaocon:
Helps to workaround #2417

The `-s` flag assures that nothing can be installed to user site-package but doesn't seem to keep `ensurepip` from looking there for pip. If pip is installed in the user site-package directory pip isn't installed for the newly built python. `-I` leaves user site-package out entirely during the pip installation.

### DIFF
--- a/plugins/.gitignore
+++ b/plugins/.gitignore
@@ -2,3 +2,4 @@
 !/.gitignore
 !/version-ext-compat
 !/python-build
+/python-build/test/build
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -2025,7 +2025,7 @@ build_package_ensurepip() {
     ensurepip_opts="--altinstall"
   fi
   # FIXME: `--altinstall` with `get-pip.py`
-  "$PYTHON_BIN" -s -m ensurepip ${ensurepip_opts} 1>/dev/null 2>&1 || build_package_get_pip "$@" || return 1
+  "$PYTHON_BIN" -I -m ensurepip ${ensurepip_opts} 1>/dev/null 2>&1 || build_package_get_pip "$@" || return 1
   build_package_symlink_version_suffix
 }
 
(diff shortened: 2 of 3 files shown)


xaocon:
May also fix https://github.com/asdf-community/asdf-python/issues/148 and https://github.com/asdf-community/asdf-python/issues/127.

native-api:
While this will install Pip in the https://github.com/pyenv/pyenv/issues/2417 case -- it won't be actually used, will it?
The Pip installation in user site-packages will prevail.

That's not a bad thing though -- if it works correctly.

Will this trigger distro-specific customizations of the system Pip for Pyenv-installed Pythons?

xaocon:
Yes, by default the user site packages are still going to be first in `sys.path`. Users could use the same workaround or others to make sure to use the version of pip installed by `python-build` if they care too. The main problems I wanted to solve was providing some additional consistency to the builds and assuring pip shims work correctly. This problem breaks the pip shims for both pyenv and asdf-python which has to be a relatively large population of python users. Even if having a version of pip installed in local won't be that common this affects a lot of people.

I'm not sure I understand the last question, could you point me at an example?

The change does cause `make test-build && make test` to pass on my system now.

native-api:
> I'm not sure I understand the last question, could you point me at an example?

Distros patch the system-provided Pip to better suit the distro's ecosystem.
E.g. for https://packages.ubuntu.com/jammy/python3-pip, the patches they apply can be found in http://archive.ubuntu.com/ubuntu/pool/universe/p/python-pip/python-pip_22.0.2+dfsg-1ubuntu0.3.debian.tar.xz , `patches` directory.

I was wondering: if one
* installs an upgrade for the system's Pip with `pip3 install -U pip` (which goes into `~/.local/lib/pythonX.Y/site-packages`)
* switches to a Pyenv-provided version with the same `X.Y`
* uses Pip from it
* => will the distro patches have an effect?

---

Anyway, I've just checked that no, they don't.
By checking in Ubuntu Jammy:

```python
>>> import pip._vendor.certifi
>>> pip._vendor.certifi.core.where()
```

With distro patches, it should print `'/etc/ssl/certs/ca-certificates.crt'`.
But it prints the unpatched value, `'<home dir>/.local/lib/python<X.Y>/site-packages/pip/_vendor/certifi/cacert.pem'`.

---

So, we're clear to go.

PR #2778, opened 2023-09-06
Add CPython 3.12.0rc2

edgarrmondragon:

### Prerequisite

### Description

### Tests


plugins/python-build/share/python-build/3.12.0rc2
```diff
@@ -0,0 +1,9 @@
+prefer_openssl11
+export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
+install_package "openssl-3.1.1s" "https://www.openssl.org/source/openssl-3.1.1s.tar.gz#b3aa61334233b852b63ddb048df181177c2c659eb9d4376008118f9c08d07674" mac_openssl --if has_broken_mac_openssl
```
zsol, line 3: FYI: this URL is a 404 nowadays. Maybe we want 3.1.2 instead?

edgarrmondragon: Seems to have been fixed in https://github.com/pyenv/pyenv/pull/2789 👍

Issue #2746, opened 2023-07-25
CPython 3.12-dev fails to build if openssl@1.1 is Homebrew-installed


brechtm:
Recording this here for anyone else running into this problem.

With both openssl@1.1 and openssl@3 installed with Homebrew, `pyenv install 3.12-dev` will choose openssl@1.1 and fail with 
```
[ERROR] _hashlib failed to import: dlopen(/private/var/folders/t3/n6h1s2ss58n5l5llnt6c4zbh0000gn/T/python-build.20230725102939.1328/Python-3.12-dev/build/lib.macosx-13.4-arm64-3.12/_hashlib.cpython-312-darwin.so, 0x0002): symbol not found in flat namespace '_EVP_MD_block_size'
[ERROR] _ssl failed to import: dlopen(/private/var/folders/t3/n6h1s2ss58n5l5llnt6c4zbh0000gn/T/python-build.20230725102939.1328/Python-3.12-dev/build/lib.macosx-13.4-arm64-3.12/_ssl.cpython-312-darwin.so, 0x0002): symbol not found in flat namespace '_SSL_get_peer_certificate'
```

Uninstalling openssl@1.1 temporarily will make the CPython build pick up openssl@3 instead and complete succesfully. I reinstalled openssl@1.1 afterwards since it is needed by older CPython versions.

This is on macOS Ventura 13.4.1 (c) on M1 with ARM64 Homebrew.

native-api:
Please add full diagnostic information as per the issue template.

native-api:
Also please specify the version and architecture of the OpenSSL package.

native-api:
WFM with openssl@1.1 1.1.1u x64.

bebleo:
I can report that I have the same issue with 3.12.0b4 after installing openssl with homebrew on macOS 13.5

```sh
openssl version
```

results in:

```sh
OpenSSL 3.1.1 30 May 2023 (Library: OpenSSL 3.1.1 30 May 2023)
```

And

```sh
which openssl
```

Gives:

```sh
/opt/homebrew/bin/openssl
```

[pyenv_install.3.12.0b4.verbose.error.txt](https://github.com/pyenv/pyenv/files/12209963/pyenv_install.3.12.0b4.verbose.error.txt)
[pyenv_install.3.12.0b4.error.txt](https://github.com/pyenv/pyenv/files/12209964/pyenv_install.3.12.0b4.error.txt)

native-api:
[...]

PR #2781, opened 2023-09-07
Prefer OpenSSL 3 in Homebrew since 3.12

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2746

### Description

The ARM64 build of OpenSSL in MacOS 13 Homebrew lacks some necessary functions for CPython 3.12.
While it's a workaround for third-party problem, CPython has been using OpenSSL 3 in CI for several releases now so the support can be considered official now.

### Tests
N/A


anton-petrov:
See #2805

cpython 3.12 build also fails on my arm64 mac.

Issue #2779, opened 2023-09-07
version has no --bare option, but versions has a --bare option


jjwatt:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x ] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x ] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x ] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x ] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [ x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
Not really a bug. Just a missing flag in a sub-command. `--bare` with `versions` is really useful in a test I'm building. Without it, `versions` is non-deterministic. e.g.,:
```
~/git/github.com/pyenv/pyenv master ⇣                                                                      
❯ pyenv shell "$(pyenv versions | grep 3.11 | tr -d '[[:space:]]')"
~/git/github.com/pyenv/pyenv master ⇣
❯ pyenv shell "$(pyenv versions | grep 3.11 | tr -d '[[:space:]]')"               
pyenv: version `* 3.11.5 (set by PYENV_VERSION environment variable)' not installed
```
That makes sense because before I select it:
```
~/git/github.com/pyenv/pyenv master ⇣                                                                      
❯ pyenv versions | grep 3.11                                                                               
  3.11.5 
```
and after
```
~/git/github.com/pyenv/pyenv master ⇣
❯ pyenv versions | grep 3.11
* 3.11.5 (set by PYENV_VERSION environment variable)
```
Different output because it's decorated with the explanation, which is great, and it's also great to have the `--bare` option for parsing:
```
~/git/github.com/pyenv/pyenv master ⇣
❯ pyenv versions --bare | grep 3.11.5
3.11.5
```
So, now I can use it like this
```
[...]

PR #2783, opened 2023-09-08
Add --bare option to `pyenv version`

jjwatt:
It just points at pyenv version-name.  This adds consistency with the behavior of `pyenv versions --bare`

### Prerequisite

I don't use `rbenv`, but from a quick look, I *could* make a similar change there. However, the files in question are already different (though the behavior is similar).

* [] My PR addresses the following pyenv issue (if any)
  - Closes https://github.com/pyenv/pyenv/issues/2779

### Description
Adds a `--bare` flag to `pyenv version`

### Tests

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -268,8 +268,14 @@ version of Python, or install a package that provides binaries.
 Displays the currently active Python version, along with information on
 how it was set.
 
+Usage: pyenv version [--bare]
+
+    --bare    show just the version name. An alias to `pyenv version-name'
+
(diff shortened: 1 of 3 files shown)


native-api:
Please add a test for the new option.

jjwatt:
Sure thing! `bats` for testing. Excellent. I think I see just where to add it.

anton-petrov:
@jjwatt Hi, so what about tests? I think, PR is ready for merge, except of tests.

PR #2786, opened 2023-09-14
Fix fish shell config docs

BooleanCat:

### Prerequisite

### Description

Copying the instructions for configuring the fish shell for pyenv verbatim fails because `fish_add_path` fails when the target directory doesn't exist.

### Tests

N/A - doc change only.


native-api (changes requested):
It's pointless to create `$PYENV_ROOT/bin` if it doesn't exist.
The command should then be conditional on the directory existing.

Same goes for the 2nd snippet just below.

PR #2788, opened 2023-09-15
Fix get-pip urls for older pypy versions

TimPansino:

### Prerequisite

### Description

On older recipes for `PyPy` the current `get-pip.py` is called, when it returns and incompatible script. Older `CPython` work around this by supplying the appropriate unsupported minor version number in the URL, such as `https://bootstrap.pypa.io/pip/2.7/get-pip.py`. This change would incorporate this same logic into `PyPy` installs such as `pypy2.7-7.3.12` where the minor version of Python is specified.

### Tests

* Adds unittests to `pyenv_ext.bats` for previously untested and new pinned bootstrap script versions.


TimPansino:
Let me know if an accompanying issue is required and I can open one.

Issue #1417, opened 2019-10-15
Add python-build-standalone pre-built redistributable Python releases


jayvdb:
https://github.com/indygreg/python-build-standalone [releases](https://github.com/indygreg/python-build-standalone/releases) contain the following assets

- `cpython-3.7.4-linux64-<isodate>.tar.zst` (24 MB)
- `cpython-3.7.4-linux64-musl-<isodate>.tar.zst` (23.1 MB)
- `cpython-3.7.4-macos-<isodate>.tar.zst` (20.9 MB)
- `cpython-3.7.4-windows-amd64-<isodate>.tar.zst` (18.3 MB)
- `cpython-3.7.4-windows-x86-<isodate>.tar.zst` (16.5 MB)

These are pre-built as the environment and configuration to build them is quite specific to ensure that output is suitable for redistributable builds of Python.  i.e. it uses static links against the dependencies, and therefore it needs to choose dependencies to avoids static linkages where the license doesnt permit them, such as removing GPL dependencies (GDBM and readline).

Notably it provides musl builds.

native-api:
I guess as long as you're willing to write and maintain the necessary installation scripts yourself... What do other team members think?

For starters, how likely are these builds to have incompatibilities (since it's here where users are going to go with their problems and we have no idea what to respond to them)?
Since these are prebuilt, they must be built with some assumptions about the system in mind. You'll likely need to insert some kind of guard into the insallation script (like [`require_distro()`](https://github.com/pyenv/pyenv/blob/debf1abd8bd624ba8f9f902f00e5c9098e99dede/plugins/python-build/bin/python-build#L1264-L1277)) to make sure the user's system is compatible.

anton-petrov:
Personally, I do not support this idea of ​​depending on an external source of dubious origin. What is the need for this, especially with regard to standard CPython? Standard python-build works well and that's more than enough.  So, I see no reason to invest additional resources in development of this feature.

native-api:
@jayvdb Spelunking in `python-build`, I discovered that you don't even need to fork!

Just place your definitions into `$PYENV_ROOT/plugins/*/share/python-build` -- [and `pyenv install` will see them](https://github.com/pyenv/pyenv/blob/d02b929d07182578bc16eb256b6b9a5e4c7a2db8/plugins/python-build/bin/pyenv-install#L33)!

So you can add them with a plugin.

ketozhang:
[...]

PR #2796, opened 2023-09-19
Add graalpy-23.1.0 and split between graalpy and graalpy-community

msimacek:
# Please read before merging
Apart from updating GraalPy version to 23.1.0, I'm introducing another change. We now release two versions of GraalPy:
- *Oracle GraalPy*, which is distributed under [Graal Free License](https://www.oracle.com/downloads/licenses/graal-free-license.html). This license permits free use, including commercial use, but is not open source.
- *GraalPy Community*, which is distributed under the [Universal Permissive License](https://github.com/oracle/graalpython/blob/master/LICENSE.txt). This hasn't changed, that's the version we've always had here in pyenv under the same open source license.

The difference in the versions is mostly performance, as *Oracle GraalPy* features more compiler optimizations. The python runtime is identical with the exception of two modules, `struct` and `pickle` that have optimized versions in *Oracle GraalPy*.

We would like to have both versions available, so I added separate install scripts for them. The naming follows the current upstream naming, where `graalpy` refers to *Oracle GraalPy* and `graalpy-community` refers to *GraalPy Community*. Please confirm if this is ok.

CC @timfel

### DIFF
--- a/plugins/python-build/share/python-build/graalpy-community-23.1.0
+++ b/plugins/python-build/share/python-build/graalpy-community-23.1.0
@@ -1,0 +1,54 @@
+# Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+# of the Software, and to permit persons to whom the Software is furnished to do
+# so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


native-api (changes requested):
Many thanks for the warning!

IMO we cannot provide non-OSS flavors in the core codebase. At the very least, because we can't help users solve problems with them -- which they will come here with.

So the `graalpy` file has to go.

@pyenv/pyenv-core-maintainers any second opinions?

native-api:
@msimacek If you wish to supply non-OSS GraalPy, you can do that with a Pyenv plugin as per https://github.com/pyenv/pyenv/issues/1417#issuecomment-845954847 .

timfel:
I would still suggest to rename the community version of graalpy to graalpy-community (as already done in this PR), if only to communicate to users that this is not the best version to do benchmarking with. @native-api would you mind if we inform the users about the faster version, either when installing graalpy-community or by leaving the graalpy script in place and only informing the user that they need to get it using a plugin?

native-api:
> I would still suggest to rename the community version of graalpy to graalpy-community (as already done in this PR)

Agree.



> would you mind if we inform the users about the faster version, either when installing graalpy-community or by leaving the graalpy script in place and only informing the user that they need to get it using a plugin?

I think the rename to `graalpy-community` is enough of a hint that there's a "non-community" version, too. You can detail how to install it prominently at the project's site or blog.

msimacek:
> So the `graalpy` file has to go.

Done, now there's just `graalpy-community`.

native-api (approved):

native-api:
> or by leaving the graalpy script in place and only informing the user that they need to get it using a plugin?

I considered this, too, and it seems like polluting the codebase with unrelated stuff.
Plus, a broken script is going to cause trouble for all sorts of automated machinery that picks the latest version.

PR #2810, opened 2023-10-03
Install named version support

raven42:

### Prerequisite

### Description

This PR will add support for installing a named version using the
following:

```
pyenv install --name <name> <version>
```

This will add the user supplied `<name>` to the end of the version as
`<version>-<name>`. For example:

```
~$ pyenv install --name test-environment 3.11.3
Downloading Python-3.11.3.tar.xz...
-> https://www.python.org/ftp/python/3.11.3/Python-3.11.3.tar.xz
Installing Python-3.11.3...
Installed Python-3.11.3 to /home/user/.pyenv/versions/3.11.3-test-environment
~$ pyenv versions
* system (set by /home/user/.pyenv/version)
  3.8.5
  3.11.3
  3.11.3-test-environment
~$ pyenv 3.11.3-test-environment
~$ pyenv global 3.11.3-test-environment
~$ pyenv versions
  system
  3.8.5
  3.11.3
* 3.11.3-test-environment (set by /home/user/.pyenv/version)
~$
```

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ This project was forked from [rbenv](https://github.com/rbenv/rbenv) and
   * [Install additional Python versions](#install-additional-python-versions)
     * [Prefix auto-resolution to the latest version](#prefix-auto-resolution-to-the-latest-version)
     * [Python versions with extended support](#python-versions-with-extended-support)
-    * [Install a named python version](#intstall-a-named-python-version)
+    * [Install a named python version](#install-a-named-python-version)
   * [Switch between Python versions](#switch-between-python-versions)
   * [Uninstall Python versions](#uninstall-python-versions)
   * [Other operations](#other-operations)
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ This project was forked from [rbenv](https://github.com/rbenv/rbenv) and
   * [Install additional Python versions](#install-additional-python-versions)
     * [Prefix auto-resolution to the latest version](#prefix-auto-resolution-to-the-latest-version)
     * [Python versions with extended support](#python-versions-with-extended-support)
+    * [Install a named python version](#intstall-a-named-python-version)
   * [Switch between Python versions](#switch-between-python-versions)
   * [Uninstall Python versions](#uninstall-python-versions)
   * [Other operations](#other-operations)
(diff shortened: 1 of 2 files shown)


anton-petrov:
Whats the reason for this feature? Why not use specialized software to create a virtual environment?

raven42:
I suppose I can make venv work. My thoughts were to limit the tools needed for shell integration for a shared installation while maintaining my own version of the same python version.

anton-petrov:
@raven42  
Ok, I understood your point, thanks! Personally, I’m not against this functionality, if anyone needs it, let’s add it.

@yyuu @native-api how do you think?

anton-petrov:
PR is open

raven42:
Note: I actually found a bug in this PR if anybody resurrects it. I am moving forward with using the venv module for my purposes.

The following change is needed. It was missing the `! -z "$NAME"` check:
```python
# If the `name` argument has been used, validate the name given is valid
if [[ ! -z "$NAME" &&  ! $(grep '^[-0-9a-zA-Z_]*$' <<< $NAME) ]]; then
  echo -e "Error: --name argument can only contain alpha-numeric, '-', and '_' characters\n" >&2
  usage 1 >&2
fi
```

native-api:
Such a feature has been requested a number of times. Currently, the suggested solution is https://github.com/s1341/pyenv-alias but it's been broken since Pyenv 2.3.10 . Perhaps it's time to include this into the base.

There are a few points though:

* IMO "name" is misleading -- because it's only a suffix, not a full name.
  * I actually like the idea of a suffix though -- as it provides a clear and consistent way to specify both the origin and the customizations made.
* Will this work if multiple versions are passed to `pyenv install`? Which of them will it be applied to?

AdrianDAlessandro:
I forked that pyenv-alias plugin to make it work with newer versions, but was only able to make it work with suffixes (which I prefer and think makes more sense) - https://github.com/AdrianDAlessandro/pyenv-suffix

I'm not intending on developing it into the future (only so far as to make it continue working for me). It would be useful to have as part of the in-built package and use a suffix taken from an environment variable (in my fork it is PYENV_VERSION_SUFFIX)

To explain my use case: I use an Apple Silicon MacBook and I use pyenv to switch between ARM and x86 versions of Python, but I like to use the exact same version numbers of Python and want to be able to tell which architecture I am using. This is what my installed versions list looks like:

```bash
$ pyenv versions
  system
  3.8.16
  3.9.16
  3.9.16x86
* 3.10.9 (set by /home/user/.pyenv/version)
  3.10.9x86
  3.11.1
  3.11.1x86
```

Methodology is [explained in this blog post](https://blogs.imperial.ac.uk/research-software-engineering/2023/04/20/python-development-on-m1-macs/), if anyone is interested.

PR #2813, opened 2023-10-05
Install ncurses from Homebrew, if available

aphedges:
XCode Command Line Tools 15.0 was recently released, and it contains a broken version of ncurses 6.0. Some uses of Python's `curses` module will segfault when compiled with it. The solution is to switch to using the version of ncurses from Homebrew, which is currently 6.4. Support for ncurses 6 was added to Python 3.7 and was backported to 3.6 and 2.7, so this change should not break any recently supported Python versions.

I tested this commit with Python 3.12, 3.11, and 2.7, and all tests in the `test.test_curses` module passed without issue.

See https://github.com/python/cpython/issues/109617 and https://github.com/python/cpython/issues/69906 for more information.

---

Feel free to edit the commit message as needed, but would you consider keeping more of it than in #2768? It's sometimes useful to grep the Git logs when developing.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -803,6 +803,7 @@ build_package_standard_build() {
     use_homebrew || true
     use_tcltk || true
     use_homebrew_readline || use_freebsd_pkg || true
+    use_homebrew_ncurses || true
     if is_mac -ge 1014; then
       use_xcode_sdk_zlib || use_homebrew_zlib || true
     else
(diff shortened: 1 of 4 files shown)


aphedges:
I'd appreciate some help with these test failures. They don't seem to be related to my change, and I can't reproduce them locally on my macOS 13 machine.

native-api:
The failures, at least the ones with `unstub`, are related to the changes because they change the number of calls to `uname`. You need to adjust the number of `stub uname` calls in the corresponding tests.

native-api:
Will also need a test for the added functionality. There are tests for other lib searches which you can take as example.

aphedges:
Thank you very much! I figured out how to reproduce these errors, and increasing the number of calls to `uname` fixed some of the failures. I'll have the tests fixed and new tests added later this week.

aphedges:
@native-api, I've fixed the broken tests and added a new one.

aphedges:
Thanks for merging!

Issue #2794, opened 2023-09-18
RuntimeError: tk.h version (8.5) doesn't match libtk.a version (8.6) python 2.7.18


startergo:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [ ] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [ ] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [ ] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [ ] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [ ] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [ ] Platform information macOS 13.4.1 (c) (22F770820d)
- [ ] OS architecture x64:
- [ ] pyenv version: 2.7.18
- [ ] Python version: 2.7.18
- [ ] C Compiler information Apple clang version 14.0.3 (clang-1403.0.22.14.1)

```
python
Python 2.7.18 (default, Sep 18 2023, 17:53:40) 
[GCC Apple LLVM 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import Tkinter
>>> Tkinter._test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mbp151/.pyenv/versions/2.7.18/lib/python2.7/lib-tk/Tkinter.py", line 3842, in _test
    root = Tk()
  File "/Users/mbp151/.pyenv/versions/2.7.18/lib/python2.7/lib-tk/Tkinter.py", line 1827, in __init__
    self._loadtk()
  File "/Users/mbp151/.pyenv/versions/2.7.18/lib/python2.7/lib-tk/Tkinter.py", line 1843, in _loadtk
    % (_tkinter.TK_VERSION, tk_version)
[...]

PR #2820, opened 2023-10-15
Fix linking against Homebrew's Tcl/Tk 8.6.13 in MacOS

startergo:
Correct the path for `tcl-tk` `include` for homebrew

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2794

### Description
https://github.com/pyenv/pyenv/issues/2794#issuecomment-1763182002
### Tests
```
pyenv install 2.7.18
python -m Tkinter
```


plugins/python-build/bin/python-build
```diff
@@ -1645,7 +1645,7 @@ use_homebrew_tcltk() {
     if [[ -z "$PYTHON_BUILD_TCLTK_USE_PKGCONFIG" ]]; then
       local tcltk_version="$(sh -c '. '"$tcltk_libdir"'/lib/tclConfig.sh; echo $TCL_VERSION')"
       package_option python configure --with-tcltk-libs="-L$tcltk_libdir/lib -ltcl$tcltk_version -ltk$tcltk_version"
-      package_option python configure --with-tcltk-includes="-I$tcltk_libdir/include"
+      package_option python configure --with-tcltk-includes="-I$tcltk_libdir/include/tcl-tk"
```
native-api, line 1648: This seems to only be true for the latest Homebrew package. So we can't apply this universally.
```
$ apt-file search /tcl.h
<...>
tcl8.6-dev: /usr/include/tcl8.6/tcl.h
```
```
% find / -name tcl.h 2>/dev/null
/usr/local/Cellar/tcl-tk/8.6.13_5/include/tcl-tk/tcl.h
<...>
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include/tcl.h
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/Tcl.framework/Versions/8.5/Headers/tcl.h
/Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/tcl.h
/Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/System/Library/Frameworks/Tcl.framework/Versions/8.5/Headers/tcl.h
```

native-api (changes requested):

startergo (commented):
Added conditional coding of `include` folder rather than hardcoded one

startergo (commented):
`tcl.h ` can be in multiple locations. How can we ensure `tcl-tk` from Homebrew is used in the `use_homebrew_tcltk()`?

native-api:
This still links 2.7.18 against 8.5 . [As I said, 2.7.18's Tkinter build is bugged](https://github.com/pyenv/pyenv/issues/2794#issuecomment-1727580772), it appends rather than prepends the required directory to the compiler search path -- so it doesn't pick the intended version when other directories or the compiler search path also have it.

Looks like to work around that, we'd need to include the mode to configure Tcl/Tk via `CPPFLAGS`/`LDFLAGS` rather than through `--with-tcltk-*`.

---

However.

I've looked through the history... and we seem to have never done this before! So, configuring 2.7 with Homebrew's Tcl/Tk never worked! I was suspecting that this worked before and then we broke it -- then this would be a bugfix.
But it's not. This'll be a new feature.

And because 2.7 is EOL, adding new features just for it would be against our [Deprecation policy](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md#deprecation-policy).

---

So I suggest that we should rather make a patch for 2.7.18 that would properly prepend the dirs to the compiler's search paths.

native-api:
> And because 2.7 is EOL, adding new features just for it would be against our [Deprecation policy](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md#deprecation-policy).

That said, PyPy-2.7 is NOT EOL -- so it this feature applies to it as well, it's fine to add.

startergo:
> This still links 2.7.18 against 8.5

True, but at least it is not borked. Interestingly if I hardcode it like this:
```
package_option python configure --with-tcltk-includes="-I$tcltk_libdir/include/tcl-tk"
```
it links `2.7.18` against `8.6`, but if I do conditional setup like this:
```
if [ -d "$tcltk_libdir/include/tcl-tk" ]; then
   package_option python configure --with-tcltk-includes="-I$tcltk_libdir/include/tcl-tk"
fi
package_option python configure --with-tcltk-includes="-I$tcltk_libdir/include"	
```
It links against `8.5`. Go figure...

startergo:
> Looks like to work around that, we'd need to include the mode to configure Tcl/Tk via `CPPFLAGS`/`LDFLAGS` rather than through `--with-tcltk-*`.

Isn't it better to rely on `pkg-config` and include it in the `brew` requirement? At least it would use relative paths.

startergo (commented):
aah! my bad...

native-api:
Wow! I don't know how but it works.

```
% find ~/.pyenv/versions/2.7.18 -name _tkinter\* | xargs otool -L
/Users/admin/.pyenv/versions/2.7.18/lib/python2.7/lib-dynload/_tkinter.so:
	/usr/local/opt/tcl-tk/lib/libtcl8.6.dylib (compatibility version 8.6.0, current version 8.6.13)
	/usr/local/opt/tcl-tk/lib/libtk8.6.dylib (compatibility version 8.6.0, current version 8.6.13)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
```

startergo (commented):
Makes sense. Thanks.

native-api:
Thanks to you, too! Way to prove me wrong!

PR #2869, opened 2023-12-22
Add micropython 1.20.0 and 1.21.0

cpzt:

### Prerequisite

### Description

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -925,7 +925,13 @@ build_package_ironpython_builder() {
   ( cd "Stage/Release/IronPython-"* && build_package_ironpython )
 }
 
+build_package_micropython_1_9() {
+  # supported version 1.9.3 and 1.9.4
+  build_package_micropython "with_axtls"
+}
+
 build_package_micropython() {
+  # supported version 1.10 and higher
   if [ "${MAKEOPTS+defined}" ]; then
     MAKE_OPTS="$MAKEOPTS"
   elif [ -z "${MAKE_OPTS+defined}" ]; then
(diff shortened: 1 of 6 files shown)


plugins/python-build/bin/python-build
```diff
@@ -934,7 +934,7 @@ build_package_micropython() {
   { cd mpy-cross
     "$MAKE" $MAKE_OPTS
     cd ../ports/unix
-    "$MAKE" $MAKE_OPTS axtls
+    grep -qE '^axtls:' Makefile && "$MAKE" $MAKE_OPTS axtls
```
native-api, line 937: A comment explaining the idea behind this would be welcome.
From what I saw, this was needed in some old versions.

native-api (approved):

native-api (changes requested):
Please also review the test failures

plugins/python-build/bin/python-build
cpzt: In the latest few versions, installing MicroPython doesn't require running the `make axtls` command; otherwise, it might result in an error. This change checks the makefile for dependencies on axtls to determine if the `make axtls` command needs to be executed.

cpzt:
> Please also review the test failures

These failures seem unrelated to the current PR

plugins/python-build/bin/python-build
native-api: > In the latest few versions, installing MicroPython doesn't require running the `make axtls` command; otherwise, it might result in an error. This change checks the makefile for dependencies on axtls to determine if the `make axtls` command needs to be executed.

Yes. I meant, a comment in the code saying that briefly would be welcome -- 'cuz that knowledge is a specter (piece of knowledge crucial for the code to work that cannot be expressed in the code itself).

native-api:
> > Please also review the test failures
> 
> These failures seem unrelated to the current PR

`not ok 66 passthrough CFLAGS_EXTRA to micropython compiler` is definitely related

cpzt:
I adjusted the code. Version 1.9.x requires make axtls, while later versions do not require it.

native-api:
Looks like axtls cannot build in MacOS because it's using a Linux-specific function: https://github.com/pfalcon/axtls/issues/1
Not sure if we should place a warning. Let's leave it as is, if there are complaints, we can add a warning.

native-api (approved):

native-api:
Thank you for resolving the Micropython issue!

PR #2903, opened 2024-02-15
Add CPython 3.13.0a4

saaketp:
3.13.0a3 -> 3.13.0a4

### DIFF
--- a/plugins/python-build/share/python-build/3.13.0a4
+++ b/plugins/python-build/share/python-build/3.13.0a4
@@ -1,6 +1,6 @@
 prefer_openssl3
 export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
-install_package "openssl-3.1.2" "https://www.openssl.org/source/openssl-3.1.2.tar.gz#a0ce69b8b97ea6a35b96875235aa453b966ba3cba8af2de23657d8b6767d6539" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-3.2.1" "https://www.openssl.org/source/openssl-3.2.1.tar.gz#83c7329fe52c850677d75e5d0b0ca245309b97e8ecbcfdc1dfdc4ab9fac35b39" mac_openssl --if has_broken_mac_openssl
 install_package "readline-8.2" "https://ftpmirror.gnu.org/readline/readline-8.2.tar.gz#3feb7171f16a84ee82ca18a36d7b9be109a52c04f492a053331d7d1095007c35" mac_readline --if has_broken_mac_readline
 if has_tar_xz_support; then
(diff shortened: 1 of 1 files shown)


plugins/python-build/share/python-build/3.13.0a4
```diff
@@ -3,7 +3,7 @@ export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
 install_package "openssl-3.1.2" "https://www.openssl.org/source/openssl-3.1.2.tar.gz#a0ce69b8b97ea6a35b96875235aa453b966ba3cba8af2de23657d8b6767d6539" mac_openssl --if has_broken_mac_openssl
```
edgarrmondragon, line 3: It probably doesn't hurt to get the latest OpenSSL when needed

```suggestion
install_package "openssl-3.2.1" "https://www.openssl.org/source/openssl-3.2.1.tar.gz#83c7329fe52c850677d75e5d0b0ca245309b97e8ecbcfdc1dfdc4ab9fac35b39" mac_openssl --if has_broken_mac_openssl
```

native-api:
The MacOS build fails. Looks like `python` is looking for `libpython` at a wrong prefix. Is this a known issue?

native-api:
This seems to be an upstream issue:

```
2024-02-16T08:17:42.5537610Z dyld: Library not loaded: /lib/libpython3.13.dylib
2024-02-16T08:17:42.5566900Z   Referenced from: /Users/runner/work/pyenv/pyenv/versions/3.13.0a4/bin/python3.13
2024-02-16T08:17:42.5610750Z   Reason: image not found
```
(earlier)
```
2024-02-16T08:16:56.0094360Z clang -dynamiclib <...> -Wl,-rpath,/Users/runner/work/pyenv/pyenv/versions/3.13.0a4/lib <...> -Wl,-install_name,/lib/libpython3.13.dylib <...> -o libpython3.13.dylib <...>
```
the argument before `/lib` is Makefile var `PYTHONFRAMEWORKINSTALLNAMEPREFIX` which is only set for a framework build.

nedbat:
This shouldn't have been merged if it doesn't build. Yes, it's an upstream issue, but now there isn't a way to get a 3.13 alpha on Mac.  Have you reported the bug upstream?

native-api:
The logic was, it's a prerelease, more so, an alpha, so you can expect issues and the need to work them out.

nedbat:
Yes, except now Mac users cannot work them out.  There's no version that is buildable.  This error can be reported to CPython, and leave the PR unmerged.  Then CPython can fix the build errors with a4, while Mac users can continue to try out a3.  Or is someone else working on the build issue with a4?

native-api:
Okay, [reverted](https://github.com/pyenv/pyenv/commit/f9a2bb81b69bc2fc45753f7da5d246bc2706f01d).

nedbat:
Thanks, sorry to be pushy about it.. <3

native-api:
[Upstream report](https://github.com/python/cpython/issues/115737).

freakboy3742:
[Candidate fix](https://github.com/python/cpython/pull/115750)

PR #2908, opened 2024-02-24
Handle case where `pyenv-commands --sh` returns nothing

aphedges:
When running `eval "$(pyenv init -)"` to initialize `pyenv`, the command generates a shell script to be executed. When not using the `fish` shell, this script contains a `case` statement where the first case matches against the output of `pyenv-commands --sh` and the second case handles everything else. However, if no commands are returned, then the matching expression will be only `)`, which is invalid Bash.

I have solved this by only including the first case if there are commands to match. The resulting `case` statement will have only a single, default case, which is somewhat silly, but I wanted to minimize the variation across runs of `pyenv init -`. To catch any regressions, I have added a test to catch unexpected changes in the generated `case` statement.

I do not know if `pyenv-commands --sh` can ever return nothing under normal conditions, but it happened to me due to completely unrelated problems in underlying Unix utilities called by `pyenv`. Although this bug will rarely occur, I feel that this increased robustness will be useful.

---

Feel free to edit the commit message as needed, but I would appreciate if you kept most of it. I sometimes find it useful to grep the Git logs when developing.

This bug might also exist in the generated `fish` script, but I know nothing about the `fish` shell. [`case`](https://fishshell.com/docs/current/cmds/case.html) doesn't explain what happens when no pattern is provided.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -297,7 +297,7 @@ EOS
   fi
 
   case "\$command" in
-  ${commands[*]})
+  ${commands[*]:-/})
     eval "\$(pyenv "sh-\$command" "\$@")"
     ;;
   *)
--- a/test/init.bats
+++ b/test/init.bats
@@ -2,6 +2,18 @@
 
 load test_helper
 
+setup() {
+  export PATH="${PYENV_TEST_DIR}/bin:$PATH"
+}
+
+create_executable() {
+  local name="$1"
+  local bin="${PYENV_TEST_DIR}/bin"
+  mkdir -p "$bin"
+  sed -Ee '1s/^ +//' > "${bin}/$name"
+  chmod +x "${bin}/$name"
+}
+
 @test "creates shims and versions directories" {
   assert [ ! -d "${PYENV_ROOT}/shims" ]
   assert [ ! -d "${PYENV_ROOT}/versions" ]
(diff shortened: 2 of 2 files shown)


libexec/pyenv-init
```diff
@@ -297,9 +297,15 @@ EOS
   fi
 
   case "\$command" in
+EOS
+  if [ ${#commands[@]} -ne 0 ]; then
+    cat <<EOS
   ${commands[*]})
```
native-api, line 303: I think the fix can be simplified to

```suggestion
  "${commands[@]}")
```

native-api: Scratch that. That would match the absense of a command argument -- which is valid input.
Still, it rubs me the wrong way to add loads of code for a case that never happens in practice. Would very much prefer smth more naturally flowing.

aphedges: I'm not really a fan of this code, either, but it's the best I could come up with to avoid difficult-to-debug errors caused by `grep` and `sort` being miscompiled on my system.

An alternative implementation I considered was adding a placeholder command if the array is empty, but as far as I know, a plugin author is free to name custom commands whatever they want, which could cause a conflict.

native-api: Okay, take 2:

```suggestion
  "${commands[@]:-/}")
```

`/` is the only character in Linux besides `\0` that filenames cannot contain. So a valid command can never be "`/`".

native-api: Take 3:

```suggestion
  ${commands[*]:-/})
```

This one expands correctly both when the array in empty and non-empty

aphedges: @native-api, I can confirm that works! I have modified `pyenv-init`, the test, and the commit message accordingly.

PR #2909, opened 2024-02-27
Add new miniforge and mambaforge definitions

aphedges:
I used `add_miniforge.py` to generate definitions files for missed versions released over the past year.

In addition, I realized that the script had some problems pulling all versions, so I adjusted the code accordingly.

I tested this by successfully installing and running `miniforge3-23.11.0-0` on x86_64 macOS.


aphedges:
I'm not entirely sure what is causing the failures. All of them were caused by `403` errors in conda packages being pulled, but when I checked <https://conda.anaconda.org/conda-forge/osx-64/libev-4.33-haf1e3a3_1.tar.bz2>, which caused the failure in [macos_build (mambaforge-22.11.1-2, macos-12)](https://github.com/pyenv/pyenv/actions/runs/8057336828/job/22008241123?pr=2909#logs), I got a `200` status.

I suspect we hit a rate limit or some other sort of network-related problem. Can you rerun just the tests that failed?

native-api:
Did a rerun, same problem.
Looks like the same problem as https://github.com/conda/conda/issues/11638

native-api:
Tested at home and got the same errors.
The failed versions (22.11.1-0 through 22.11.1-2) are clearly unusable as of now.

Reported to https://github.com/conda/conda/issues/13626

native-api (approved):
I suggest we commit this without the broken versions and re-add them when they're fixed.

Is this okay for you?

aphedges:
> I suggest we commit this without the broken versions and re-add them when they're fixed.
> 
> Is this okay for you?

Works for me!

I find this problem very weird, though. I ran `./plugins/python-build/bin/python-build mambaforge-22.11.1-2 versions/mambaforge-22.11.1-2` and it failed, but I was able to `curl -I https://conda.anaconda.org/conda-forge/osx-64/c-ares-1.18.1-h0d85af4_0.tar.bz2` the failing URL without issue.

PR #2922, opened 2024-03-06
minor grammar fix in libffi backport patch in 2.5.x

cuinix:

### Prerequisite

### Description
remove repetitive words 

### Tests


native-api:
Wow... I wonder what we can even use to test the change! It definitely isn't compatible with current compilers. The change looks innocent but I'd like to try just for the challenge.
Looks like MacOS 10.3 supports 2.5.

native-api:
Succeeded in MacOS 10.13.6 VM. Many extensions failed to build but ctypes (which is the module that uses `libffi`) built and tests passed.

PR #2927, opened 2024-03-19
feat(python-build): Add PACKAGE_LDFLAGS to support custom LDFLAGS per package

cfxegbert:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2926

### Description
Adds PACKAGE_LDFLAGS to be able to pass LDFLAGS separately to readline and python

### Tests


native-api (commented):
Wow. I was sure this was already available.

Could you adjust the tests, too? We can probably test this in the same test that tests `PACKAGE_CFLAGS`.

native-api:
> We can probably test this in the same test that tests `PACKAGE_CFLAGS`.

...And my memory fails me here, too: there is not such test.
There is however `"yaml is installed for python"` is `build.bats` which builds a dependency package and shows its build flags -- this can be taken as a base.

---

It's also worth documenting the new envvar.

native-api:
Superseded by https://github.com/pyenv/pyenv/pull/3130

PR #2966, opened 2024-05-21
WIP: Add version aliases

### LINKED ISSUE #218: Install python version under a different name
What I would like is to install multiple copies of the same version of python, with different build options.

My current solution for this is to download the `python-build` script and build with 
`PYTHON_CONFIGURE_OPTS="..." python-build -v 2.7.8 ~/.pyenv/versions/2.7.8-custom1`
but what I would rather do is something like
`PYTHON_CONFIGURE_OPTS="..." pyenv install 2.7.8 2.7.8-custom1`
I think this would most closely match the way virtualenvs are created.

kmsquire:

### Prerequisite
    * **This is a modification of an existing plugin (pyenv-build)**
    * **Not applicable; this modification happens in the pyenv-build plugin, which does not exist upstream**
  - Closes https://github.com/pyenv/pyenv/issues/218
 
### Description

An older plugin, pyenv-alias, allowed naming the desired python version with an alias, which allowed installing multiple versions of the same python version.  This is particularly useful on Apple silicon if you want to install both arm and x86_64 versions of python.

Unfortunately, this plugin broke when the code was modified to allow multiple versions to be specified at once on the command line, and the updated code does not allow non-invasive modification of the version number.

### Tests

    * None

### DIFF
--- a/plugins/python-build/bin/pyenv-install
+++ b/plugins/python-build/bin/pyenv-install
@@ -123,6 +123,36 @@ DEFINITIONS=("${ARGUMENTS[@]}")
 [[ "${#DEFINITIONS[*]}" -eq 0 ]] && DEFINITIONS=($(pyenv-local 2>/dev/null || true))
 [[ "${#DEFINITIONS[*]}" -eq 0 ]] && usage 1 >&2
 
+is_array() {
+    # no argument passed
+    [[ $# -ne 1 ]] && echo 'Supply a variable name as an argument'>&2 && return 2
+    local var=$1
+    # use a variable to avoid having to escape spaces
+    local regex="^declare -[aA] ${var}(=|$)"
+    [[ $(declare -p "$var" 2> /dev/null) =~ $regex ]] && return 0
+}
+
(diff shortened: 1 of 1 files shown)


kmsquire:
I just realized that I didn't add documentation.  If this functionality looks reasonable to add, I'll be sure to do so!

native-api:
I really don't like the idea of having to supply an argument via an envvar -- but can't readily recall anything better.

I guess we should do a brainstorm in a discussion. Someone must have been solving the same problem before!

An idea from the top of my head is an array option -- one that can be specified multiple times.
Another is to somehow pair arguments and the options.

kmsquire:
I'm okay with either an array option, or pairing the arguments and options.  I think the array option would be easier to implement.

kmsquire:
So it turns out that I broke non-aliased builds in this.  I'll look into it.

native-api:
Got a nice-looking idea of how to couple a version and its alias: with a colon separator. A colon shouldn't normally be present in directory names since it's `PATH` separator.

native-api:
Superseded by https://github.com/pyenv/pyenv/pull/3484

PR #2969, opened 2024-05-23
Add script to install graalpy development builds

timfel:
GraalPy now publishes regular early access builds at https://github.com/graalvm/graal-languages-ea-builds. It'd be nice to be able to easily install the latest one for users to try bugfixes that are not in a release, yet.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -397,7 +397,7 @@ http_get_aria2c() {
   fi
 
   if aria2c --allow-overwrite=true --no-conf=true -d "${dir_out}" -o "${out}" ${ARIA2_OPTS} "$1" >&4; then
-    [ -n "$2" ] || cat "${out}"
+    [ -n "$2" ] || cat "${dir_out:-.}/${out}"
   else
     false
   fi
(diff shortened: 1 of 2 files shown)


timfel:
I'm not sure what the issue is with the macos builds, is there a way to get more output? I tried it just now with a clean mac instance from scaleway.com and it works 🤔

native-api:
> is there a way to get more output?

https://github.com/pyenv/pyenv/commit/948f53f6d838eb4ebe2770d19a43e2eb76e425ed
and https://github.com/pyenv/pyenv/commit/0e0ca0966bf61dc89de376537a2fcc8492d70347

plugins/python-build/bin/python-build
```diff
@@ -397,7 +397,7 @@ http_get_aria2c() {
   fi
 
   if aria2c --allow-overwrite=true --no-conf=true -d "${dir_out}" -o "${out}" ${ARIA2_OPTS} "$1" >&4; then
-    [ -n "$2" ] || cat "${out}"
+    [ -n "$2" ] || cat "${dir_out}/${out}"
```
timfel, line 400: This would never `cat` the right file. I suppose no one else is using `http get` on systems with aria2c and without an output filename 😉

native-api:
Nice job hunting the issue! :tada:

PR #2974, opened 2024-06-01
feat: Allow major or minor version lookup

reneleonhardt:

### Prerequisite

### Description

### Features
* `pyenv local system` -> if system is installed, removes existing .py-env.version if it exists printing the previous version
> pyenv: using system version instead of 3.12.3 now

* `pyenv local latest` -> 3.12.3 (latest installed version)
* `pyenv local 3` -> 3.12.3 (latest installed major version)
* `pyenv local 12` -> 3.12.3 (latest installed minor version when combined with a major version)
* `pyenv local 3.12` -> 3.12.3 (latest installed minor version)
* `pyenv 3` assumes `pyenv local 3` if any of those single version identifiers are given
* That means new "virtual" commands `pyenv latest` and `pyenv system`
* `pyenv global` supports the same version identifiers like `local` (both use pyenv-version-file-write)
* Extracted new `pyenv installed` command (leaving `prefix` unchanged for now)
* Boost grep performance

### Fixes
* Sort versions correctly (3.10.9 comes before 3.10.10), only in `prefix` and `local` for now (not inside `versions` yet)

### TODO
* Add tests if features get accepted
* Rewrite `prefix` to use `installed` (logic needs to be changed)

### Tests

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+export LC_ALL=C # boost grep performance by disabling unicode
 
 if [ "$1" = "--debug" ]; then
   export PYENV_DEBUG=1
@@ -118,16 +119,21 @@ case "$command" in
   exec pyenv-help
   ;;
 * )
-  command_path="$(command -v "pyenv-$command" || true)"
-  if [ -z "$command_path" ]; then
-    if [ "$command" == "shell" ]; then
-      abort "shell integration not enabled. Run \`pyenv init' for instructions."
-    else
-      abort "no such command \`$command'"
+  # Just a version number given (or `latest` or `system`) -> assume `pyenv local $@`
+  if grep -q -E "^[0-9]+(\.[0-9]+){0,2}$" <<<"${command}" || [ "$command" == "latest" ] || [ "$command" == "system" ]; then
+    command_path="pyenv-local"
+  else
+    command_path="$(command -v "pyenv-$command" || true)"
+    if [ -z "$command_path" ]; then
+      if [ "$command" == "shell" ]; then
+        abort "shell integration not enabled. Run \`pyenv init' for instructions."
+      else
+        abort "no such command \`$command'"
+      fi
     fi
-  fi
 
-  shift 1
+    shift 1
+  fi
   if [ "$1" = --help ]; then
     if [[ "$command" == "sh-"* ]]; then
       echo "pyenv help \"$command\""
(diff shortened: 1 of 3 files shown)


native-api:
> ### Features
> 
>     * `pyenv local system` -> if system is installed, removes existing .py-env.version if it exists printing the previous version

Introduces ambiguity (which violates Python Zen koan 12). Do you want to unset the local version or override it to `system`?
Printing stuff just in this case also introduces inconsistency.

It seems to be an XY problem; the underlying issue seems to be either `--unset` is too inconvenient to type or e.g. a way to see/change all set versions in a filesystem subtree is needed. What's your use case?

> > pyenv: using system version instead of 3.12.3 now
> 
>     * `pyenv local latest` -> 3.12.3 (latest installed version)

Not sure if hardcoding "latest" to "3" is a good idea. Granted, 4 is not planned at the moment, but why such special treatment if there's `pyenv latest` resolution? We aren't adding such special "latest" aliases for other flavors, too, are we? Let's have a wider discussion before deciding either way.

This should be possible with a plugin; if it's not, we can add a hook -- e.g. a "version resolution" hook.

>     * `pyenv local 3` -> 3.12.3 (latest installed major version)

Already works

>     * `pyenv local 12` -> 3.12.3 (latest installed minor version when combined with a major version)

Same as above for "latest". I'm hesitant to give "3" a hardcoded special treatment without a wider discussion.
 
>     * `pyenv local 3.12` -> 3.12.3 (latest installed minor version)

Already works

>     * `pyenv 3` assumes `pyenv local 3` if any of those single version identifiers are given

Ambiguous. Why `local` and not `shell`? I understand that this is probably your preferred choice -- but you cannot just force it onto everyone -- without asking them, at least.

This could be done with a plugin if we add a subcommand lookup hook -- or by just adding a "3" subcommand in the plugin since you're probably only interested in using this shorthand for a few select names.

>     * That means new "virtual" commands `pyenv latest` and `pyenv system`

Likewise.

>     * `pyenv global` supports the same version identifiers like `local` (both use pyenv-version-file-write)
> 
>     * Extracted new `pyenv installed` command (leaving `prefix` unchanged for now)
> 
>     * Boost grep performance

Fine -- if it doesn't break stuff. For one, I know from problem reports that users use unicode in `HOME`. Didn't yet encounter unicode in custom installation names, though that will become likely when we add an ability to install customized version under an alias.

> 
> ### Fixes
> 
>     * Sort versions correctly (3.10.9 comes before 3.10.10), only in `prefix` and `local` for now (not inside `versions` yet)

There was a [merged pull request to do this if the local `sort` supports `--version-sort`](https://github.com/pyenv/pyenv/pull/2405). Is this still broken in some cases?

> 
> ### TODO
> 
>     * Add tests if features get accepted
> 
>     * Rewrite `prefix` to use `installed` (logic needs to be changed)
> 
> 
> ### Tests
> 
>     * [ ]  My PR adds the following unit tests (if any)

PR #2975, opened 2024-06-01
Correct the Explanation of PATH Variable Lookup

Y-askour:
### Description
- This PR corrects a small but important detail in the documentation regarding how the system searches for executable files. The original text incorrectly stated that the operating system performs this search. The corrected text clarifies that it is actually the shell that performs the search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ to the correct Python installation.
 
 ### Understanding PATH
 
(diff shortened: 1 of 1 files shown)


native-api:
Good catch! A small but important detail :slightly_smiling_face:

PR #2990, opened 2024-06-22
Use `"` quotes for `$PYENV_ROOT` in `[[ -d ]]` check

sobolevn:
I don't think any of the checks really apply here, this is a simple typo-like change.

### Prerequisite

### Description

If a user sets `$PYENV_ROOT` to a path with ` ` (space) inside, the variable expansion might be not correct.

It is better to use explicit quoting.

### Tests

None.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -165,7 +165,7 @@ function help_() {
       fi
       echo
       echo 'export PYENV_ROOT="$HOME/.pyenv"'
-      echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"'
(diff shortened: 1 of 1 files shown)


libexec/pyenv-init
```diff
@@ -165,7 +165,7 @@ function help_() {
       fi
       echo
       echo 'export PYENV_ROOT="$HOME/.pyenv"'
-      echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"'
+      echo '[[ -d "$PYENV_ROOT/bin" ]] && export PATH="$PYENV_ROOT/bin:$PATH"'
```
native-api, line 168: Double brackets lift the requirement for quoting

sobolevn: Sorry, I don't understand your comment

native-api: Inside double brackets, there's no need to quote variables by default

Issue #2984, opened 2024-06-11
Add a nicer way to select a free-threaded build


ngoldbaum:
It would be nice if you could select a free-threaded build as a command-line option to `pyenv install`, e.g. `pyenv install --free-threaded`, just like `--debug`. Currently you need to set `CONFIGURE_OPTS=--disable-gil` in a `pyenv install` command. I also use `pyenv-version-suffix` to give it a useful name:

```
CONFIGURE_OPTS=--disable-gil PYENV_VERSION_SUFFIX='-free-threaded' pyenv install -f -v 3.13.0b2
```

So I can select it with e.g. `pyenv global 3.13.0b2-free-threaded`.

It would be nice if pyenv could handle this out of the box. It will probably be something people want to do increasly often starting with Python 3.13.

native-api:
This is possbile with a shell alias:

```
$ alias pyenv_install_free_threaded="CONFIGURE_OPTS=--disable-gil PYENV_VERSION_SUFFIX='-free-threaded' pyenv install -f -v"
$ pyenv_install_free_threaded 3.13.0b2
Downloading Python-3.13.0b2.tar.xz...
<...>
```

ngoldbaum:
Sure, that's not very discoverable though. IMO this is going to be popular enough `pyenv install` will eventually need to grow an option for it.

native-api:
> IMO this is going to be popular enough `pyenv install` will eventually need to grow an option for it.

If that'll be the case, we'll probably add separate build scripts for nogil versions. That looks like the way with the least complexity ("special cases are not special enough to break the rules").

ngoldbaum:
Alternate build scripts also fixes the problem and obviates the need for the manually-applied version suffix. That would be great.

colesbury:
I've put up a PR that adds build scripts for the free-threaded Python versions: https://github.com/pyenv/pyenv/pull/2995.

I've named them as `3.13t-dev` and 3.14t-dev`, but we can change that if there is a preference for a different naming scheme.

lysnikolaou:
Trying to install `3.13t-dev` fails for me with the following:

```
ERROR: invalid Python executable: /Users/lysnikolaou/.pyenv/versions/3.13t-dev-debug/bin/python3.13t

The python-build could not find proper executable of Python after successful build.
Please open an issue for future improvements.
https://github.com/pyenv/pyenv/issues

BUILD FAILED (OS X 14.5 using python-build 20180424)

[...]

PR #2995, opened 2024-06-27
Add support for free-threaded Python

colesbury:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2984

### Description

This adds support for free-threaded CPython (GIL disabled) with the "t" suffix. The naming is similar to what's used by [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/options/#free-threaded-support) and the official macOS and Windows installers, which use, e.g. python3.13t.

For now, I've only added 3.13t-dev and 3.14t-dev. Once 3.13.0b3 is released, we can add that too. (I don't think we want to for 3.13.0b2 because pip didn't work right out of the box.)

### Tests
N/A

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -822,6 +822,7 @@ build_package_standard_build() {
       use_homebrew_zlib || true
     fi
     use_dsymutil || true
+    use_free_threading || true
   fi
 
   ( if [ "${CFLAGS+defined}" ] || [ "${!PACKAGE_CFLAGS+defined}" ]; then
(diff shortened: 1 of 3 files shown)


native-api (commented):
I think this can be done much easier: by setting the configure options then sourcing the original script

colesbury:
I updated the scripts. Does pyenv assume the use of `bash`?

native-api:
> Does pyenv assume the use of `bash`?

Why, yes. Look at the shebangs.

native-api:
Excellent work! With users like this, who needs maintainers? :laughing:

Issue #3004, opened 2024-07-03
pyenv-prefix is slow when many virtual environments exist. 


Erotemic:
I'm noticing that my bashrc has slowed down significantly when running `pyenv prefix`, which takes >100 seconds. 

### Diagnostics

I've traced the source of the slowness to a `pyenv-which` command in the `libexec/pyenv-prefix` script.

It seems to be coming from the statement: 

```bash
      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python 2>/dev/null)" || \
          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 2>/dev/null)" || \
          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python2 2>/dev/null)"; then
```

Where the first call to `pyenv-which python` is failing, and it takes a long time when it fails. Digging into that a bit, it looks like the main line that takes time is a call to `pyenv-whence`, which is only used to print stderr information. Namely the lines in `libexec/pyenv-which` are:

```bash
  versions="$(pyenv-whence "$PYENV_COMMAND" || true)"
  if [ -n "$versions" ]; then
    { echo
      echo "The \`$1' command exists in these Python versions:"
      echo "$versions" | sed 's/^/  /g'
      echo
      echo "Note: See 'pyenv help global' for tips on allowing both"
      echo "      python2 and python3 to be found."
    } >&2
  fi
```

The `pyenv-whence` command is only slow because it is looping over the 20 different Python environments I have installed.  

However, because this is only used to echo information to stderr, and stderr is being redirected to /dev/null there is no reason to do all of that extra work. I verified that if I simply comment out those lines then everything becomes much faster.

It might make sense to add a flag to `pyenv-which` that skips generating this error message which is going to be discarded anyway. Then `pyenv-prefix` can specify this flag in its calls to `pyenv-which`, and short circuit the slowness. 

### Other Information 

I have a basic install of the lastest pyenv on Ubuntu 22.04 with no plugins. 

This looks like it is similar to https://github.com/pyenv/pyenv/issues/2918 but perhaps ultimately unrelated. 

Fortunately, I believe that I don't need to run `pyenv-prefix` in my bashrc, and removing it just fixes the issue for me, but I thought I'd report it as it might be possible to speed this command up significantly.

PR #3005, opened 2024-07-03
Speed up `pyenv prefix` by not constructing advice text when it would be discarded

Erotemic:
### Prerequisite
I don't believe this makes sense as a hook

I don't believe this makes sense to submit to rbenv, I don't see corresponding issues there. 

Closes https://github.com/pyenv/pyenv/issues/3004

### Description

This speeds up the call of `pyenv prefix` by skipping the construction of an error message that is ultimately discarded by being piped to `/dev/null`. The construction of the error message is slow when there are many (in my case 20) different Python environments. Notably, the slowness seems to only happen on the first call, but this patch eliminates it completely. 

The mechanism I used is a sneaky "_"-prefixed environment variable, which works, but I'm not sure if it is the best way to accomplish the task. However, it seemed simpler to do it this way than adding more CLI parsing, and this only needs to be handled when called by other internal code anyway.

I'm not sure if it is possible to detect if stderr for a script is being piped to /dev/null, but if there is a quick way to check for that, then that would probably make even more sense than my existing implementation, and it wouldn't require ugly local environs that I added to `pyenv-prefix`.

### Tests

I don't think this requires a test.

### DIFF
--- a/libexec/pyenv-prefix
+++ b/libexec/pyenv-prefix
@@ -30,9 +30,9 @@ OLDIFS="$IFS"
 { IFS=:
   for version in ${PYENV_VERSION}; do
     if [ "$version" = "system" ]; then
-      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python 2>/dev/null)" || \
-          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 2>/dev/null)" || \
-          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python2 2>/dev/null)"; then
+      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python --skip-advice 2>/dev/null)" || \
+          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 --skip-advice 2>/dev/null)" || \
+          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python2 --skip-advice 2>/dev/null)"; then
         shopt -s extglob
         # In some distros (Arch), Python can be found in sbin as well as bin
         PYENV_PREFIX_PATH="${PYTHON_PATH%/?(s)bin/*}"
--- a/libexec/pyenv-which
+++ b/libexec/pyenv-which
@@ -2,13 +2,14 @@
 #
 # Summary: Display the full path to an executable
 #
-# Usage: pyenv which <command> [--nosystem]
+# Usage: pyenv which <command> [--nosystem] [--skip-advice]
 #
 # Displays the full path to the executable that pyenv will invoke when
 # you run the given command.
 # Use --nosystem argument in case when you don't need to search command in the 
 # system environment.
-#
+# Internal switch --skip-advice used to skip printing an error message on a
+# failed search.
 
 set -e
 [ -n "$PYENV_DEBUG" ] && set -x
@@ -18,11 +19,27 @@ if [ "$1" = "--complete" ]; then
   exec pyenv-shims --short
 fi
 
-if [ "$2" = "--nosystem" ]; then
-  system=""
-else
-  system="system"
-fi
+system="system"
+SKIP_ADVICE=""
+PYENV_COMMAND="$1"
+
+while [[ $# -gt 0 ]]
+do
+  case "$1" in
+    --skip-advice)
+      SKIP_ADVICE=1
+      shift
+      ;;
+    --nosystem)
+      system=""
+      shift
+      ;;
+    *)
+      shift
+      ;;
+  esac
+done
+
 
 remove_from_path() {
   local path_to_remove="$1"
@@ -35,8 +52,6 @@ remove_from_path() {
   result="${result%:}"
   echo "${result#:}"
 }
-
-PYENV_COMMAND="$1"
 
 if [ -z "$PYENV_COMMAND" ]; then
   pyenv-help --usage which >&2
@@ -85,16 +100,17 @@ else
   fi
 
   echo "pyenv: $PYENV_COMMAND: command not found" >&2
-
-  versions="$(pyenv-whence "$PYENV_COMMAND" || true)"
-  if [ -n "$versions" ]; then
-    { echo
-      echo "The \`$1' command exists in these Python versions:"
-      echo "$versions" | sed 's/^/  /g'
-      echo
-      echo "Note: See 'pyenv help global' for tips on allowing both"
-      echo "      python2 and python3 to be found."
-    } >&2
+  if [ -z "$SKIP_ADVICE" ]; then
+    versions="$(pyenv-whence "$PYENV_COMMAND" || true)"
+    if [ -n "$versions" ]; then
+      { echo
+        echo "The \`$PYENV_COMMAND' command exists in these Python versions:"
+        echo "$versions" | sed 's/^/  /g'
+        echo
+        echo "Note: See 'pyenv help global' for tips on allowing both"
+        echo "      python2 and python3 to be found."
+      } >&2
+    fi
   fi
 
   exit 127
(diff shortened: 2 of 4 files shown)


libexec/pyenv-which
```diff
@@ -86,15 +86,17 @@ else
...
-  versions="$(pyenv-whence "$PYENV_COMMAND" || true)"
-  if [ -n "$versions" ]; then
-    { echo
-      echo "The \`$1' command exists in these Python versions:"
-      echo "$versions" | sed 's/^/  /g'
-      echo
-      echo "Note: See 'pyenv help global' for tips on allowing both"
-      echo "      python2 and python3 to be found."
-    } >&2
+  if [ -z "$_PYENV_SKIP_STDERR" ]; then
```
native-api, line 89: It rather makes sense to make this a meaningful option to the command (e.g. "skip advice" or even "non-verbose").
It can probably be the default if the output is not to a terminal.

native-api:
> I don't think this requires a test.

The new "silent mode" definitely requires a test, to keep ensuring that it works properly going forward. It can be added at the end though.

libexec/pyenv-which
Erotemic: I like `--skip-advice`, but I think it makes sense to add it to the internal invocations to make things explicit. I'm also not sure how to check in bash if you are outputing to a terminal. In any case, I think the default would only make sense if the output wasn't going anywhere (i.e. `/dev/null`). If the user is trying to parse stderr they should be able to do so without encountering mysterious disappearing outputs because they weren't executing the command in a terminal.

Should the current argument handling:

```bash
# Provide pyenv completions
if [ "$1" = "--complete" ]; then
  exec pyenv-shims --short
fi

if [ "$2" = "--nosystem" ]; then
  system=""
else
  system="system"
fi
```

Be replaced with something like this:

```bash
system="system"
no_advice=""

for args in "$@"
do
  if [ "$args" = "--complete" ]; then
    exec pyenv-shims --short
  fi

  if [ "$args" = "--nosystem" ]; then
    system=""
    shift
  fi

  if [ "$args" = "--no_advice" ]; then
    no_advice="1"
    shift
  fi
done
```

I'm not sure if I should leave the `exec` part as it was. The behaviors are very close, but not exactly the same. The second variant lets the `--complete` argument be in any positions, not just the first.

Erotemic:
I made a modification to add `--skip-advice`, and I documented it in the script comments. I've also added a test (I haven't use bats before but it seems reasonably straightforward).

Maybe it should just be `-q / --quiet` instead? Intuitively, to me quiet means no output at all, but it seems that in other pyenv scripts it does suppress error messages only.

I was originally thinking this wasn't going to be a user-facing feature, but now that it is, is there any other place where documentation needs to be modified?

libexec/pyenv-which
native-api: > I'm also not sure how to check in bash if you are outputing to a terminal.

`[[ -t <fd> ]]`

> I like `--skip-advice`, but I think it makes sense to add it to the internal invocations to make things explicit.  In any case, I think the default would only make sense if the output wasn't going anywhere (i.e. `/dev/null`). If the user is trying to parse stderr they should be able to do so without encountering mysterious disappearing outputs because they weren't executing the command in a terminal.

The advice output is not designed to be parsed.
We would however need to add _two_ options in this case -- "skip it" and "show it" (which would be the default) -- so that the user could explicitly choose either.

I agree though that this a further, optional improvement.

libexec/pyenv-which
```diff
@@ -2,27 +2,46 @@
 #
 # Summary: Display the full path to an executable
 #
-# Usage: pyenv which <command> [--nosystem]
+# Usage: pyenv which <command> [--nosystem] [--no-advice]
```
native-api: `--skip-advice`

libexec/pyenv-which
```diff
@@ -2,27 +2,46 @@
...
 #
 # Displays the full path to the executable that pyenv will invoke when
 # you run the given command.
 # Use --nosystem argument in case when you don't need to search command in the 
 # system environment.
-#
+# Use --skip-advice to skip outputing suggestions to stderr
```
native-api: Can document the switch as internal to save us from having to maintain compatibility

libexec/pyenv-prefix
```diff
@@ -30,9 +30,9 @@ OLDIFS="$IFS"
 { IFS=:
   for version in ${PYENV_VERSION}; do
     if [ "$version" = "system" ]; then
-      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python 2>/dev/null)" || \
-          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 2>/dev/null)" || \
-          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python2 2>/dev/null)"; then
+      if PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python --skip-advice 2>/dev/null)" || \
+          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python3 --skip-advice 2>/dev/null)" || \
+          PYTHON_PATH="$(PYENV_VERSION="${version}" pyenv-which python2 --skip-advice 2>/dev/null)"; then
```
native-api, line 35: There's another such case in `pyenv-versions`.
(And also in `pyenv.d/exec/pip-rehash/*` but I'm not sure if they need to be replaced, looks like the command always exists in those cases; fine to leave them be.)

libexec/pyenv-which
Erotemic: I ended up going with the `--quiet` flag, but I can change it to `--skip-advice` depending on what maintainers prefer.

libexec/pyenv-prefix
Erotemic: Added to `pyenv-versions` and I agree it is not needed in the pip rehash scripts.

libexec/pyenv-which
Erotemic: done

libexec/pyenv-which
Erotemic: I pushed up a bunch more commits, and this is ready for a re-review.

native-api:
> Maybe it should just be `-q / --quiet` instead? Intuitively, to me quiet means no output at all

I do agree that `--quiet` is counterintuitive and don't like it.

> but it seems that in other pyenv scripts it does suppress error messages only.

You mean `pyenv-latest`? That's the only such case.
In that case, the name is appropriate because in the case of an error, it doesn't produce any normal output.
Besides, it was likewise added to accomodate an internal use case, and the resulting code as it evolved rather suggests to replace it with "`-f`" (force)/"`-b`" (bypass) with behavior "rather than print an error message, print the input unchanged".

> I was originally thinking this wasn't going to be a user-facing feature, but now that it is, is there any other place where documentation needs to be modified? I see `COMMANDS.md`, does that need an update?

I suggest we document it as internal.

> EDIT: I did switch to using `-q|--quiet`, and I reworked the CLI input parsing to be more flexible. You can now specify option-flags before or after the positional arguments. I've also added an error message if multiple positional arguments were specified. I've added corresponding tests for these cases.
> 
> I'm not sure if this added flexibility is necessary, I can revert part or all of that if desired.

I doubt these are necessary, especially in the light of this being a performance-improving PR and the fact that the switches here are rarely used manually.

We do have an outstanding idea to consolidate command line handling, probably in a loadable module (that's where such tests should probably go). However, that would too need to be weighed agains added the added lag -- and users are already complaining that the product is too slow, up to suggesting to rewrite it in a compiled language.

Erotemic:
Based on feedback I:

* change the flag back to `--skip-advice`
* documented the flag as internal
* removed the additional flexibility for arguments and corresponding tests

I squashed all of these changes down into 1 commit.

This PR can now be summarized as:

```
Add internal skip-advice flag to pyenv-which

Use this flag in pyenv-prefix and pyenv-versions to prevent constructing
advice that is ultimately discarded.

Added corresponding tests.
```

native-api:
Thank you for the contribution! Great work! :tada:

PR #3086, opened 2024-10-10
Avoid shadowing of virtualenvs with name starting by "python-"

aarbouin:

### Prerequisite

### Description

I had a virtualenv named python-api which worked correctly. Until pyenv upgrade, which results in `pyenv: version 'python-api' is not installed` error message when I tried to use any python binary.

I am not familiar with bash so I don't know if this can be fixed in pyenv-virtualenv plugin (which is probably a better place to handle this special case).

### Tests
- [x] My PR adds the following unit tests

"pyenv version started by python-"


native-api (approved):

native-api:
Great work! I probably couldn't have done it better!

PR #3112, opened 2024-11-23
README: Fix Markdown in "Notes about python releases"

noelleleigh:
HTML and Markdown need to be separated by an empty line to co-exist, otherwise the Markdown isn't parsed correctly.

### Prerequisite

### Description


native-api (approved):

native-api:
Thanks! I completely missed that bit!

Issue #3119, opened 2024-12-04
Pyenv 2.4.20 is missing Python 3.13.1 free threading build


makukha:
The latest pyenv release is missing free threading build for CPython 3.13.1.

Shouldn't `3.13.1t` version be added?

```shell
$ pyenv --version
pyenv 2.4.20
$ pyenv install --list | grep '3\.13\.'
  3.13.0
  3.13.0t
  3.13.1
```

PR #3120, opened 2024-12-04
Add CPython 3.13.1t

makukha:
- Closes https://github.com/pyenv/pyenv/issues/3119


makukha:
Should this change be tested?

Issue #3121, opened 2024-12-06
Python 3.13.1 build has the wrong path to openssl-3.4.0 release tar


dlamblin:
Installing 3.13.1 is broken if you need to build openssl-3.4.0
```
% pyenv install 3.13.1
Downloading openssl-3.4.0.tar.gz...
-> https://github.com/openssl/openssl/releases/download/openssl-3.3.2/openssl-3.4.0.tar.gz
error: failed to download openssl-3.4.0.tar.gz

BUILD FAILED (OS X 14.6.1 using python-build 2.4.21)

Results logged to /var/folders/yt/6fv7ccy95ql3cz6_8z00_hx80000gn/T/python-build.20241206115436.6774.log

Last 10 log lines:
via: 1.1 varnish
x-served-by: cache-ewr-kewr1740040-EWR
x-cache: HIT
x-cache-hits: 0
x-timer: S1733504077.891231,VS0,VE1
vary: Accept-Encoding
x-fastly-request-id: cb5cb260b1733a73c78f7fd58d09f1701d334a6a
content-length: 9379

curl: (56) The requested URL returned error: 404
```

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

Please note it **is in the default `python-build` plugin**.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): macOS Sonoma 14.6.1
- [x] OS architecture (e.g. amd64): arm64 ie Apple M1
- [x] pyenv version: 2.4.21
- [x] Python version: 3.13.1
[...]

PR #3122, opened 2024-12-06
Fix 404 for openssl-3.4.0 release in build 3.13.1

dlamblin:
The path before this change was wrong and resulted in the following message, provided here so others may search for the error message:
```
% pyenv install 3.13.1
Downloading openssl-3.4.0.tar.gz...
-> https://github.com/openssl/openssl/releases/download/openssl-3.3.2/openssl-3.4.0.tar.gz
error: failed to download openssl-3.4.0.tar.gz

BUILD FAILED (OS X 14.6.1 using python-build 2.4.21)

Results logged to /var/folders/yt/6fv7ccy95ql3cz6_8z00_hx80000gn/T/python-build.20241206115436.6774.log

Last 10 log lines:
via: 1.1 varnish
x-served-by: cache-ewr-kewr1740040-EWR
x-cache: HIT
x-cache-hits: 0
x-timer: S1733504077.891231,VS0,VE1
vary: Accept-Encoding
x-fastly-request-id: cb5cb260b1733a73c78f7fd58d09f1701d334a6a
content-length: 9379

curl: (56) The requested URL returned error: 404
```

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3121

### Description

The URL contains an old semver in the path for 3.2.2 where the file's semver is 3.4.0. This leads to a 404 from github.com. The patch corrects this to use the right path to openssl-3.4.0's release tar gz file.

### Tests

I manually tested that the fix proceeds to get and build all parts of python 3.13.3 including openssl-3.4.0. If you're not using brew but just a plain macbook's macOS, you'll need this fix.

### DIFF
--- a/plugins/python-build/share/python-build/3.13.1
+++ b/plugins/python-build/share/python-build/3.13.1
@@ -1,6 +1,6 @@
 prefer_openssl3
 export PYTHON_BUILD_CONFIGURE_WITH_OPENSSL=1
-install_package "openssl-3.4.0" "https://github.com/openssl/openssl/releases/download/openssl-3.3.2/openssl-3.4.0.tar.gz#e15dda82fe2fe8139dc2ac21a36d4ca01d5313c75f99f46c4e8a27709b7294bf" mac_openssl --if has_broken_mac_openssl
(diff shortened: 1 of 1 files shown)


native-api:
I'll complete https://github.com/pyenv/pyenv/pull/3123 first and retest this, to ensure this never happens again.

Issue #2680, opened 2023-04-14
Use installed version from .python-version even if not all versions are installed?


cquick01:
Hi, this isn't really a build issue so I got rid of the template... I tried asking on Gitter also but didn't get a response.

```
$ pyenv --version
pyenv 2.3.17-4-gb3c91b37
```

I have a project that supports Python 3.9+, so I have a `.python-version` file that contains

```
3.11
3.10
3.9
```

in an effort to allow machines that have 3.11 installed to use that, otherwise 3.10, otherwise 3.9. But this doesn't work on machines that don't have all three versions installed. It gives an error for instance if 3.10 is not installed

```
pyenv: version `3.10' is not installed (set by /home/user/project/.python-version)
```

Is it possible to configure `pyenv` to just use one of the defined versions from `.python-version` if it is installed, rather than requiring all the defined versions to actually be installed? I'd hope that in the above example, it would see 3.11 and 3.9 are available, and use one of them (preferably whichever is defined earlier in the `.python-version` file, 3.11).

Thanks in advance!

native-api:
Not possible now without workarounds:
https://github.com/pyenv/pyenv/blob/b3c91b37d63c665c89a863423cc7fbbeb0fffa46/libexec/pyenv-version-name#L53-L54

A workaround is to create dummy `3.11` and `3.10` entries under `$(pyenv root)/versions`.

cquick01:
Thanks, that gets things working for now since we control each of the machines our app runs on. It isn't super portable outside of a closed environment though, like for open-source projects for instance.

Any chance the original workflow could get supported? I'm actually surprised not to have found much other discussion about it. I'd have figured the intuitive thing to do would choose an interpreter that is available and only give an error if all can't be found. I tried digging through the code a bit myself but couldn't get it working so far.

native-api:
Of course we can support it. We just need to decide how exactly.

I think that

* executable resolution should work without errors or warnings
  * as this corresponds to the current logic of selecting the first version that has the selected executable
* `pyenv local` should error unless `-f` is passed.
* `pyenv version` should still print an error

cquick01:
I agree with each of those points. `pyenv version` looks good as is. Also like the idea of adding a `-f` arg for `pyenv local` to force writing the `.python-version` file even if not all versions are found. `pyenv local` without any args should still print the versions from `.python-version` without any errors.

`pyenv prefix` shows an error when a version is missing. Should it list available versions, or print an error, or both?

[...]

PR #3134, opened 2024-12-15
Support nonexistent versions being present and set in a local .python-version

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/2680

### Description

* If some of the set versions do not exist, executable resolution works without errors if an executable is found
  * as this corresponds to the current logic of selecting the first version that has the requested executable
* `pyenv local` supports `-f` to skip existence check

Supports deployments with a version-controlled `.python-version` that support any of a number of Python versions

### Tests
Tests for added switches for `pyenv local` and (internal) `pyenv version-name` and `pyenv version-file-write`

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -91,6 +91,10 @@ or, if you prefer 3.3.3 over 2.7.6,
     Python 3.3.3
 
 
+You can use the `-f/--force` flag to force setting versions even if some aren't installed.
+This is mainly useful in special cases like provisioning scripts.
+
+
 ## `pyenv global`
 
 Sets the global version of Python to be used in all shells by writing
--- a/libexec/pyenv-exec
+++ b/libexec/pyenv-exec
@@ -29,9 +29,9 @@ if [ -z "$PYENV_COMMAND" ]; then
   exit 1
 fi
 
-export PYENV_VERSION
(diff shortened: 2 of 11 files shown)


marhoy:
Thanks for the merge @native-api! 

Unfortunately, this doesn't solve the problem completely.
Steps to reproduce the problem:

1. In some pyenv-virtualenv, install a command-line utility like e.g. `pre-commit` or `mypy`, and  run `pyenv rehash` to update the pyenv shims.
1. There is now a corresponding shim in `~/.pyenv/shims`
1. In some other directory (outside any pyenv virtualenv): Create a `.python-version` which contains something that is *not* a pyenv virualenv (this could for instance be a project where `uv` is used to manage dependencies).

Running the command-line utility will now cause an `exit 1` from ` pyenv-version-name`, line 79 (caused by line 65).

I think the reason is that:
1. The shim calls pyenv exec
2. The shim correctly calls `pyenv-version-name -f`
3. The shim calls `pyenv-which`
4. `pyenv-which` again calls `pyenv-version-name`, but this time without the `-f`.

Thus, the solution seems to just add a `-f` in `pyenv-which`:

image

native-api:
@marhoy Done in https://github.com/pyenv/pyenv/commit/86be59a6d53d62603d5e03e93710fba5bde80ed6

Issue #2918, opened 2024-02-29
pyenv init slows down the shell startup a bunch (by about ~100 ms!)


VorpalBlade:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [X] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [X] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [X] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [X] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [X] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [X] Platform information (e.g. Ubuntu Linux 16.04): Arch Linux
- [X] OS architecture (e.g. amd64): amd64
- [X] pyenv version: 2.3.36
- [X] Python version: 3.11.7 (on host)
- [X] C Compiler information (e.g. gcc 7.3): 13.2.1 20230801
- [X] Please attach the debug trace of the failing command as a gist: https://gist.github.com/VorpalBlade/608f04c2234f187c2cd9fad652b72137
- [ ] If you have a problem with installing Python, please also attach `config.log` from the build directory: Not the issue at hand
- [ ] If the build succeeds but the problem is still with the build process (e.g. the resulting Python is missing a feature), please attach
  * the debug trace from reinstalling the faulty version with `env PYENV_DEBUG=1 pyenv install -f -k -v <version> 2>&1 | tee trace.log`
  * `config.log` from the build directory. When using `pyenv install` with `-k` as per above, the build directory will be under `$PYENV_ROOT/sources`.

pyenv init is really slow. I have been profiling zsh startup time, and commenting out `eval "$(pyenv init -)"` reduces startup time from 111.7 ms to just 11.2 ms. I have been using `hyperfine` for statistical rigour in the measurement (full output below).

[...]

PR #3136, opened 2024-12-19
pyenv-init | performance improvements

ChristianFredrikJohnsen:

### Prerequisite
  - Addresses https://github.com/pyenv/pyenv/issues/2918

### Description

1) Changed the local pyenv function generated by `pyenv init -`, and replaced calls to `cat` with calls to `echo`. 

Instead of having:

```
pyenv() {
  local command
  command="${1:-}"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  activate|deactivate|rehash|shell)
    eval "$(pyenv "sh-$command" "$@")"
    ;;
  *)
    command pyenv "$command" "$@"
    ;;
  esac
}
```

We now get:

```
pyenv() {
  local command=$1
  [ "$#" -gt 0 ] && shift
  case "$command" in
  activate|deactivate|rehash|shell)
    eval "$(pyenv "sh-$command" "$@")"
    ;;
  *)
    command pyenv "$command" "$@"
    ;;
  esac
}
```

2) DOCUMENTATION: Update README.md to make people specify their shell when setting up `pyenv init` in their shell configuration file.

Example:

Under the section: Set up shell environment for zsh:

```
echo 'eval "$(pyenv init - zsh)"' >> ~/.zshrc
```

Instead of 

```
echo 'eval "$(pyenv init -)"' >> ~/.zshrc
```

My computer uses 50ms on `pyenv init -`, and 30ms on `pyenv init - zsh`. Simple speedup.

3) General code cleanup in `pyenv-init`
Removed unnecessary declarations of variables: (e. g. `no_rehash=""`), replaced a series of if-statements with a `case` block (code essentially imported from rbenv), and change `completions path` such that we get:

```
source '/home/christian/.pyenv/completions/pyenv.zsh'
```
(essentially `PYENV_ROOT/completions/pyenv.zsh`)

instead of:

```
source '/home/christian/.pyenv/libexec/../completions/pyenv.zsh'
```
(equal to something like: `$(dirname "$(realpath pyenv-init)")/../completions/pyenv.zsh`)

RESULTS:

The first command is run by `master` branch, second command is used with perf improvements.

Bildschirmfoto vom 2024-12-19 16-50-47
Bildschirmfoto vom 2024-12-19 16-52-16

Not much to speak of, but will hopefully come with more improvements later. Interestingly, the speedup seems to be more significant when `zsh` is specified.

### Tests

3 tests have been modified, all inside `test/init.bats`. 

* **fish instructions**: Now expects `pyenv init - fish | source`
* **setup shell completions** and **setup shell completions (fish)**:  `PYENV_ROOT` is exported before running `pyenv init - <shell>`. 

```
@test "setup shell completions" {
  root="$(cd $BATS_TEST_DIRNAME/.. && pwd)"
  export PYENV_ROOT=$root
  run pyenv-init - bash
  assert_success
  assert_line "source '${root}/completions/pyenv.bash'"
}
```

Assumes that `BATS_TEST_DIRNAME/..` is equal to `PYENV_ROOT`. Enables us to have a simpler path for sourcing of completions, as mentioned in point (3) above.

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -253,52 +253,40 @@ function print_shell_function() {
   commands=(`pyenv-commands --sh`)
   case "$shell" in
   fish )
-    cat <<EOS
-function pyenv
+    echo "function pyenv
   set command \$argv[1]
   set -e argv[1]
 
-  switch "\$command"
+  switch \"\$command\"
   case ${commands[*]}
-    source (pyenv "sh-\$command" \$argv|psub)
+    source (pyenv \"sh-\$command\" \$argv|psub)
   case '*'
-    command pyenv "\$command" \$argv
+    command pyenv \"\$command\" \$argv
   end
-end
-EOS
+end"
     ;;
   ksh | ksh93 | mksh )
-    cat <<EOS
-function pyenv {
-  typeset command
-EOS
+    echo "function pyenv {
+  typeset command=\$1"
     ;;
   * )
-    cat <<EOS
-pyenv() {
-  local command
-EOS
+    echo "pyenv() {
+  local command=\$1"
     ;;
   esac
 
   if [ "$shell" != "fish" ]; then
     IFS="|"
-    cat <<EOS
-  command="\${1:-}"
-  if [ "\$#" -gt 0 ]; then
-    shift
-  fi
-
-  case "\$command" in
+    echo "  [ \"\$#\" -gt 0 ] && shift
+  case \"\$command\" in
   ${commands[*]:-/})
-    eval "\$(pyenv "sh-\$command" "\$@")"
+    eval \"\$(pyenv \"sh-\$command\" \"\$@\")\"
     ;;
   *)
-    command pyenv "\$command" "\$@"
+    command pyenv \"\$command\" \"\$@\"
     ;;
   esac
-}
-EOS
+}"
   fi
 }
 
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -60,8 +60,6 @@ if [ -z "$shell" ]; then
   shell="${shell%%-*}"
 fi
 
-root="${0%/*}/.."
-
 function main() {
   case "$mode" in
   "help")
@@ -244,7 +242,7 @@ function print_env() {
 }
 
 function print_completion() {
-  completion="${root}/completions/pyenv.${shell}"
+  completion="${PYENV_ROOT}/completions/pyenv.${shell}"
   if [ -r "$completion" ]; then
     echo "source '$completion'"
   fi
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -237,7 +237,7 @@ function print_env() {
 }
 
 function print_completion() {
-  completion="${PYENV_ROOT}/completions/pyenv.${shell}"
+  completion="${0%/*/*}/completions/pyenv.${shell}"
   if [ -r "$completion" ]; then
     echo "source '$completion'"
   fi
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -22,35 +22,37 @@ fi
 mode="help"
 no_rehash=""
 no_push_path=""
-for args in "$@"
-do
-  if [ "$args" = "-" ]; then
-    mode="print"
-    shift
-  fi
+shell=""
 
-  if [ "$args" = "--path" ]; then
-    mode="path"
-    shift
-  fi
-
-  if [ "$args" = "--detect-shell" ]; then
-    mode="detect-shell"
-    shift
-  fi
-
-  if [ "$args" = "--no-push-path" ]; then
-    no_push_path=1
-    shift
-  fi
-  
-  if [ "$args" = "--no-rehash" ]; then
-    no_rehash=1
-    shift
-  fi
+for arg in "$@"; do
+  case "$arg" in
+    -)
+      mode="print"
+      ;;
+    zsh|bash|ksh|fish)
+      shell="$arg"
+      ;;
+    --path)
+      mode="path"
+      ;;
+    --detect-shell)
+      mode="detect-shell"
+      ;;
+    --no-push-path)
+      no_push_path=1
+      ;;
+    --no-rehash)
+      no_rehash=1
+      ;;
+    *)
+      echo "Warning: Unknown option '$arg'" >&2
+      echo "Run 'pyenv init --help' for usage." >&2
+      exit 1
+      ;;
+  esac
 done
 
-shell="$1"
+# If shell is not provided, detect it.
 if [ -z "$shell" ]; then
   shell="$(ps -p "$PPID" -o 'args=' 2>/dev/null || true)"
   shell="${shell%% *}"
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -24,13 +24,10 @@ no_rehash=""
 no_push_path=""
 shell=""
 
-for arg in "$@"; do
-  case "$arg" in
+while [ "$#" -gt 0 ]; do
+  case "$1" in
     -)
       mode="print"
-      ;;
-    zsh|bash|ksh|fish)
-      shell="$arg"
       ;;
     --path)
       mode="path"
@@ -45,11 +42,10 @@ for arg in "$@"; do
       no_rehash=1
       ;;
     *)
-      echo "Warning: Unknown option '$arg'" >&2
-      echo "Run 'pyenv init --help' for usage." >&2
-      exit 1
+      shell="$1"
       ;;
   esac
+  shift
 done
 
 # If shell is not provided, detect it.
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -159,7 +159,7 @@ function help_() {
       echo
       echo 'export PYENV_ROOT="$HOME/.pyenv"'
       echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"'
-      echo "eval \"\$(pyenv init - $shell)\""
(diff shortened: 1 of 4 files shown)


libexec/pyenv-init
```diff
@@ -244,7 +237,7 @@ function print_env() {
 }
 
 function print_completion() {
-  completion="${root}/completions/pyenv.${shell}"
+  completion="${PYENV_ROOT}/completions/pyenv.${shell}"
```
native-api, line 240: This must be `$root` -- this is executable root, not data root. The distinction that I was speaking about earlier.

libexec/pyenv-init
```diff
@@ -260,52 +253,40 @@ function print_shell_function() {
   commands=(`pyenv-commands --sh`)
   case "$shell" in
   fish )
-    cat <<EOS
-function pyenv
+    echo "function pyenv
```
native-api, line 256: Can probably use single quotes here and further to save ourselves from having to track what to escape.

test/init.bats
```diff
@@ -35,12 +35,12 @@ create_executable() {
   assert_line "command pyenv rehash 2>/dev/null"
 }
 
-
 @test "setup shell completions" {
   root="$(cd $BATS_TEST_DIRNAME/.. && pwd)"
+  export PYENV_ROOT=$root
```
native-api, line 40: Same as above, do not confuse the two roots.

libexec/pyenv-init
```diff
@@ -166,7 +159,7 @@ function help_() {
       echo
       echo 'export PYENV_ROOT="$HOME/.pyenv"'
       echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"'
-      echo 'eval "$(pyenv init -)"'
+      echo "eval \"\$(pyenv init - $shell)\""
```
native-api, line 162: You can break single quoting here to insert an envvar: `echo 'eval "$(pyenv init - '$shell')"'`

native-api (changes requested):
Great work! An almost twofold speedup just from specifying the shell is definitely a noteworthy discovery!
I guess the rest comes from `case`?
It might be worth specifying in the README, in the advanced shell setup section, that `pyenv init -` does work without an argument but specifying the shell explicitly results in an almost twofold speedup.

libexec/pyenv-init
ChristianFredrikJohnsen: Should be fine to do `'eval "$(pyenv init -'$shell')"'`
Instead of `'eval "$(pyenv init -'"$shell"')"'`

native-api: Yes, if you want the value of `$shell` to be subject to word split since it isn't quoted.
In this case though, it's supposed to be a single word so it doesn't matter, fine either way.

libexec/pyenv-init
ChristianFredrikJohnsen: Rather accidentally when writing documentation in the advanced section, I actually found that pyenv indeed assumes that completions shall be located directly under $PYENV_ROOT, so assuming that completions shall not be under $PYENV_ROOT should not be a problem ? 😇

This is from README.md
```
3. **Installs autocompletion.** This is entirely optional but pretty
   useful. Sourcing `$(pyenv root)/completions/pyenv.bash` will set that
   up. There are also completions for Zsh and Fish.
```

test/init.bats
ChristianFredrikJohnsen: The reason why I am setting this, is because the original PYENV_ROOT (which I guess is determined by the machine which runs the tests) is overridden by `test/test_helper.bash`

```bash
if [ -z "$PYENV_TEST_DIR" ]; then
  PYENV_TEST_DIR="${BATS_TMPDIR}/pyenv"
  export PYENV_TEST_DIR="$(mktemp -d "${PYENV_TEST_DIR}.XXX" 2>/dev/null || echo "$PYENV_TEST_DIR")"

  if enable -f "${BATS_TEST_DIRNAME}"/../libexec/pyenv-realpath.dylib realpath 2>/dev/null; then
    export PYENV_TEST_DIR="$(realpath "$PYENV_TEST_DIR")"
  else
    if [ -n "$PYENV_NATIVE_EXT" ]; then
      echo "pyenv: failed to load \`realpath' builtin" >&2
      exit 1
    fi
  fi

  export PYENV_ROOT="${PYENV_TEST_DIR}/root"
  export HOME="${PYENV_TEST_DIR}/home"
  export PYENV_HOOK_PATH="${PYENV_ROOT}/pyenv.d"

  PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
  PATH="${PYENV_TEST_DIR}/bin:$PATH"
  PATH="${BATS_TEST_DIRNAME}/../libexec:$PATH"
  PATH="${BATS_TEST_DIRNAME}/libexec:$PATH"
  PATH="${PYENV_ROOT}/shims:$PATH"
  export PATH

  for xdg_var in `env 2>/dev/null | grep ^XDG_ | cut -d= -f1`; do unset "$xdg_var"; done
  unset xdg_var
fi
```

On this line:
  export PYENV_ROOT="${PYENV_TEST_DIR}/root"
  
 Where it sets PYENV_ROOT to a non-existing directory.
 It points to something like `/tmp/pyenv.mbz/root`
 I am trying to *undo* this by setting PYENV_ROOT to what it probably was originally.

libexec/pyenv-init
native-api: > Rather accidentally when writing documentation in the advanced section

See? I knew what I was doing when I directed you there! :innocent: 

> so assuming that completions shall not be under $PYENV_ROOT should not be a problem ? 😇

Not a problem at all, they are not:

```
% find $(brew --prefix) -name pyenv.bash
/opt/homebrew-custom/Cellar/pyenv/2.4.22/completions/pyenv.bash
```

So it's your fortunate find that needs fixing!

ChristianFredrikJohnsen: 💀

ChristianFredrikJohnsen: I guess I'll install homebrew just to understand how the folder structure works there.
Of course, we can just revert to:

```
function print_completion() {
  completion="${0%/*}/../completions/pyenv.${shell}"
  if [ -r "$completion" ]; then
    echo "source '$completion'"
  fi
}
```

image
image

native-api: > to understand how the folder structure works there

Basically, whatever is part of the repo is code, whatever is created by the installation is data. `plugins` can be in both places IIRC.

ChristianFredrikJohnsen:
Reverted to using:

```bash
function print_completion() {
  completion="${0%/*/*}/completions/pyenv.${shell}"
  if [ -r "$completion" ]; then
    echo "source '$completion'"
  fi
}
```

Inside `/libexec/pyenv-init`.
I see this as a quick fix, but it does the job.

The major obstacle in making any further progress is the fact that homebrew has a directory setup which is quite bad from my viewpoint. It has core functionality like libexec and completions in this path:

```
/home/linuxbrew/.linuxbrew/Cellar/pyenv/2.4.23/libexec/pyenv-init
```

While shims and python versions are stored in PYENV_ROOT, which is set to `~/.pyenv`. I don't see why you wouldn't want to set 

```bash
PYENV_ROOT=/home/linuxbrew/.linuxbrew/Cellar/pyenv/2.4.23
```

Instead of having it set to `~/.pyenv`, since we now have a situation where **shims** and python **versions** are stored outside of the actual executable directory of pyenv.

Additonally, if you install **pyenv-virtualenv** via homebrew, it is installed in a directory like:

```
/home/linuxbrew/.linuxbrew/Cellar/pyenv-virtualenv/1.2.4
```

instead of

```
/home/linuxbrew/.linuxbrew/Cellar/pyenv/2.4.23/plugins
```

It's very confusing, since even in the README, we have this note:

```
Pyenv plugins
...
The main idea is that most things that you can put under $PYENV_ROOT/<whatever> you can also put under $PYENV_ROOT/plugins/your_plugin_name/<whatever>.
...
```

Which tells us to drop plugins in PYENV_ROOT/plugins, but if you do this with homebrew, you will end up with:

```
/home/linuxbrew/.linuxbrew/Cellar/pyenv/2.4.23/plugins
/home/linuxbrew/.linuxbrew/Cellar/pyenv-virtualenv/1.2.4
~/.pyenv/plugins
```

And your plugins are now all over the place. Of course, it works, but is very confusing to a developer which comes from the setup you get when installing via [pyenv run](https://pyenv.run/)
Crucially though, it doesn't allow us to optimize very aggressively, because scripts can be in all sorts of places.

My conclusion is that it would be very nice to have the `homebrew` installation standardized to the same format as the installation you get from [pyenv run](https://pyenv.run/).

libexec/pyenv-init
```diff
@@ -260,52 +253,40 @@ function print_shell_function() {
...
-EOS
+end'
     ;;
   ksh | ksh93 | mksh )
-    cat <<EOS
-function pyenv {
-  typeset command
-EOS
+    echo 'function pyenv {
+  typeset command=$1'
```
native-api, line 270: Need to retain `$(1:-}` because the resulting shell function runs in user's shell which may have `set -u` active. Then this'll fail with `$1: unbound variable`

libexec/pyenv-init
```diff
@@ -260,52 +253,40 @@ function print_shell_function() {
   commands=(`pyenv-commands --sh`)
   case "$shell" in
   fish )
-    cat <<EOS
-function pyenv
-  set command \$argv[1]
+    echo 'function pyenv
```
native-api, line 256: Might be worth using
```bash
echo \
'function pyenv
```

the indentation won't break as badly and the code will be easier to read.

native-api (commented):
Only some nitpicks remaining. Good job!

native-api (approved):

Issue #3031, opened 2024-08-11
`pip install` behaves differently from `python -m pip install`


bersbersbers:
History:
- https://github.com/astral-sh/uv/issues/2486
- https://github.com/pypa/pip/issues/12909

`python -m pip` does not always install stubs while `pip` does. See https://github.com/pypa/pip/issues/12909 for more details.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): Ubuntu 24.04
- [x] OS architecture (e.g. amd64): WSL, amd64
- [x] pyenv version: 2.4.10
- [x] Python version: 3.12.5
- [x] C Compiler information (e.g. gcc 7.3): (don't know)
- [x] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`.
 
<details>

```
TODO
```

</details>

ChristianFredrikJohnsen:
This is a simple issue to fix.

When pyenv sees that you are writing pip, as in `pip install black`, there is a hook located in:

```
pyenv.d/exec/pip-rehash
```

which essentially does an automatic `pyenv rehash` after installation.

A quick fix would be to add this line:

```bash
if [[ $1 == "python" && $2 == "-m" && $3 == "pip" ]]; then
  PYENV_REHASH_COMMAND="pip"
fi
```

to `pyenv.d/exec/pip-rehash.bash`

native-api:
[...]

PR #3140, opened 2024-12-22
[WIP] fix: add `python -m pip` command to pip hook-script.

ChristianFredrikJohnsen:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3031

### Description

Currently, `pyenv` does NOT do an automatic `rehash` after doing `python -m pip install ...` which is different from the behavior you get when doing `pip install ...`

As was mentioned in #3031, if you do:

`pip install black`
Then you can immediately do:
`black --version`

But if you do:
`python -m pip install black` 

You get something like **command not found**, if you do `black --version`. The following PR will fix this issue.

DETAILED:

When doing `pip install ...`, the command is redirected to
`pyenv.d/exec/pip-rehash/pip`, which does a `pyenv rehash` after
installation. A simple fix to issue #3031 is to make `python -m pip`
commands go through the same special hook script.

If you do `python -m pip install black`, then the main scripts which are read are:

$PYENV_ROOT/shims/python ->
./libexec/pyenv ->
./libexec/pyenv-exec ->
./pyenv.d/exec/pip-rehash/pip

NEEDS TO BE DONE:

Will need to make this work with edge cases like: `python3.12 -m pip install ...` and so forth.

One possible solution, as mentioned by @native-api is to simply check for `-m pip` in the command which is being executed.

The solution should not be too heavy, as the `pyenv.d/exec/pip-rehash.bash` script is sourced every time a python related command is being run (e.g. `python main.py`), and on my computer it already takes ~180ms to run `python --version` :O

### Tests

No tests have been added, but we should DEFINITELY add some tests which check that a rehash is performed when doing `python -m pip install ...` and when doing `pip install ...`, as such tests are not present in the `test` folder.

### DIFF
--- a/pyenv.d/exec/pip-rehash.bash
+++ b/pyenv.d/exec/pip-rehash.bash
@@ -6,9 +6,13 @@ if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.\d)?$ ]]; then
   PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
 fi
 
-if [[ $1 == "python" && $2 == "-m" && $3 == "pip" ]]; then
-  PYENV_REHASH_COMMAND="pip"
-fi
+for (( i=1; i<$#; i++ )); do
+  next=$((i+1))
+  if [[ ${!i} == "-m" && ${!next} == "pip" ]]; then
+    PYENV_REHASH_COMMAND="pip"
+    break
+  fi
+done
 
 if [ -x "${PYENV_PIP_REHASH_ROOT}/${PYENV_REHASH_COMMAND}" ]; then
   PYENV_COMMAND_PATH="${PYENV_PIP_REHASH_ROOT}/${PYENV_REHASH_COMMAND##*/}"
--- a/pyenv.d/exec/pip-rehash.bash
+++ b/pyenv.d/exec/pip-rehash.bash
@@ -6,6 +6,10 @@ if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.\d)?$ ]]; then
   PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
 fi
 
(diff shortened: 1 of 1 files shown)


pyenv.d/exec/pip-rehash.bash
```diff
@@ -6,6 +6,14 @@ if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.\d)?$ ]]; then
   PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
 fi
 
+for (( i=1; i<$#; i++ )); do
+  next=$((i+1))
+  if [[ ${!i} == "-m" && ${!next} == "pip" ]]; then
```
native-api, line 11: I think that just grepping `$*` (which is all arguments as one line) for `\<-m pip\>` would suffice.

ChristianFredrikJohnsen: Using built-in bash regex instead. I am pretty sure that the task is so small (looking for -m pip in a shell command) that launching a grep process is not worth it performance-wise.
It's this fixed cost of setting up a new process which I think we'd rather avoid.

ChristianFredrikJohnsen:
I am pretty sure the implementation is good, will need to find out how to make the tests work on the CI machines. They run fine locally. Only the new tests are failing.

native-api:
> will need to find out how to make the tests work on the CI machines. They run fine locally. Only the new tests are failing.

The tests's job is to check that `pip-rehash` has been onvoked. Rather than try to check its result and rely on certain programs to be present, we can likely override `pyenv-rehash` with a stub that would print something. IIRC there are other tests that add stubs to check how it's done. If not, you can take the `stub` executable and related functions from `python-build` tests.

native-api:
> IIRC there are other tests that add stubs to check how it's done.

see `create_executable`

native-api:
You can comment out everything in CI except one or two things you need. THen in the end revert that commit with "[no ci]" in the message.

PR #3155, opened 2025-01-04
Add standard curl arguments

JayBazuzi:
Most importantly, `-f` instructs curl to fail if the request fails. The other arguments are not strictly necessary, but `-fsSL` is standard boilerplate that many users of curl will recognize.

More details here; https://explainshell.com/explain?cmd=curl+-fsSL+example.org


JayBazuzi:
It would make sense to me if you decide you don't want all of `-fsSL`, for simplicity, and only use `-f` instead. Up to you.

PR #3156, opened 2025-01-06
Support compiler sanitizer options in python-build

ngoldbaum:

### Prerequisite

### Description

Adds a way to enable compiler sanitizers at build time via environment variables in a script, similar to other options.

Also add docs as well as docs for variables that are already in the plugin but aren't documented yet.

### Tests

I tested this locally - it's not clear to me how to add a formal test for this.


native-api:
* What's the use case?
  Looks like this only makes sense for pre-releases.
* This can already be done by setting `PYTHON_CONFIGURE_OPTS` (or `CONFIGURE_OPTS` to also use this for building dependencies). Why add extraneous entities?
* Do both GCC and Clang support these options? If not, this will break the build if the wrong kind of compiler happens to be used.

native-api:
This does not belong in the core Pyenv. We already provide a facility to set arbitrary Configure options.

You can add these options in a plugin via a hook and link to that plugin at https://github.com/pyenv/pyenv/wiki/Plugins .

Moreover, other users with a similar use cases can contribute their own options to the plugin.

native-api:
cc @pyenv/pyenv-core-maintainers

PR #3159, opened 2025-01-07
Fix command substitution in python-build

glephraug:
### Description
- Line 759 of plugins/python-build/bin/python-build seems to be missing the $ in the command subtitution.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -756,7 +756,7 @@ build_package() {
 
   echo "Installing ${package_name}..." >&2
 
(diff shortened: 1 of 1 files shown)


plugins/python-build/bin/python-build
```diff
@@ -756,7 +756,7 @@ build_package() {
 
   echo "Installing ${package_name}..." >&2
 
-  [ -n "$HAS_PATCH" ] && apply_patch "$package_name" <(cat "${package_name}.patch")
```
native-api, line 759: This is [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html)

PR #3173, opened 2025-01-19
Run some pyenv_tests on an ARM processor

cclauss:
Try speeding up some GitHub Action tests by running them on an ARM processor. https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Previous runs: https://github.com/pyenv/pyenv/actions/workflows/pyenv_tests.yml

### Prerequisite

### Description

### Tests

### DIFF
--- a/.github/workflows/pyenv_tests.yml
+++ b/.github/workflows/pyenv_tests.yml
@@ -10,8 +10,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-24.04
-          - ubuntu-22.04
+          - ubuntu-24.04-arm
+          - ubuntu-22.04-arm
           - ubuntu-20.04
           - macos-14
           - macos-13


native-api:
I don't see a difference... about a minute per job either way.
Ubuntu tests are quite fast as they are. In fact, they are the fastest check of all. MacOS jobs are the time hog.
We could use additional jobs on ARM if there's a reason to expect there to be platform-specific issues.

Finally, ARM runners are listed at https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources as "Public preview" (read: alpha). Is it stable enough for production use yet?

cclauss:
Agreed...  Closing as not a good example of a long-running process.

Issue #3075, opened 2024-09-27
Support MacPorts in MacOS


native-api:
Now that MacPorts aren't as niche as they used to be, it's worth supporting them.

Considerations:
* It's probably a bad idea to mix dependencies from different ecosystems -- so it's going to be either Homebrew or MacPorts, whichever is first detected on PATH.

untouchable:
Having pyenv available on macports would be awesome. BTW, thanks for adding it to FreeBSD ports. Anyway, while it doesn't land on macports here's my setup. Maybe we could verify if it works alright for everyone on all supported macOS versions and platforms, and update the docs.

First, let's start by installing all the dependencies as Python 3.13.0 including the latest clang as of now:

`sudo port install -N clang-19 clang_select automake autoconf pkgconfig openssl zlib xz gdbm catsqlite3 sqlite3-tcl tcl tk -z11 +quartz`

Then let's setup our env vars:

`
export PYENV_ROOT="$HOME/.pyenv"
export CC="/opt/local/bin/clang"
export CXX="/opt/local/bin/clang"
export CFLAGS="-O2 -arch $(uname -m) $CFLAGS"
export CPPFLAGS="$CPPFLAGS -I/opt/local/include"
export LDFLAGS="$LDFLAGS -L/opt/local/lib"
export PKG_CONFIG="/opt/local/bin/pkg-config"
export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/opt/local/share/pkgconfig:/opt/local/lib/pkgconfig"
export PKG_CONFIG_LIBDIR="$PKG_CONFIG_LIBDIR:/opt/local/share/pkgconfig:/opt/local/lib/pkgconfig"
export PYTHON_CONFIGURE_OPTS="--enable-framework --with-pkg-config=yes"
export TCLTK_LIBS="-ltk8.6 -ltcl8.6"
`

Lastly, a little hack to make pyenv think homebrew really exists to provide openssl paths; install it somewhere in your $PATH; courtesy of @romank0:

`
#!/bin/sh

# this dir should have include and lib subfolders which have openssl headers and libraries
OPENSSL_PATH=/opt/local/libexec/openssl3
LIBRARY_PATH=/opt/local/lib

if [ "$1 $2" == "--prefix openssl" ]; then
    echo $OPENSSL_PATH
elif [ "$1 $2" == "--prefix readline" ]; then
    echo $LIBRARY_PATH
else
[...]

PR #3186, opened 2025-02-09
Add MacPorts support for python-build

studnitskiy:
This commit allows building Python on macOS without Homebrew,
with the following libraries installed via MacPorts:
 - ncurses
 - openssl
 - readline
 - zlib (XCode SDK zlib is still tried first)

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3075

### Description

If either Homebrew or MacPorts system is the only one installed, that system
will be used automatically.
When both systems are installed, then whatever is found first in PATH - brew
or port - will be used for all dependencies.
The following environment variables can be set to non-empty value to override
the selection:
PYTHON_BUILD_USE_HOMEBREW
PYTHON_BUILD_USE_MACPORTS

Tcl/Tk specific support is omitted due to CPython incompatibility with
Tcl/Tk 9, according to the comments in use_homebrew_tcltk().
Those who need Tcl/Tk module, can use PYTHON_CONFIGURE_OPTS --with-tcltk-libs=
pointing to the Tcl/Tk location which can be installed either from sources or
via MacPorts.

### Tests

Homebrew is used when brew is before port in PATH
MacPorts is used over Homebrew when both present and port is first in PATH
Non-empty PYTHON_BUILD_USE_HOMEBREW overrides PATH order selection
Non-empty PYTHON_BUILD_USE_MACPORTS overrides PATH order selection
Tests for added use_macports* functions are based off of the
corresponding brew counterparts.

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -128,10 +128,36 @@ can_use_homebrew() {
   }
   [[ -n "$PYTHON_BUILD_USE_HOMEBREW" ]] && return 0
   [[ -n "$PYTHON_BUILD_SKIP_HOMEBREW" ]] && return 1
-  is_mac && return 0
+  is_mac && command -v brew &>/dev/null && return 0
   # In Linux, if Pyenv itself is installed with Homebrew,
   # we assume the user wants to take dependencies from there as well by default
   command -v brew &>/dev/null && [[ $(abs_dirname "${BASH_SOURCE}") == "$(abs_dirname "$(brew --prefix 2>/dev/null ||true)")"/* ]] && return 0
+  return 1
+}
+
+can_use_macports() {
+  [[ -n "$PYTHON_BUILD_USE_MACPORTS" && -n "$PYTHON_BUILD_SKIP_MACPORTS" ]] && {
+    echo "error: mutually exclusive environment variables PYTHON_BUILD_USE_MACPORTS and PYTHON_BUILD_SKIP_MACPORTS are set" >&3
+    exit 1
+  }
+  [[ -n "$PYTHON_BUILD_USE_MACPORTS" ]] && return 0
+  [[ -n "$PYTHON_BUILD_SKIP_MACPORTS" ]] && return 1
+  command -v port &>/dev/null && return 0
+  return 1
+}
+
+is_homebrew_preferred() {
+  can_use_homebrew || return 1
+  [ -n "$PYTHON_BUILD_USE_HOMEBREW" ] && return 0
+  local brew_path port_path path paths
+  command -v port &>/dev/null && port_path=$(dirname "$(command -v port)")
+  [ -z "$port_path" ] && return 0
+  [ -n "$PYTHON_BUILD_USE_MACPORTS" ] && return 1
+  command -v brew &>/dev/null && brew_path=$(dirname "$(command -v brew)")
+  [ -z "$brew_path" ] && return 1
+  # Homebrew and MacPorts found in PATH
+  # find out what to use based on the PATH order
+  [[ $PATH == *$brew_path*":"*$port_path* ]] && return 0
   return 1
 }
 
@@ -815,14 +841,28 @@ build_package_standard_build() {
   local PACKAGE_LDFLAGS="${package_var_name}_LDFLAGS"
 
   if [ "$package_var_name" = "PYTHON" ]; then
-    use_homebrew || true
-    use_custom_tcltk || use_homebrew_tcltk || true
-    use_homebrew_readline || use_freebsd_pkg || true
-    use_homebrew_ncurses || true
-    if is_mac -ge 1014; then
-      use_xcode_sdk_zlib || use_homebrew_zlib || true
+    if is_homebrew_preferred; then
+      use_homebrew || true
+      use_custom_tcltk || use_homebrew_tcltk || true
+      use_homebrew_readline || true
+      use_homebrew_ncurses || true
+      if is_mac -ge 1014; then
+        use_xcode_sdk_zlib || use_homebrew_zlib || true
+      else
+        use_homebrew_zlib || true
+      fi
+    elif can_use_macports; then
+      use_macports || true
+      use_custom_tcltk || true
+      use_macports_readline || true
+      use_macports_ncurses || true
+      if is_mac -ge 1014; then
+        use_xcode_sdk_zlib || use_macports_zlib || true
+      else
+        use_macports_zlib || true
+      fi
     else
-      use_homebrew_zlib || true
+      use_freebsd_pkg || true
     fi
     use_dsymutil || true
     use_free_threading || true
@@ -1438,9 +1478,24 @@ use_homebrew() {
   fi
 }
 
+use_macports() {
+  can_use_macports || return 1
+  local port_location="$(command -v port)"
+  if [ -n "$port_location" ]; then
+    local prefix="${port_location%/bin/port}"
+    export CPPFLAGS="${CPPFLAGS:+$CPPFLAGS }-I${prefix}/include"
+    append_ldflags_libs "-L${prefix}/lib -Wl,-rpath,${prefix}/lib"
+  fi
+}
+
 needs_yaml() {
-  ! configured_with_package_dir "python" "yaml.h" &&
-  ! use_homebrew_yaml
+  if ! configured_with_package_dir "python" "yaml.h"; then
+    if is_homebrew_preferred; then
+      use_homebrew_yaml && return 1
+    elif can_use_macports; then
+      use_macports_yaml && return 1
+    fi
+  fi
 }
 
 use_homebrew_yaml() {
@@ -1450,6 +1505,21 @@ use_homebrew_yaml() {
     echo "python-build: use libyaml from homebrew"
     export CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"
     export LDFLAGS="-L$libdir/lib${LDFLAGS:+ ${LDFLAGS% }}"
+  else
+    return 1
+  fi
+}
+
+use_macports_yaml() {
+  can_use_macports || return 1
+  local prefix="$(port -q location libyaml 2>/dev/null || true)"
+  if [ -n "$prefix" ]; then
+    local libdir="$prefix/opt/local"
+    if [ -d "$libdir" ]; then
+      echo "python-build: use libyaml from MacPorts"
+      export CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"
+      export LDFLAGS="-L$libdir/lib${LDFLAGS:+ ${LDFLAGS% }}"
+    fi
   else
     return 1
   fi
@@ -1488,9 +1558,13 @@ use_freebsd_pkg() {
 has_broken_mac_readline() {
   # Mac OS X 10.4 has broken readline.
   # https://github.com/pyenv/pyenv/issues/23
-  is_mac &&
-  ! configured_with_package_dir "python" "readline/rlconf.h" &&
-  ! use_homebrew_readline
+  if is_mac && ! configured_with_package_dir "python" "readline/rlconf.h"; then
+    if is_homebrew_preferred; then
+      use_homebrew_readline && return 1
+    elif can_use_macports; then
+      use_macports_readline && return 1
+    fi
+  fi
 }
 
 use_homebrew_readline() {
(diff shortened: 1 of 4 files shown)


plugins/python-build/bin/python-build
```diff
@@ -817,12 +817,12 @@ build_package_standard_build() {
   if [ "$package_var_name" = "PYTHON" ]; then
     use_homebrew || true
```
native-api, line 818: We should probably avoid using using both ecosystems at once since that'll likely lead to crashes from binary incompatibilities between their libraries.

So need to somehow mark which one we've selected and avoid using the other.

Also like with Homebrew, likely need user settings like envvars to let the user override the choice.

plugins/python-build/bin/python-build
```diff
@@ -1455,6 +1455,17 @@ use_homebrew_yaml() {
   fi
 }
 
+use_macports_yaml() {
+  local prefix="/opt/local"
```
native-api, line 1459: Is the prefix "`/opt/local`" exclusively? What about the dual architecture Macs -- which architecture does it use?

plugins/python-build/bin/python-build
```diff
@@ -1519,6 +1530,30 @@ use_homebrew_ncurses() {
   fi
 }
 
+use_macports_readline() {
+  if ! configured_with_package_dir "python" "readline/rlconf.h"; then
+    local prefix="/opt/local"
+    if [ -d "$prefix/include/readline" ]; then
+      echo "python-build: use readline from MacPorts"
+      export CPPFLAGS="-I$prefix/include/readline ${CPPFLAGS}"
```
native-api, line 1538: Should probably concatenate uniformly everywhere like `CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"` to avoid any surprise dangling/duplicate spaces.

studnitskiy:
This was mostly a hackjob, I re-wrote based on the review comments, will try another PR

plugins/python-build/bin/python-build
studnitskiy: Changed to conform with other code

plugins/python-build/bin/python-build
studnitskiy: Normally, when installing using the installer, the location is /opt/local
Non-default location is possible when building and installing MacPorts from the source code.
I added prefix handling, it should be dynamically resolved from the port command location.

plugins/python-build/bin/python-build
studnitskiy: I added environment variables PYTHON_BUILD_USE_MACPORTS, PYTHON_BUILD_SKIP_MACPORTS, these should work similarly to Homebrew counterparts. Also added some code to handle variable mismanagement.
If none variables are set, then Homebrew is used first as it's more widely installed.
If only MacPorts is installed on the host, there will be no need for special configuration though, it should be found and used.

plugins/python-build/test/build.bats
```diff
@@ -617,6 +705,33 @@ make -j 2
 make install
 OUT
 }
+
+@test "MacPorts is not touched if PYTHON_BUILD_SKIP_MACPORTS is set" {
```
native-api, line 709: also add a case for the new mutually exclusive envvars

native-api (commented):
I don't see any logic preventing the use of the other ecosystem when one is being used.

native-api:
See my comments in https://github.com/pyenv/pyenv-virtualenv/pull/502#pullrequestreview-2615988289 for diagnosing stuff in tests

plugins/python-build/bin/python-build
```diff
@@ -126,6 +126,11 @@ can_use_homebrew() {
     echo "error: mutually exclusive environment variables PYTHON_BUILD_USE_HOMEBREW and PYTHON_BUILD_SKIP_HOMEBREW are set" >&3
     exit 1
   }
+  [[ -n "$PYTHON_BUILD_USE_HOMEBREW" && -n "$PYTHON_BUILD_USE_MACPORTS" ]] && {
```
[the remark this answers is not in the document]
studnitskiy: This is a failsafe to prevent both systems to be configured as allowed. It should fail if PYTHON_BUILD_USE_HOMEBREW and PYTHON_BUILD_USE_MACPORTS both non-empty, similar to the previous condition
[[ -n "$PYTHON_BUILD_USE_HOMEBREW" && -n "$PYTHON_BUILD_SKIP_HOMEBREW" ]]

studnitskiy:
Thank you for taking a look!

> I don't see any logic preventing the use of the other ecosystem when one is being used.
Yes, that was intentional.
On systems where Homebrew is present, it is attempted by default unless PYTHON_BUILD_SKIP_HOMEBREW is set.
I thought this is a safe bet. However, I guess there might be a situation when, say, ncurses is installed with brew, and maybe readline is installed with ports, and then both systems would be used.
I'll make a change for either to be used exclusively, probably with brew to still be the default, as it is the bigger user base by far.

plugins/python-build/test/build.bats
studnitskiy: I added a couple of tests
- install aborted when PYTHON_BUILD_USE_HOMEBREW and PYTHON_BUILD_USE_MACPORTS both set
- MacPorts is not used in non-MacOS (setting PYTHON_BUILD_USE_MACPORTS should not make a difference in Linux, similar to PYTHON_BUILD_CONFIGURE_WITH_DSYMUTIL)

plugins/python-build/bin/python-build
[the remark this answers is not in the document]
native-api: https://github.com/pyenv/pyenv/blob/47eef10ce172125f2c8761f08f1efa2385042897/plugins/python-build/README.md?plain=1#L165

The envvar overrides the default choice, not forces the use.
It's not illegal to enable both systems, in fact, it is enabled by default in MacOS. The first one discovered will be used.

plugins/python-build/test/build.bats
native-api: > * MacPorts is not used in non-MacOS

Is there a particular reason? AFAIK BSD ports are available for nearly everything.
While we may not officially support such a scenario, why add extra code to make the product worse?

`PYTHON_BUILD_CONFIGURE_WITH_DSYMUTIL` is different. It adds a CPython configure option which, by Configure's logic, is invalid outside of MacOS. Leaving it in would cause a Configure error.

jacob-lee:
There may be few mac ports users off of Mac, but it is designed to be
platform independent:

https://trac.macports.org/wiki/InstallingMacPortsOnUbuntuLinux

plugins/python-build/test/build.bats
studnitskiy: > Is there a particular reason? AFAIK BSD ports are available for nearly everything.

MacPorts technically can be used in Linux, although they say the support is experimental and installation requires building from source
https://trac.macports.org/wiki/InstallingMacPortsOnUbuntuLinux

Furthermore, in the install guide on their main page at https://www.macports.org/install.php#installing they say this

> Running on platforms other than macOS is not the main focus of The MacPorts Project, so remaining cross-platform is not an actively-pursued development goal.


> While we may not officially support such a scenario, why add extra code to make the product worse? Just make it off by default.


Ok, I'll make a change so that if PYTHON_BUILD_USE_MACPORTS is set, it will be enabled regardless of the platform.

plugins/python-build/bin/python-build
```diff
@@ -128,13 +128,57 @@ can_use_homebrew() {
...
+  [ -z "$port_path" ] && return 1
+  IFS=: paths=($PATH)
+  IFS="$OLDIFS"
+  for path in "${paths[@]}"; do
+    [[ "$path" == "$port_path" ]] && return 0
+    [[ "$path" == "$brew_path" ]] && return 1
+  done
+  return 1
+}
+
```
native-api, line 181: So, you are suggesting to prefer whichever has its PATH entry earlier? This looks counteruntuitive: ususally, PATH order doesn't matter for different executables, it only matters if it's on PATH at all.

This also looks overcomplicated.
The idea I had in mind was to simply set an envvar prohibiting the use of the other ecosystem after we've selected a library from one.
While during initial selection, we would prefer Homebrew as the more prevalent ecosystem, by simply trying it first.

I'm ready to make a child PR if that'll be easier that trying to understand me

studnitskiy: Hi @native-api 
>So, you are suggesting to prefer whichever has its PATH entry earlier?

Yeah, I thought that's what you meant when you said earlier
>The first one allowed and discovered is used.

My idea was that if we have both systems installed, and there's no clear indication which system to use (e.g. PYTHON_BUILD_SKIP_ * env var), PATH order would be the way to discover which one is "preferred", i.e. the first to use.  I agree that the order does not matter for different execs, but it does matter for the same name execs, and if you had, say, openssl installed via both brew and port, then you'd need brew's bin to be first in the PATH if using brew-installed openssl is preferred.


>The idea I had in mind was to simply set an envvar prohibiting the use of the other ecosystem after we've selected a library from one.
>While during initial selection, we would prefer Homebrew as the more prevalent ecosystem, by simply trying it first.

OK, I'll have another go at it with this approach

native-api: > if you had, say, openssl installed via both brew and port, then you'd need brew's bin to be first in the PATH if using brew-installed openssl is preferred.

That's a good observation! If MacPorts and Homebrew's executable will never be in the same PATH entry, this does look like a way to guess the user's preference!

The code still could be much simpler.
E.g. `is_homebrew_preferred` and `is_macports_preferred` duplicate each other, only one is needed.
It also looks unnecessary to loop over PATH entries: find one tool via `command -v`, get its `PATH` entry via `dirname`, then check if you can `command -v` the other tool with a `PATH` cut to _before_ that entry. If you can, it's earlier; if you cannot, it's either later or absent.
My above proposition about simplification still stands, just the picking method has changed.

We should probably ask input from others subscribed to the issue: once we establish a public interface, changing it would be a breaking change.

native-api (commented):
Finally found time to look into this using part of a day off.
My initial idea was to write a nested PR cleaning up the duplications. But I failed to work out a clear structure. So, we'll just let the code evolve naturally through refactoring.

In fact, refactoring would probably be needed because the current logic is rather inconsistent and has many duplicate parts.

* USE/SKIP envvars as they are currently documented now do not override priority as the current code does. They are only supposed to override the default permission to use. As it currently is, we do not have a setting to override priority.
  * User feedback in the issue suggests that a separate envvar for priority would be too much. The only explicit statements suggest to give Homebrew higher priority, and to lock into an ecosystem if Pyenv itself is installed with it.
  I do not have a clear preference of whether to give Homebrew priority or use the PATH order. Ready to try either and see user feedback.

* Once we pick a library, we need to lock into the corresponding ecosystem -- so that we don't pick libs from multiple ones.
As a corollary, we need to `use_homebrew`/`use_macports` after trying specific libraries.

studnitskiy:
>USE/SKIP envvars as they are currently documented now do not override priority as the current code does. They are only supposed to override the default permission to use. As it currently is, we do not have a setting to override priority.

PYTHON_BUILD_USE/SKIP_MACPORTS mirror functionality of *_HOMEBREW env vars that was present.
Having two systems installed on the same host is a corner case. I think in majority of the cases there will be only one of the systems present. In rare cases when both systems are installed, one of them will be selected automatically, and likely correctly, based on PATH. But if that not preferred, then any USE/SKIP variables can be (re)used to explicitly set the preference.

>Once we pick a library, we need to lock into the corresponding ecosystem -- so that we don't pick libs from multiple ones.

if/elif/else block in build_package_standard_build() currently takes care of that, although it just picks one of the systems rather than looking for a dependency library first, and then locking to a system that found library is installed with. I think this implementation of picking a system first should be easier than backtracking from a found library, and there should not be a difference in the outcome - the build will use libraries from one of the systems, and none from the other.

native-api:
> if/elif/else block in `build_package_standard_build()` currently takes care of that

Not quite. Readline and OpenSSL are also picked from `has_broken_mac_*`

studnitskiy:
Yes, I forgot about those. There are similar if/elif blocks based on the `is_homebrew_preferred` check in each of the `has_broken_mac_*` functions. Since there's no expectation of the preference changing while python-build is running, I guess that'll do. If not, let me know what you think.

native-api:
@studnitskiy Please register at https://opencollective.com/, set up a money transfer method in you profile and write me your OpenCollective handle name on the e-mail in my profile here -- so that I can set up the payout.

Issue #3241, opened 2025-04-29
Pyston install fails because pip dropped support for Python 3.8


dmrlawson:
e.g. 

```
$ pyenv install pyston-2.3.5
Downloading pyston_2.3.5_portable_amd64.tar.gz...
-> https://github.com/pyston/pyston/releases/download/pyston_2.3.5/pyston_2.3.5_portable_amd64.tar.gz
Installing pyston_2.3.5_portable_amd64...
Installing pip from https://bootstrap.pypa.io/get-pip.py...
error: failed to install pip via get-pip.py

BUILD FAILED (Ubuntu 24.04 using python-build 20180424)

Inspect or clean up the working tree at /tmp/python-build.20250429101813.96636
Results logged to /tmp/python-build.20250429101813.96636.log

Last 10 log lines:
x-served-by: cache-lon420145-LON
x-cache: HIT
x-cache-hits: 0
x-timer: S1745921894.028071,VS0,VE1
vary: Accept-Encoding
x-fastly-request-id: 533ab98fd69a95af8e2d4c6d93717f8d44294c9c
content-length: 9379

/tmp/python-build.20250429101813.96636/pyston_2.3.5_portable_amd64 /tmp/python-build.20250429101813.96636 ~
ERROR: This script does not work on Python 3.8. The minimum supported Python version is 3.9. Please use https://bootstrap.pypa.io/pip/3.8/get-pip.py instead.
```

I think we need something like https://github.com/pyenv/pyenv/pull/2788 to add the 3.8 specific URL.

PR #3242, opened 2025-04-29
Fix get-pip urls for pyston

dmrlawson:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3241

### Description
Fixes the get-pip url for pyston versions.

We might want to do the same for all python 3.8.* (and pypy3.8?) but I haven't tested them.

### Tests


native-api:
I see no reason why it wouldn't work for `3.8*`. It's the official Pip script specifically for those versions.

Issue #3253, opened 2025-05-20
python-build can't seem to find readline after the new MacPorts support


fofoni:
Using commit 5e82f5cf (just before https://github.com/pyenv/pyenv/pull/3186), `pyenv install 3.13.2` works fine in my system (Ubuntu).

Using f24ff8d1 (the merge commit for https://github.com/pyenv/pyenv/pull/3186, and the most recent commit as of right now), I get the following behavior from `pyenv install -v 3.13.2`.

It starts by _building_ readline (which I've never seen pyenv do before) and installing it to the local pyenv version path:
```
$ pyenv install -v 3.13.2
/tmp/python-build.20250520170019.512445 ~
/tmp/python-build.20250520170019.512445/readline-8.2 /tmp/python-build.20250520170019.512445 ~
Installing readline-8.2...
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu

Beginning configuration for readline-8.2 for x86_64-pc-linux-gnu

...

Installed readline-8.2 to /home/pedro/.local/lib/pyenv/versions/3.13.2
```

It then goes on to build Python itself, but it can't find readline:

```
/tmp/python-build.20250520170019.512445 ~
/tmp/python-build.20250520170019.512445/Python-3.13.2 /tmp/python-build.20250520170019.512445 ~
Installing Python-3.13.2...
...
checking for readline... no
checking for readline/readline.h... yes
checking for readline in -lreadline... no
checking how to link readline... no
```

Finally, it does finish building Python, but without the `readline` Python module:
```
ModuleNotFoundError: No module named 'readline'
```

I don't have brew or linuxbrew installed:
```
$ type brew
-bash: type: brew: not found
$ type linuxbrew
-bash: type: linuxbrew: not found
$
```

PR #3254, opened 2025-05-20
Fix python-build's `has_broken_mac_readline`

fofoni:

### Prerequisite

> * [ ] Please consider implementing the feature as a hook script or plugin as a first step.
>   * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins](https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a plugin if possible.

Not a feature.

> * [ ] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
>   * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv will be imported to pyenv too, eventually.
>   * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with rbenv.

Not a feature.

  - Closes https://github.com/pyenv/pyenv/issues/3253

### Description

The `has_broken_mac_openssl` python-build function, introduced in #3186, is careful enough to start with `is_mac || return 1` in order not to affect non-mac systems. That PR however didn't do so in `has_broken_mac_readline`.

### Tests

Should I add any?

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1585,7 +1585,8 @@ use_freebsd_pkg() {
 has_broken_mac_readline() {
   # Mac OS X 10.4 has broken readline.
   # https://github.com/pyenv/pyenv/issues/23
-  if is_mac && ! configured_with_package_dir "python" "readline/rlconf.h"; then
+  is_mac || return 1
+  if ! configured_with_package_dir "python" "readline/rlconf.h"; then
     if can_use_homebrew; then
       use_homebrew_readline && return 1
     fi


fofoni:
Actually, the call to `configured_with_package_dir` was also refactored to inside the `if` condition, which therefore also looks like a mistake.

I'll also revert it (but please tell me if this was a correct change in the first place).

plugins/python-build/bin/python-build
```diff
@@ -1586,13 +1586,12 @@ has_broken_mac_readline() {
...
   # https://github.com/pyenv/pyenv/issues/23
   is_mac || return 1
-  if ! configured_with_package_dir "python" "readline/rlconf.h"; then
-    if can_use_homebrew; then
-      use_homebrew_readline && return 1
-    fi
-    if can_use_macports; then
-      use_macports_readline && return 1
-    fi
+  ! configured_with_package_dir "python" "readline/rlconf.h" || return 1
```
native-api, line 1589: Double negation is confusing. Could you transpose it?

native-api (approved):

plugins/python-build/bin/python-build
fofoni: The idea was to reuse the `|| return 1` idiom to make it clear in the context of the above line. I feel that this:

```diff
   # Mac OS X 10.4 has broken readline.
   # https://github.com/pyenv/pyenv/issues/23
   is_mac || return 1
-  ! configured_with_package_dir "python" "readline/rlconf.h" || return 1
+  configured_with_package_dir "python" "readline/rlconf.h" && return 1
   if can_use_homebrew; then
     use_homebrew_readline && return 1
   fi
```

would be even more confusing.

How about making these first guards more explicit, like this:

```diff
 has_broken_mac_readline() {
   # Mac OS X 10.4 has broken readline.
   # https://github.com/pyenv/pyenv/issues/23
-  is_mac || return 1
-  ! configured_with_package_dir "python" "readline/rlconf.h" || return 1
+  if ! is_mac || configured_with_package_dir "python" "readline/rlconf.h"; then
+    # Not applicable.
+    return 1
+  fi
   if can_use_homebrew; then
     use_homebrew_readline && return 1
   fi
```

Which do you prefer?

fofoni: I'll go with the second (more explicit) version for now. I can change to the first (`&& return 1`) if you think it's fine.

Issue #3256, opened 2025-05-23
Add defensive check for unset $PYENV_ROOT in pyenv-init


acoalson:
While reviewing the [libexec/pyenv-init](https://github.com/pyenv/pyenv/blob/master/libexec/pyenv-init) script, I noticed that the init_dirs function assumes that $PYENV_ROOT is set, without checking whether it's empty or unset.

`function init_dirs() {
  mkdir -p "${PYENV_ROOT}/"{shims,versions}
}`

If $PYENV_ROOT is not set this could either create directories at the root of the filesystem or fail silently, leading to hard-to-debug errors later in the session.

Fix: Add a defensive check before the mkdir command and exit with an informative error message if $PYENV_ROOT is empty

native-api:
See the comment in the PR

PR #3257, opened 2025-05-24
Add defensive check for unset PYENV_ROOT in init_dirs

acoalson:
### Prerequisite
  *  Not suitable as a plugin because `init_dirs` is part of `pyenv-init`, which is core initialization logic run directly by `pyenv init`. The check must occur within the core script to be effective.
  * This logic is pyenv-specific. rbenv does not use `PYENV_ROOT`, and there is no equivalent `init_dirs` function. Upstreaming would not be applicable.
  - Closes Issue [#3256](https://github.com/pyenv/pyenv/issues/3256)
 
### Description
This PR adds a defensive check to the `init_dirs` function in `libexec/pyenv-init` to ensure that the `PYENV_ROOT` environment variable is set before directory creation.  Without this check, the script may create directories like /shims and /versions if $PYENV_ROOT is unset.

Currently, if `PYENV_ROOT` is unset, the following command silently creates directories in the root of the filesystem:

```bash
mkdir -p "${PYENV_ROOT}/"{shims,versions}
```

This update adds:
```bash
if [ -z "$PYENV_ROOT" ]; then
  echo "Error: PYENV_ROOT is not set. Aborting." >&2
  exit 1
fi
```

Improvements:
- Prevents silent failure or root-level directory creation
- Improves user experience and error clarity
- Maintains Bash 3.2 compatibility

### Tests
   - No tests have been added

### DIFF
--- a/libexec/pyenv-init
+++ b/libexec/pyenv-init
@@ -169,6 +169,11 @@ function help_() {
 }
 
 function init_dirs() {
+  if [ -z "$PYENV_ROOT" ]; then
+    echo 'Error: PYENV_ROOT is not set. Aborting.' >&2
+    exit 1
+  fi
+
   mkdir -p "${PYENV_ROOT}/"{shims,versions}
 }
 


native-api (commented):
> to ensure that the PYENV_ROOT environment variable is set

This variable is set by the `pyenv` launcher. All the subcommands are supposed to be run through it -- that's why `libexec` is not on `PATH`.

Issue #3229, opened 2025-04-03
pyenv stuck on "python-build: use zlib from xcode sdk"


micheledemeo:
- MAC OS Sequoia Version 15.3.2 (24D81)
- pyenv 2.5.4

pyenv is very slow and remains stuck for every python version:

![Image](https://github.com/user-attachments/assets/cc30b59a-1a7f-4fea-a6e5-e63b1d01b840)

Any idea?

native-api:
Please add the necessary diagnostic details as per thge issue template.

revatir:
I'm also seeing the same issue after running `pyenv install --skip-existing 3.11.11`. I also don't see an issue template :\

```
python-build: use openssl@3 from homebrew
python-build: use readline from homebrew
Downloading Python-3.11.11.tar.xz...
-> https://www.python.org/ftp/python/3.11.11/Python-3.11.11.tar.xz
Installing Python-3.11.11...
python-build: use readline from homebrew
python-build: use ncurses from homebrew
python-build: use zlib from xcode sdk
```

gebeuhler:
Any fix for this? Can't install new versions of python using pyenv. Using Mac OS 15.5

native-api:
@gebeuhler As per above, need diagnostic information as per the [issue template](https://github.com/pyenv/pyenv/blob/master/.github/ISSUE_TEMPLATE.md) to say anything.

gebeuhler:
> [@gebeuhler](https://github.com/gebeuhler) As per above, need diagnostic information as per the [issue template](https://github.com/pyenv/pyenv/blob/master/.github/ISSUE_TEMPLATE.md) to say anything.

I did reinstall `homebrew` and `pyenv` before attempting these steps. it did not fix the issue.

Platform information (e.g. Ubuntu Linux 16.04): Mac OSX 15.5
 OS architecture (e.g. amd64): apple m1 pro
 pyenv version: 2.6.2
 Python version: 3.11
 C Compiler information (e.g. gcc 7.3):
`$ gcc --version
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin`

@native-api attaching gist of `env PYENV_DEBUG=1 pyenv install -v 3.11 2>&1 | tee trace.log`

https://gist.github.com/gebeuhler/124178d921730ff84b3717cc7aff1da2

[...]

PR #3272, opened 2025-06-13
Support linking to MacPorts ports installed from binary archives

0916dhkim:

### Prerequisite
  - Related to https://github.com/pyenv/pyenv/issues/3075

### Description

Macports support added from https://github.com/pyenv/pyenv/pull/3186 was not working on my machine. `python-build` was not picking up the correct `CPPFLAGS` and `LDFLAGS`.
I tried to modify the script to match what works on my setup (M1 macbook), so let me know if there's a concern about compatibility.

### Tests

### DIFF
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -1538,13 +1538,14 @@ use_homebrew_yaml() {
 
 use_macports_yaml() {
   can_use_macports || return 1
-  local prefix="$(port -q location libyaml 2>/dev/null || true)"
+  local port_location="$(command -v port)"
+  local prefix="${port_location%/bin/port}"
   if [ -n "$prefix" ]; then
-    local libdir="$prefix/opt/local"
-    if [ -d "$libdir" ]; then
+    local installation="$(port -q list libyaml)"
+    if [ -n "$installation" ]; then
       echo "python-build: use libyaml from MacPorts"
-      export CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"
-      export LDFLAGS="-L$libdir/lib${LDFLAGS:+ ${LDFLAGS% }}"
+      export CPPFLAGS="-I$prefix/include${CPPFLAGS:+ $CPPFLAGS}"
+      export LDFLAGS="-L$prefix/lib${LDFLAGS:+ ${LDFLAGS% }}"
       lock_in macports
     fi
   else
@@ -1615,13 +1616,14 @@ use_homebrew_readline() {
 use_macports_readline() {
   can_use_macports || return 1
   if ! configured_with_package_dir "python" "readline/rlconf.h"; then
-    local prefix="$(port -q location readline 2>/dev/null || true)"
+    local port_location="$(command -v port)"
+    local prefix="${port_location%/bin/port}"
     if [ -n "$prefix" ]; then
-      local libdir="$prefix/opt/local"
-      if [ -d "$libdir" ]; then
+      local installation="$(port -q list readline)"
+      if [ -n "$installation" ]; then
         echo "python-build: use readline from MacPorts"
-        export CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"
-        export LDFLAGS="-L$libdir/lib${LDFLAGS:+ $LDFLAGS}"
+        export CPPFLAGS="-I$prefix/include${CPPFLAGS:+ $CPPFLAGS}"
+        export LDFLAGS="-L$prefix/lib${LDFLAGS:+ $LDFLAGS}"
         lock_in macports
       fi
     else
@@ -1645,13 +1647,14 @@ use_homebrew_ncurses() {
 
 use_macports_ncurses() {
   can_use_macports || return 1
-  local prefix="$(port -q location ncurses 2>/dev/null || true)"
-  if [[ -n "$prefix" ]]; then
-    local libdir="$prefix/opt/local"
-    if [ -d "$libdir" ]; then
+  local port_location="$(command -v port)"
+  local prefix="${port_location%/bin/port}"
+  if [ -n "$prefix" ]; then
+    local installation="$(port -q list ncurses)"
+    if [ -n "$installation" ]; then
       echo "python-build: use ncurses from MacPorts"
-      export CPPFLAGS="-I$libdir/include${CPPFLAGS:+ $CPPFLAGS}"
-      export LDFLAGS="-L$libdir/lib${LDFLAGS:+ $LDFLAGS}"
+      export CPPFLAGS="-I$prefix/include${CPPFLAGS:+ $CPPFLAGS}"
+      export LDFLAGS="-L$prefix/lib${LDFLAGS:+ $LDFLAGS}"
       lock_in macports
     fi
   else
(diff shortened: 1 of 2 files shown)


plugins/python-build/bin/python-build
```diff
@@ -1538,13 +1538,14 @@ use_homebrew_yaml() {
 
 use_macports_yaml() {
   can_use_macports || return 1
-  local prefix="$(port -q location libyaml 2>/dev/null || true)"
+  local port_location="$(command -v port)"
+  local prefix="${port_location%/bin/port}"
   if [ -n "$prefix" ]; then
-    local libdir="$prefix/opt/local"
-    if [ -d "$libdir" ]; then
```
0916dhkim, line 1544: The original code `port -q location libyaml` produces `/opt/local/var/macports/software/libyaml/libyaml-0.2.5_0.darwin_24.arm64` on my machine, so `libdir` becomes `/opt/local/var/macports/software/libyaml/libyaml-0.2.5_0.darwin_24.arm64/opt/local`.

I've changed the script to detect the macport prefix, add `$prefix/include` (`/opt/local/include` on my machine) to `CPPFLAGS`, and add `$prefix/lib` to `LDFLAGS`.

plugins/python-build/bin/python-build
```diff
@@ -1538,13 +1538,14 @@ use_homebrew_yaml() {
...
+    local installation="$(port -q list libyaml)"
```
0916dhkim, line 1544: This line is for checking if the library is installed.
`installation` is probably not the best name. Please suggest a better one.

native-api:
On my machine, the existing command works as expected:

```
% port -q location readline  
/opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64
% tree -d /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64
/opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64
└── opt
    └── local
        ├── include
        │   └── readline
        ├── lib
        │   └── pkgconfig
        └── share
            ├── doc
            │   └── readline
            │       └── html
            ├── info
            ├── man
            │   └── man3
            └── readline
% port version 
Version: 2.10.7
```

native-api (changes requested):
AFAICS, MacPorts, like Homebrew, has a subtree for each package ("port"), and a separate set of files under prefix for "activated" ports -- but unlike Homebrew, it seems to plain copy files rather an symlink:

```
% find /opt/local -name libreadline\* -exec ls -ld {} \+
-rwxr-xr-x  1 root  wheel  275080 Aug 14  2024 /opt/local/lib/libreadline.8.2.dylib
lrwxr-xr-x  1 root  wheel      21 Aug 14  2024 /opt/local/lib/libreadline.8.dylib -> libreadline.8.2.dylib
-rw-r--r--  1 root  wheel  432536 Aug 14  2024 /opt/local/lib/libreadline.a
lrwxr-xr-x  1 root  wheel      21 Aug 14  2024 /opt/local/lib/libreadline.dylib -> libreadline.8.2.dylib
-rwxr-xr-x  1 root  wheel  275080 Aug 14  2024 /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64/opt/local/lib/libreadline.8.2.dylib
lrwxr-xr-x  1 root  wheel      21 Aug 14  2024 /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64/opt/local/lib/libreadline.8.dylib -> libreadline.8.2.dylib
-rw-r--r--  1 root  wheel  432536 Aug 14  2024 /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64/opt/local/lib/libreadline.a
lrwxr-xr-x  1 root  wheel      21 Aug 14  2024 /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_21.x86_64/opt/local/lib/libreadline.dylib -> libreadline.8.2.dylib
```

So it is logical to link to libs in `/opt/local/lib` so that linking survives version upgrades.
But we need to check if the port is installed _and_ activated.

native-api:
> we need to check if the port is installed and activated.

The quickest way to check that seems to be:

```bash
[[ $(port -q installed $name | awk '{print $3}') == "(active)" ]]
```

native-api:
> The original code port -q location readline produces /opt/local/var/macports/software/readline/readline-8.2.013_0.darwin_24.arm64.tbz2 (not a directory) on my machine

I guess this might be because it's installed from a binary archive.

Running `port location <name>` (without `-q`) should clarify that.

So we're essentially supporting this case.

0916dhkim:
@native-api Thanks for mentioning the difference between binary archives and standard archives.
> But we need to check if the port is installed and activated.

Let me add this check 👍

plugins/python-build/bin/python-build
```diff
@@ -1509,6 +1509,7 @@ use_macports() {
     local prefix="${port_location%/bin/port}"
     export CPPFLAGS="${CPPFLAGS:+$CPPFLAGS }-I${prefix}/include"
     append_ldflags_libs "-L${prefix}/lib -Wl,-rpath,${prefix}/lib"
+    export PKG_CONFIG_PATH="$prefix/lib/pkgconfig/:${PKG_CONFIG_PATH}"
```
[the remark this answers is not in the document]
native-api: I would check what CPython's `configure` expects to find the packages.

You can the patch from https://github.com/pyenv/pyenv/issues/3229#issuecomment-2969921195 to include `configure`'s trace into the build log to check how exactly it locates the packages we are supposed to guide it towards.

plugins/python-build/bin/python-build
```diff
@@ -1508,7 +1508,8 @@ use_macports() {
   if [ -n "$port_location" ]; then
     local prefix="${port_location%/bin/port}"
     export CPPFLAGS="${CPPFLAGS:+$CPPFLAGS }-I${prefix}/include"
-    append_ldflags_libs "-L${prefix}/lib -Wl,-rpath,${prefix}/lib"
+    prepend_ldflags_libs "-L${prefix}/lib -Wl,-rpath,${prefix}/lib"
```
0916dhkim, line 1511: I tried to get `configure`'s trace as you suggested and found an interesting behavior.

Appending the `-L/opt/local/lib -Wl,-rpath,/opt/local/lib` resulted in
```
*** WARNING: renaming "readline" since importing it failed: dlopen(/Users/danny/.pyenv/sources/3.11.13/Python-3.11.13/build/lib.macosx-15.5-arm64-3.11/readline.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '_rl_catch_signals'
```
while prepending the flag didn't have any issues. It's a little confusing because I see `use_homebrew()` appends the flags without issues.

Attaching the build logs with traces below:
[build-append.log](https://github.com/user-attachments/files/20732338/build-append.log)
[build-prepend.log](https://github.com/user-attachments/files/20732346/build-prepend.log)

native-api: > Do you suggest prepending `CPPFLAGS` to match `LDFLAGS` order like this?

Yes

> Or is this an upstream issue in macports like the homebrew issue you linked.

What I linked is not an "upstream issue". It's just a thread I googled to find out what this error might mean -- which just happened to some Homebrew issue thread.

plugins/python-build/bin/python-build
```diff
@@ -1863,15 +1840,11 @@ use_xcode_sdk_zlib() {
...
-      if [[ -d "$libdir" ]]; then
-        echo "python-build: use zlib from MacPorts"
-        export CPPFLAGS="-I$prefix/include ${CPPFLAGS}"
-        export LDFLAGS="-L$prefix/lib ${LDFLAGS}"
-        lock_in macports
-      fi
+  if [[ $(port -q installed zlib | awk '{print $3}') == "(active)" ]]; then
+    echo "python-build: use zlib from MacPorts"
+    export CPPFLAGS="-I$prefix/include ${CPPFLAGS}"
+    export LDFLAGS="-L$prefix/lib ${LDFLAGS}"
```
native-api, line 1846: ?

0916dhkim: Oops 🙈
Removed

native-api:
Excellent work! :tada:

PR #3289, opened 2025-07-17
Update README to improve the guidance

ykqmain:
### Prerequisite
My PR did not modify the code, but only updated README.

### Description
I referred to the current README guidance on my Mac, which caused trouble. I improved the zsh configuration instructions in the case of using homebrew to install pyenv

### Tests
The steps of installing pyenv and configuring zsh on Mac are clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,13 @@ See [Advanced configuration](#advanced-configuration) for details and more confi
   ```
   
   If you wish to get Pyenv in noninteractive login shells as well, also add the commands to `~/.zprofile` or `~/.zlogin`.
+  
+  If you use Homebrew to install Pyenv, the `pyenv` command should be in the `bin` folder of Homebrew. There is no `bin` folder in `$HOME/.pyenv`, you need to add `/$HOME/.pyenv/shims` to PATH:
+
+  ```zsh
+  echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.zprofile
+  echo '[[ -d $PYENV_ROOT/shims ]] && export PATH="$PYENV_ROOT/shims:$PATH"' >> ~/.zprofile
+  ```
   </details>
 
 #### Fish


native-api (commented):
The currently suggested code supports both cases: when Pyenv is isntalled via Homebrew and when Pyenv is installed via Git.

The code suggested in the PR is incorrect. Instead of adding `shims` to PATH, it needs to run `eval "$(pyenv init - <shell>)"` -- which does that and also other things.

If you had "trouble" with the stock code, either you didn't follow the instructions properly, or you have some unusual setup and as such, need to consult the "Advanced configuraiton" section as specified in the README.

native-api:
The instructions may indeed be unclear -- but the current patch is definitely not the way to fix them.
Maybe you can tell what exactly was unclear -- then we can figure out a way to clarify exactly that.

ykqmain:
> The instructions may indeed be unclear -- but the current patch is definitely not the way to fix them.
> 
> Maybe you can tell what exactly was unclear -- then we can figure out a way to clarify exactly that.

I use macOS, homebrew, zsh, and install and configure pyenv according to the official guidance, it doesn't work and is confusing. For example, the location of the `bin` folder, I guess the official did not go through the complete installation configuration process on Mac. I tried to find the method myself, pyenv works well. My zprofile configuration is as follows:

```zsh
export HOMEBREW="/opt/homebrew"
export PATH="$PYENV_ROOT/shims:$HOMEBREW/bin:$PATH"
```

If I don't add `shims` to PATH, pyenv will print a prompts message every time I start the Terminal.

I hope this experience can help more people, so I submitted the PR. If it doesn't meet the rules, I will close it.

I hope that the official can keep pace with the times and improve README from beginning to end. pyenv is really good and easy to use, thank you.

native-api:
> according to the official guidance, it doesn't work and is confusing

What "official guidance"? Your further phrases cast a doubt that it is indeed [the README](https://github.com/pyenv/pyenv?tab=readme-ov-file#simple-python-version-management-pyenv).
If it is, could you recite step by step how you were looking for the necessary information in it? That'll show where exactly the gap in ease of navigation is so we'll know what and where to add.

(Since your last message, I've added an explanation phrase to the Zsh setup section as other users reported about it earlier. So don't be surprised that you didn't see it before.)

> For example, the location of the `bin` folder

That's what makes me think that you were using some third-party advice. The README never even mentions the word "bin" in the setup sections! (It only mentions it in usage and advanced sections.)

> If I don't add `shims` to PATH, pyenv will print a prompts message every time I start the Terminal.

That's probably a message from OhMyZsh (and as such, has nothing to do with us). Last time I checked, it included a pointer to the official README. If it doesn't, please recite it here so we can work out and suggest a fix to them.

ykqmain:
> > according to the official guidance, it doesn't work and is confusing
> 
> What "official guidance"? Your further phrases cast a doubt that it is indeed [the README](https://github.com/pyenv/pyenv?tab=readme-ov-file#simple-python-version-management-pyenv). If it is, could you recite step by step how you were looking for the necessary information in it? That'll show where exactly the gap in ease of navigation is so we'll know what and where to add.
> 
> (Since your last message, I've added an explanation phrase to the Zsh setup section as other users reported about it earlier. So don't be surprised that you didn't see it before.)
> 
> > For example, the location of the `bin` folder
> 
> That's what makes me think that you were using some third-party advice. The README never even mentions the word "bin" in the setup sections! (It only mentions it in usage and advanced sections.)
> 
> > If I don't add `shims` to PATH, pyenv will print a prompts message every time I start the Terminal.
> 
> That's probably a message from OhMyZsh (and as such, has nothing to do with us). Last time I checked, it included a pointer to the official README. If it doesn't, please recite it here so we can work out and suggest a fix to them.

@native-api 

Yes, the official guidance I mentioned refers to README. I can describe in detail the process of installing pyenv and the problems I encountered, I use macOS, homebrew, zsh, oh-my-zsh.

I installed pyenv with the `brew install pyenv` command, and the bin directory of homebrew has been added to my `.zprofile` file. I can directly use pyenv to install the Python version, such as 3.10.

At this time, the directory under my `.pyenv` folder is as follows:

```
yang:~/.pyenv > tree -L 1
.
├── shims
├── version
└── versions
```

In README's "B. Set up your shell environment for Pyenv" in the `Zsh` part, there is a command `echo '[[ -d $PYENV_ROOT/bin ]] && export PATH= "$PYENV_ROOT/bin:$PATH"' >> ~/.zshrc`, what I mean by `bin` folder is here. In the case of using homebrew to install pyenv, there is no `bin` directory.

If I didn't add `$PYENV_ROOT/shims` to my `.zprofile`, I entered `python3` at the terminal, and the interpreter started was my Xcode's built-in python3.9.6, not 3.10.18 installed by pyenv.

I control the variable step by step. In the plugins section of `.zshrc`, add [pyenv plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/pyenv), and don't add `$PYENV_ROOT/shims` to my `.zprofile`. I restart the terminal, and it will prompt:

```
Found pyenv, but it is badly configured (missing pyenv shims in $PATH). pyenv might not
work correctly for non-interactive shells (for example, when run from a script).

To fix this message, add these lines to the '.profile' and '.zprofile' files
in your home directory:

export PATH="$PYENV_ROOT/bin:$PATH"
eval "$(pyenv init --path)"

You'll need to restart your user session for the changes to take effect.
For more information go to https://github.com/pyenv/pyenv/#installation.
```

In this case, I enter `python3`, and the start-up interpreter is 3.10.18 installed by pyenv.

Finally, I added `$PYENV_ROOT/shims` to my `.zprofile`, and everything worked well. This is my final `.zprofile` configuration:

```
export HOMEBREW="/opt/homebrew"
export PATH="$PYENV_ROOT/shims:$HOMEBREW/bin:$PATH"
```

Compare it with README:

```
echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.zshrc
echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.zshrc
echo 'eval "$(pyenv init - zsh)"' >> ~/.zshrc
```

The biggest difference is that you write `$PYENV_ROOT/bin` to `.zshrc`, while I add `$PYENV_ROOT/shims` to `.zprofile`.

This is also the reason why I have been emphasizing the `bin` folder. I guess your tutorial is based on the method of curl installation or git installation, and the packages installed by brew are actually linked to homebrew's own bin folder (`/opt/homebrew/bin`, `/opt/homebrew/Cellar/pyenv/2.6.1/bin/`). 

In a word, if I don't add `$PYENV_ROOT/shims` to my `.zprofile`, I can't call the python version installed by pyenv in the terminal or other external Editor, IDE.

native-api:
> In README's "B. Set up your shell environment for Pyenv" in the `Zsh` part, there is a command `echo '[[ -d $PYENV_ROOT/bin ]] && export PATH= "$PYENV_ROOT/bin:$PATH"' >> ~/.zshrc`, what I mean by `bin` folder is here. In the case of using homebrew to install pyenv, there is no `bin` directory.

Hmm... Okay.
Those lines of code are designed to work in all the described installation scenarios -- to make the instructions as simple and straightforward as possible.
In the case of Homebrew installation, the line you cited is a no-op. ( `[[ -d $PYENV_ROOT/bin ]]` checks if `$PYENV_ROOT/bin` is present and is a directory and `&&` only executes the following command if it is.)
I guess that's not immediately clear to someone not well-versed in Bash.
(But you did consider yourself well-versed enough in Bash to second-guess and doubt us rather than yourself. Anyway, these are our users, so we'd rather put things in ways that they are able to process. We already had cases of explanations unexpectedly being too complex for the Average Joe. So let's see if we can make this part incite fewer undesired thoughts in the reader's mind...)

For some time now, we have made `pyenv init [<shell>]` print the necessary startup lines for `<shell>` (and without an argument, for the current shell).
So that section can be written like this:

> Add Pyenv startup commands to `~/.zshrc` by running the following in your terminal:
> 
> ```
> pyenv init >>~/.zshrc
> ```
> 
> If you wish to get Pyenv in noninteractive login shells as well, also add the commands to ~/.zprofile or ~/.zlogin.
> 
>```
> pyenv init >>~/.zprofile
>```
>
> (You can run `pyenv init` without redirection to see what lines would be added.)

Would this be clearer?

---

> To fix this message, add these lines to the '.profile' and '.zprofile' files
> in your home directory:
> 
> export PYENV_ROOT="$HOME/.pyenv"
> export PATH="$PYENV_ROOT/bin:$PATH"
> eval "$(pyenv init --path)"
> 

Now, this advice does not look quite correct.
It does work but results in a bogus PATH entry in the case of Homebrew installation and no `pyenv` shell function.
It doesn't seem that OhMyZsh specifically requires it to be so, looking through https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/pyenv/pyenv.plugin.zsh , it seems it would work either way because it reruns `eval "$(pyenv init -)"` itself.

So we can suggest them to fix the advice to use the same code as I suggested above.

Issue #3300, opened 2025-08-12
macOS CLT-only: python-build skips SDK zlib after PR #3298, breaking installs (no full Xcode)


pb-dod:
### Prerequisites
- [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
- [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues?q=is%3Aissue).
- [x] Make sure you are reporting a problem in Pyenv and not seeking consultation with Pyenv usage.
- [x] Make sure your problem is not derived from packaging (e.g. Homebrew).
- [x] Make sure your problem is not derived from plugins.

### Describe the bug
After [PR #3298](https://github.com/pyenv/pyenv/pull/3298) “Skip Zlib from XCode for XCode 16+” (commit `685f281`), `use_xcode_sdk_zlib` calls `xcodebuild -version` to decide whether to add SDK include/lib paths. On Command Line Tools–only systems (no full Xcode), `xcodebuild` errors out, the version is treated as “unknown,” and the function returns early. SDK `CPPFLAGS`/`LDFLAGS` are not added, so Python 3.9.x installs fail unless an alternative zlib (e.g., Homebrew) is found.

Expected: CLT-only hosts should continue to build Python by using the SDK from `xcrun --show-sdk-path`, consistent with the zlib guidance in the Common build problems page.

### Reproduction steps
```bash
# CLT-only environment (no full Xcode app)
xcode-select -p
# /Library/Developer/CommandLineTools

# Install Python (via asdf or pyenv)
asdf install python 3.9.8
# also reproducible with any 3.9.x including 3.9.23
# or: pyenv install -v 3.9.8
```

Key early error message:
```text
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
```

Tail:
```text
BUILD FAILED (OS X 15.6 using python-build 2.6.6-1-g685f2815)
```

If `brew install zlib` is performed first, the install proceeds despite the `xcodebuild` error being printed (console shows `python-build: use zlib from homebrew` and installation completes).

### Diagnostic details
- [x] Platform information: macOS 15.6 (CLT only; no full Xcode app)
- [x] OS architecture: arm64
- [x] pyenv version: python-build 2.6.6-1-g685f2815 (via pyenv/asdf)
[...]

PR #3298, opened 2025-08-07
Skip Zlib from XCode for XCode 16+

native-api:
It breaks the build due to yet unknown structural changes from XCode 15

Make sure you have checked all steps below.

### Prerequisite
* [x] Please consider implementing the feature as a hook script or plugin as a first step.
  * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins](https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a plugin if possible.
* [x] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv will be imported to pyenv too, eventually.
  * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with rbenv.
* [x] My PR addresses the following pyenv issue (if any)
  - Closes https://github.com/pyenv/pyenv/issues/3068

### Description
- [x] Here are some details about my PR

Due to yet unknown structural differences between XCode 16 and XCode 15, adding directories from the former to `*FLAGS` breaks the build.

### Tests
- [x] My PR adds the following unit tests (if any)
N/A

PR #3301, opened 2025-08-12
Use Zlib from XCode SDK in a way compatible with XCode 16

JarheadStever:
### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3300

### Description
- Fix python build failure when full Xcode is not installed 
- See https://github.com/pyenv/pyenv/issues/3300 for further context/discussion

**Problem:**
We used to be able to install Python (e.g. `asdf install python 3.9.8`) with only Command Line Tools installed (without full Xcode). The `use_xcode_sdk_zlib()` function would succeed before the Xcode major version check was added in #3298. This version check now causes the build to fail with:
```
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
BUILD FAILED (OS X 15.6 using python-build 2.6.6-1-g685f2815)
```

**Root Cause:**
The function now calls `xcodebuild -version` unconditionally, but this command fails when only Command Line Tools are installed (without full Xcode).

**Solution:**
- check for XCode16+ using just `xcrun` without relying on `xcodebuild` or `xcodebuild -version` being present.
  - determine XCode version by checking the major version of the MacOS SDK (`>=15.x` corresponds to XCode `>=16.x`)
  - set `CPPFLAGS` and `LDFLAGS` for either case, making appropriate tweaks depending on SDK version

### Tests
N/A


JarheadStever:
Closing for #3302

native-api:
The quick fix is done.
This PR might come in handy yet, don't close it just now. Let's discuss the proper fix in the issue.

plugins/python-build/bin/python-build
```diff
@@ -1826,24 +1826,42 @@ use_homebrew_zlib() {
...
-  if [ -d "$xc_sdk_path" ]; then
-    echo "python-build: use zlib from xcode sdk"
+
+  local sdkroot="$(xcrun --sdk macosx --show-sdk-path 2>/dev/null || true)"
+  local sdkver="$(xcrun --sdk macosx --show-sdk-version 2>/dev/null || true)"
+
+  # If we can't determine SDK info, be conservative and don't enable this logic
+  if [ -z "$sdkroot" ] || [ -z "$sdkver" ]; then
+    return 1
+  fi
```
JarheadStever, line 1838: I assume this is how we'd want to handle this case?

native-api:
Looks like "LLM"'s advice was not quite on point:
```
ld: unknown options: -sysroot 
```

I don't think we even need to pass "sysroot". The compiler will use it anyway, and looking at https://github.com/python/cpython/blob/3.9/setup.py#L1616 and `find_file` and `find_library_file` implementation, it does look in the default sysroot.

I'll try to not pass anything at all and see what happens.

JarheadStever:
Getting caught back up here - thanks for your continued work on this @native-api & sorry for the confusion and extra code. As I said, I'm a bit out of my element here, but this has been a good learning experience.

Looking at the current diff, it seems like we're not explicitly checking the versions of XCode or MacOS SDK at all anymore?

native-api:
According to the listed research results, we no longer need to do anything differently for XCode 16, adding just `-isysroot` in both cases should work for everything since 3.5.2.

@puhitaku This change will likely break Zlib search in MacOS for 2.7 . If you need it, you'll probably need to backport CPython's "Issue #25136" there (search CPython commit history for messages with that string for relevant commits).

native-api:
LLM's info on ".tbd" files wasn't accurate, either: it's not new in XCode 16 (thus not a direct cause of the current breakage), it goes back all the way to XCode 7 (https://bugs.python.org/issue25136).

native-api:
Thank you for your contribution! Even though there turned out to be a better solution, you still put in the groundwork and shared knowledge that I didn't have that saved a lot of time and effort! So you were rightfully credited in the commit metadata and are a full-fledged contributor now!

Issue #2082, opened 2021-09-30
Bump Bats to the latest release


native-api:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [x] Platform information (e.g. Ubuntu Linux 16.04): all
- [x] OS architecture (e.g. amd64): all
- [x] pyenv version: all
- [x] Python version: all
- [x] C Compiler information (e.g. gcc 7.3): all
- [x] Please attach verbose build log as gist
  * You can turn on verbose debug logging using by setting `PYENV_DEBUG=1`, e.g. `env PYENV_DEBUG=1 pyenv install -v 3.6.4`

Latest Bats version has improved facilities -- in particular, allows to print arbitrary info to fd 3 to diagnose test failures -- I've had lots of trouble without that.

It'd also allow to use latest assert lib with more functionality.

HOWEVER, https://github.com/jasonkarns/bats-mock is abandoned, so we WILL have to maintain it ourselves.

native-api:
Bats-Mock also doesn't support "do this for these arguments, regardless of how many times it's called" which we need.

Due to its absense, we have to constantly adjust call counts in lots of tests for stuff like `brew` and `uname`.

native-api:
Completed in https://github.com/pyenv/pyenv/pull/3324

Side tasks like refactoring are going to be completed separately

PR #3324, opened 2025-09-10
Bats test suite in docker

rockandska:
### Description
This is a following of this [discussion](https://github.com/orgs/pyenv/discussions/3321) with @native-api as a POC
Should squash some commits to clean up the history, but let them as it is for now for clarity.

What achieves this PR ?

- Running bats suite under docker and locally
- Running test in a closed environment (no shared files/dirs, no symlink)
- Running bats suite with different bash version with/without gnu tools
  - docker container use alpine/busybox
- Ability to add new bash version to test easily by updating the var in Makefile
- Fix some test who were not isolated
- Update shell detection (did not work under alpine docker)
- Add a README draft for test
- Ability to filter file/test file when running the make target 

Open to discussion / comments

Regards,


native-api:
Okay, this is huge!
In particular, this also completes a long-standing task, https://github.com/pyenv/pyenv/issues/2082 !
So we can make a payout from donated money if you're interested!

rockandska:
> Okay, this is huge!

Don't make me blush ^^ the hardest part was to discover why some tests were failing, everything else was adapted from what I already use in other project.

>  So we can make a payout from donated money if you're interested!

No, thanks, I'm just happy to using an old [PR](https://github.com/bats-core/bats-core/pull/413) of mine in bats, and contribute to a project I use often.
If it is okay for you, you could donate to a non-profit organization you want (Wikipedia, Signal, ...)

Let me know if you want me to fix / squash some commits or separate some functionalities in different PR

Regards

native-api (commented):
Still wrapping my head over this :smile:  I don't have much personal experience with makefiles of dockerfiles!

* AFAICS, there are at least a few separate concerns here:
  1. porting the tests to Bats 1.10;
  2. adding testing under Docker
  3. fixing found bugs in tests
  4. optimizing shell detection

They should be grouped together if possible so that when browsing history, it's clear why each change was made and all changes that achieved a certain objective are seen together. It is okay to leave multiple commits per concern if it was achieved in multiple independent steps (e.g. migrating to Bats 1.5.0 and then to 1.10.0 if it's important to understand the changes for some reason).

native-api (commented):
* I see that you migrated to Bast 1.5.0 and then to 1.10.0. But there are version checks for 1.5.0 left in the code. Are they still relevant? Do we want to support multiple Bats releases? If you want to leave a check in case someone tries to run this in an old version of Bats, maybe just check in `setup_suite `?

Makefile
```diff
@@ -1,3 +1,86 @@
+TEST_BATS_VERSION = v1.10.0
+TEST_BASH_VERSIONS = 3.2.57 4.1.17
```
native-api, line 2: It would be preferable if we have some kind of moving targets for Bash versions so that we don't have to track Bash releases and make changes each time a new Bash release is made.

rockandska: Just add a new version in TEST_BASH_VERSIONS and all the needed target are "created"
All the targets are dynamic and based on those variables

rockandska: > Perhaps it's possible to somehow extract the versions used in other major distros?
Is is really practical to test with arbitrary Bash versions in Alpine?

You can add all the bash versions needed.
Testing different bash versions is only to be sure that code is portable to all bash version who could be used, so yes you could target only BASH_VERSION you like.
I don't find a gist I've once, but there is so many bugs in bash with array for example and another things

rockandska:
> * I see that you migrated to Bast 1.5.0 and then to 1.10.0. But there are version checks for 1.5.0 left in the code. Are they still relevant? Do we want to support multiple Bats releases? If you want to leave a check in case someone tries to run this in an old version of Bats, maybe just check in `setup_suite`?

ok I didn't get what you're meaning at first but get it now.
It is not really a check, if this line is not present where `run -127` is used, you have some warnings because you just "run" but the result is "127" and there is no assertion about the success or not.
It was added to not have some warnings displayed and minimal required version to use `run -XXX` is `1.5`

Makefile
native-api: Okay. Having an option to pick arbitrary versions is still great for a start! We can figure something out going forward if we wish to test in different distros with their stock and potentially patched versions.

rockandska:
> Still wrapping my head over this 😄 I don't have much personal experience with makefiles of dockerfiles!
> 
>     * AFAICS, there are at least a few separate concerns here:
>       
>       1. porting the tests to Bats 1.10;
>       2. adding testing under Docker
>       3. fixing found bugs in tests
>       4. optimizing shell detection
> 
> 
> They should be grouped together if possible so that when browsing history, it's clear why each change was made and all changes that achieved a certain objective are seen together. It is okay to leave multiple commits per concern if it was achieved in multiple independent steps (e.g. migrating to Bats 1.5.0 and then to 1.10.0 if it's important to understand the changes for some reason).

I resume what I could push as a dedicated PR in order I should open them, this PR is just a POC to let you test the final result :
- fix tests isolation/bug discovered by isolating them
- updating SHELL detection
- updating bats
- test isolation
- docker targets with filtering

Makefile
rockandska: The version who could be added with this are only the ones available as official bash image on [dockerhub](https://hub.docker.com/_/bash)

I could check within few days, if you want, the possibility to use official distribution image published instead of the bash ones

Something like :

```
BASE_IMAGES: bash:3.2.57 ubuntu:22.0 alpine:10
```

native-api:
> I resume what I could push as a dedicated PR in order I should open them, this PR is just a POC to let you test the final result

That's not needed at all. It's okay to have multiple related concerns within the same PR, they just need to be each in its separate commit -- effectively, cleaning up commit history. See [CONTRIBUTING.md](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md).
I do this by rebasing interactively in a GUI, rearranging, squashing and splitting commits as necessary. Since it's a complicated and error-prone process, I do it in a clone of the branch to be able to go back if I mess up. Ultimately, I overwrite the initial branch with the result.

rockandska:
> That's not needed at all. It's okay to have multiple related concerns within the same PR, they just need to be each in its separate commit -- effectively, cleaning up commit history. See [CONTRIBUTING.md](https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md).
> I do this by rebasing interactively in a GUI, rearranging, squashing and splitting commits as necessary. Since it's a complicated and error-prone process, I do it in a clone of the branch to be able to go back if I mess up. Ultimately, I overwrite the initial branch with the result.

Ok so I will rework the history to let only the 5 commits listed as above

Makefile
native-api: > I could check within few days, if you want, the possibility to use official distribution image published instead of the bash ones

Looking at the tags of the Bash image, they seem to indeed have rolling-release ones: e.g. `5` and `4` as well as `latest`.

For the practical goals above, It seems enough to test in `bash:latest` (or whatever is considered the latest stable version that we want to be ready for) and `bash:4` as well as `latest` and maybe 1 other non-EOL release of a few other mainstream distros.

This looks like a job for a separate PR -- so let's not hold up this one for that.

Issue #3325, opened 2025-09-11
Python-build does not work with curl as a snap


eha-cs:
We use python-build to provision specific python versions.
This fails on my machine like this:

```bash
❯ /home/eha/.local/share/spin/pyenv/plugins/python-build/bin/python-build -verbose 3.9.8 /home/eha/contact/tmp
/tmp/python-build.20250911085657.46647 ~/contact/etc/pyhon-snippeteha/.local/share/spin/pyenv/plugins/python-build/bin/python-build -verbose 3.9.8 /home/eha/contact/tmp                              
Downloading Python-3.9.8.tar.xz...
-> https://www.python.org/ftp/python/3.9.8/Python-3.9.8.tar.xz
error: failed to download Python-3.9.8.tar.xz

BUILD FAILED (Ubuntu 24.04 using python-build 2.6.7-13-g2ecd676a)

Results logged to /tmp/python-build.20250911085657.46647.log

Last 10 log lines:
/tmp/python-build.20250911085657.46647 ~/contact/etc/pyhon-snippet
```
If I exchange curl with wget using 
```bash
export PYTHON_BUILD_HTTP_CLIENT=wget
```
the same command succeeds.
I suspect it has to do with the fact I use curl as a snap.
```bash
❯ curl --version
curl 8.16.0 (x86_64-pc-linux-gnu) libcurl/8.16.0 OpenSSL/3.5.2 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.59.0 nghttp3/1.11.0 librtmp/2.3 OpenLDAP/2.6.7
Release-Date: 2025-09-10
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```
The curl snap cannot download to hidden files, e.g. folders starting with dots. Perhaps you download the tar.xz to somewhere like ~/.tmp ? That could be what's breaking things.

A workaround for me is to set the environment variable as explained above, but I feel like this should also work with curl as a snap.

PR #3329, opened 2025-09-14
Support curl from a snap

native-api:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3325

### Description
Same as for aria2c from a snap

### Tests


rockandska:
Stupid question, but what is the advantage to using curl instead of wget ? Since wget is always installed on system and not prone to snap problem ?

native-api:
> Stupid question, but what is the advantage to using curl instead of wget ? Since wget is always installed on system and not prone to snap problem ?

The immediate reason is historical. [It was so all the way in Ruby-Build's initial commit](https://github.com/rbenv/ruby-build/blame/cacc406e04c7fd51bec76dffd72a0c7b5b278608/bin/ruby-build#L97).
My guess is that on the surface, `curl` appears to be more script-friendly. Wget writes into a file that it itself decides on by default and its output is not very suitable for embedding, either. Curl's availability on the system was probably not a concern.

PR #3330, opened 2025-09-15
upgrade pip detetcion in pip-rehash to accept multiple variations:

rockandska:
- pip
- pip3
- pip3.10
- python -m pip
- uv pip

Co-author: Christian Fredrik Johnsen <christian@johnsen.no> @ChristianFredrikJohnsen 

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3031
* [X] My PR addresses the following pyenv PR (if any)
  - Closes https://github.com/pyenv/pyenv/issues/3140

### Tests

tests added in new pyenv-rehash file

### DIFF
--- a/pyenv.d/exec/pip-rehash.bash
+++ b/pyenv.d/exec/pip-rehash.bash
@@ -4,6 +4,9 @@ PYENV_REHASH_COMMAND="${PYENV_COMMAND##*/}"
 # Remove any version information, from e.g. "pip2" or "pip3.10".
 if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.[0-9]+)?$ ]]; then
   PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
+# Check for ` -m pip ` in arguments
+elif [[ "$*" =~ [[:space:]]-m[[:space:]]pip[[:space:]] ]]; then
+  PYENV_REHASH_COMMAND="pip"
 fi
 
 if [ -x "${PYENV_PIP_REHASH_ROOT}/${PYENV_REHASH_COMMAND}" ]; then
--- a/test/pip-rehash.bats
+++ b/test/pip-rehash.bats
@@ -1,0 +1,67 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+create_executable() {
+  local bin="${PYENV_ROOT}/versions/${PYENV_VERSION}/bin"
+  mkdir -p "$bin"
+  echo "#!/bin/sh" > "${bin}/$1"
+  chmod +x "${bin}/$1"
+}
+
+copy_src_pyenvd() {
+  mkdir -p "${PYENV_ROOT}"
+  cp -r "${BATS_TEST_DIRNAME}/../pyenv.d" "${PYENV_ROOT}"
+}
+
+@test "pip-rehash triggered when using 'pip'" {
+  export PYENV_VERSION="3.7.14"
+  create_executable "example"
+  create_executable "pip"
+  copy_src_pyenvd
+  run command -v example 2> /dev/null
+  assert_failure
+  run pyenv-exec pip install example
+  assert_success
+  run command -v example 2> /dev/null
+  assert_success
+}
+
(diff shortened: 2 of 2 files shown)


pyenv.d/exec/pip-rehash.bash
```diff
@@ -4,6 +4,12 @@ PYENV_REHASH_COMMAND="${PYENV_COMMAND##*/}"
 # Remove any version information, from e.g. "pip2" or "pip3.10".
 if [[ $PYENV_REHASH_COMMAND =~ ^(pip|easy_install)[23](\.[0-9]+)?$ ]]; then
   PYENV_REHASH_COMMAND="${BASH_REMATCH[1]}"
+# Check for `uv pip ` in arguments
+elif [[ "$*" =~ uv[[:space:]]pip[[:space:]] ]]; then
+  PYENV_REHASH_COMMAND="pip"
```
native-api, line 9: This deals with `uv` specifics so if it's needed, it belongs in a middleware plugin (e.g.  in another `exec` hook).

I've no idea how or why this code get invoked on such an invocation and whether it'll continue working going forward. It's `uv` maintainers' job to keep track of this.

test/pip-rehash.bats
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+@test "pip-rehash set PYENV_REHASH_COMMAND right when using 'pip'" {
+  PYENV_COMMAND="pip"
+  source "${BATS_TEST_DIRNAME}/../pyenv.d/exec/pip-rehash.bash"
```
native-api, line 7: The tests should probably rather check if `pyenv-rehash` is called after invoking a stub executable with the corresponding name.

This `source` command does not run the hook in the same environment as it would normally run in so it can behave differently than in real use and we won't see that.

native-api (changes requested):

pyenv.d/exec/pip-rehash.bash
native-api: CC @bersbersbers @ChristianFredrikJohnsen @lczyk

rockandska: It was here so I've let it, but for now it's removed

test/pip-rehash.bats
rockandska: Done

rockandska:
my test is wrong, sorry for the false alert, need to rework it

rockandska:
Don't get why last push fails only on MacOs and don't have one, any idea @native-api ?

native-api:
I guess you need to write  a shebang to the stub for it to be a valid executable

rockandska:
wow, I discovered by using [mxschmitt/action-tmate](https://github.com/mxschmitt/action-tmate) (since I don't have a macos in my hand) that the `PATH` declaration in the global setup()` doesn't work for an obscure reason.
Need to investigate more in depth to see what's going on

native-api:
I could reproduce this locally. Investigating.

native-api:
It's indeed the lack of shebang

rockandska:
> It's indeed the lack of shebang

I've add it in another branch without success

native-api:
Well, when I added the shebang, the tests still failed for some other reason.

I think I can fix the tests in the next few hours. Since I can reproduce the problem locally, I seem to be in a better position for that.

rockandska:
> Well, when I added the shebang, the tests still failed for some other reason.
> 
> I think I can fix the tests in the next few hours. Since I can reproduce the problem locally, I seem to be in a better position for that.

Alright, but will be able to investigate further more in a few days when I'll have a better connection.
In the meantime, I force push an update I didn't catch before even if it is not related with the strange behavior we observed

rockandska:
Found the culprit.... a "/" at the end of my `cp` command.....

```
cp -r "${BATS_TEST_DIRNAME}/../pyenv.d/" "${PYENV_ROOT}"
```
instead of
```
cp -r "${BATS_TEST_DIRNAME}/../pyenv.d" "${PYENV_ROOT}"
```

PR #3332, opened 2025-09-17
pyenv-install upgrade message based on PYENV_ROOT

rockandska:
### Description
PYENV_ROOT used in pyenv-install for upgrade message because used everywhere and more robust

### Tests
Test upgraded in consequence

### DIFF
--- a/plugins/python-build/bin/pyenv-install
+++ b/plugins/python-build/bin/pyenv-install
@@ -256,7 +256,6 @@ for DEFINITION in "${DEFINITIONS[@]}"; do
   # Display a more helpful message if the definition wasn't found.
   if [ "$STATUS" == "2" ]; then
     { candidates="$(definitions "$DEFINITION")"
-      here="$(dirname "${0%/*}")/../.."
       if [ -n "$candidates" ]; then
         echo
         echo "The following versions contain \`$DEFINITION' in the name:"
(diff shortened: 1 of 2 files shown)


plugins/python-build/bin/pyenv-install
```diff
@@ -266,12 +265,12 @@ for DEFINITION in "${DEFINITIONS[@]}"; do
       echo "See all available versions with \`pyenv install --list'."
       echo
       echo -n "If the version you need is missing, try upgrading pyenv"
-      if [ "$here" != "${here#$(brew --prefix 2>/dev/null)}" ]; then
+      if [ "$PYENV_ROOT" != "${PYENV_ROOT#$(brew --prefix 2>/dev/null)}" ]; then
```
native-api, line 268: Here, we need to check installation root, not user data root.
In the case of Homebrew installation, installation root is under Homebrew while user data root is still under home directory.

So no...

You are not the first person to be confused here.
So it's probably worth detecting and setting this other root explicitly in the `pyenv` launcher.

native-api (changes requested):

plugins/python-build/bin/pyenv-install
rockandska: Alright, wasn't aware that installation path could be different of versions/shims path.
So, the better option is to have a PYENV_INSTALL_PATH declared in pyenv and use it anywhere, right ?
Could try to implement it if you would like

PR #3334, opened 2025-09-18
PYENV_INSTALL_PATH created and exported

rockandska:
### Description
Add PYENV_INSTALL_PATH as an exposed variable by `pyenv` corresponding to the directory where pyenv is installed

### Tests
Test updated to reflect the change


rockandska:
Not interested in this one ?
Should I close it ?

native-api:
I was just occupied by other things. In fact, I looked at this a few hours ago.

libexec/pyenv
```diff
@@ -76,23 +76,27 @@ export PYENV_DIR
 
 shopt -s nullglob
 
-bin_path="$(abs_dirname "$0")"
-for plugin_bin in "${bin_path%/*}"/plugins/*/bin; do
+PYENV_INSTALL_PATH="$(abs_dirname "$0")"
```
native-api, line 79: Maybe `PYENV_PREFIX`?

It's canonically called "prefix" in UNIX (and GNU autotools and CPython's `configure` in particular) because in the FHS, there's no such thing as "install path", as many applications can be installed alongside. But there are possible autonomous subtrees, with the root of such subtree being "prefix".

Perhaps also with a leading underscore to show that this is an internal envvar that is not supposed to be user-overridable (unlike e.g. `PYENV_ROOT`) so we can overwrite it with impunity.

plugins/python-build/test/pyenv.bats
```diff
@@ -140,7 +140,8 @@ OUT
   stub_python_build 'echo ERROR >&2 && exit 2'
   stub_python_build "--definitions : echo 2.6.9 2.7.9-rc1 2.7.9-rc2 3.4.2 | tr ' ' $'\\n'"
 
-  run pyenv-install 2.7.9
+  mkdir "$BATS_TEST_TMPDIR/.git"
+  PYENV_INSTALL_PATH="$BATS_TEST_TMPDIR" run pyenv-install 2.7.9
```
native-api, line 144: Not sure why adaptations like this are needed (same elsewhere). Setting `PYENV_INSTALL_PATH` in the common `setup()` should be all that's needed. If it's not, something's probably not right.

test/init.bats
```diff
@@ -37,7 +37,7 @@ create_executable() {
 
 @test "setup shell completions" {
   exec_root="$(cd $BATS_TEST_DIRNAME/.. && pwd)"
-  run pyenv-init - bash
+  PYENV_INSTALL_PATH="${exec_root}" run pyenv-init - bash
```
native-api, line 40: Looks like we rather need to remove `exec_root` altogether in favor of `PYENV_INSTALL_PATH` (same elsewhere)

test/--version.bats
```diff
@@ -3,7 +3,7 @@
 load test_helper
 
 _setup() {
-  export GIT_DIR="${PYENV_TEST_DIR}/.git"
+  export GIT_DIR="${HOME}/.git"
```
native-api, line 6: Not sure about the reason behind this change. We set `HOME` in order to safely run `git set --global`. The `.git` directory should normally be under `PYENV_INSTALL_PATH` (but we probably cannot fake it lest scripts won't be able to find stuff). AFAICS if `GIT_DIR` is set, `git` uses it regardless of where it is -- so just setting it and creating the fake directory as necessary should successfully mock the git-related logic.

native-api (commented):
In general, could you check all instances of `BASH_SOURCE` usage? This is usually the sign that we're trying to detect the installation prefix. I saw one such place in `pyenv---version`.

rockandska:
Thanks for the feedbacks, will try to make the change in a few days

native-api:
I apologize for taking long to review. I looked at this soon after you made it, had doubts about it and thought that perhaps I can allocate time to make some own changes to address them myself. But about some of my changes, I wasn't sure, either, so it ultimately went nowhere. Should've probably started a discussion right away -- now all that my strategy achieved is waste 3 weeks.

PR #3385, opened 2026-01-08
CI: exclude unsupported MacOS-x64 for recent Miniconda versions

binbjz:

### Prerequisite
  * pyenv has some powerful support for plugins and hook scripts. Please refer to [Authoring plugins]
(https://github.com/pyenv/pyenv/wiki/Authoring-plugins) for details and try to implement it as a
plugin if possible.
* [ ] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv),
since we have borrowed most of the code from that project.
  * We occasionally import the changes from rbenv. In general, you can expect changes made in rbenv
will be imported to pyenv too, eventually.
  * Generally speaking, we prefer not to make changes in the core in order to keep compatibility with
rbenv.

### Description
  - Split the macOS Intel build into a dedicated job.
  - Build the Intel matrix from versions that actually include `MacOSX-x86_64` in their definition
files.
  - Keep non-Intel macOS builds unchanged.

### Tests
  - None (workflow-only change; relies on CI jobs)

### DIFF
--- a/.github/workflows/modified_scripts_build.yml
+++ b/.github/workflows/modified_scripts_build.yml
@@ -6,7 +6,7 @@ jobs:
     outputs:
       versions: ${{steps.modified-versions.outputs.versions}}
       versions_cpython_only: ${{steps.modified-versions.outputs.versions_cpython_only}}
-      versions_macos_intel_supported: ${{steps.modified-versions.outputs.versions_macos_intel_supported}}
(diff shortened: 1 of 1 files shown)


native-api (approved):
Thank you!

Changed the logic to prescribing rather than detecting if MacOS-x64 is supported -- because we do want to verify that all architectures that we expect are present.

PR #3389, opened 2026-01-12
Simplify resolving symlinks: replace native extension with `pwd -P`

native-api:

### Prerequisite
  - N/A

### Description

This method is used in Rbenv (https://github.com/rbenv/rbenv/pull/1428) and Homebrew (https://github.com/Homebrew/brew/blob/fa5368edd38eb8d43f8cfcca15c7cfb83ee5a78d/Library/Homebrew/brew.sh#L76-L78).

`pwd -P` is portable between Linux and BSD (MacOS), unlike `readlink -f`.

### Tests

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -20,41 +20,6 @@ abort() {
   exit 1
 }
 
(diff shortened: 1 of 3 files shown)


native-api:
`readlink -f` is now supported by all no-EOL macOS and FreeBSD versions.

So we can just use that instead.

PR #3425, opened 2026-03-14
suppress "Invalid argument" error for Ubuntu coreutils

razielvamp666:
In Ubuntu 26.04 used  new `uutils coreutils` written in Rust. `readlink` behavior is little bit different in comparison with `GNU coreutils`. If `readlink` parameter is not a link it will print `Invalid argument` error. What is making a bunch of noise in console:

```text
readlink: pyenv: Invalid argument
readlink: pyenv: Invalid argument
readlink: conda.bash: Invalid argument
readlink: source.bash: Invalid argument
readlink: envs.bash: Invalid argument
```

Here is a possible option `-s` (or `--silent`) to suppress the errors. Option `-s` also exists in `GNU coreutils` so it should be okay for both versions

### DIFF
--- a/libexec/pyenv
+++ b/libexec/pyenv
@@ -33,7 +33,7 @@ else
   [ -n "$READLINK" ] || abort "cannot find readlink - are you missing GNU coreutils?"
 
   resolve_link() {
-    $READLINK "$1"
+    $READLINK -s "$1"
   }
 
   abs_dirname() {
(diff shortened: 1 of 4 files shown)


native-api (changes requested):
As you may have seen from the test failures, BSD `readlink` does not support `-s`.

native-api:
[Since uutils 0.7.0](https://github.com/uutils/coreutils/commit/1c023bfad9a5618cbf3a89a2042e9f22d575ba59), the silent mode is the default.

Confirmed that it's in effect in Ubuntu 26.04:

```
$ readlink --version
readlink (uutils coreutils) 0.8.0
```

So this change is no longer needed.

PR #3440, opened 2026-05-07
Add CPython 3.15.0b1

jsirois:

### Prerequisite

### Description

### Tests

---
## Summary by cubic
Add CPython 3.15.0b1 to `python-build` so users can install the first 3.15 beta. Updated source URLs and checksums for both `.tar.xz` and `.tgz`, and added the `3.15.0b1t` variant.

- **New Features**
  - New definitions: `3.15.0b1` and `3.15.0b1t`.
  - Uses official python.org tarballs with SHA256 validation.
  - Keeps existing steps: `verify_py315`, `copy_python_gdb`, and `ensurepip`.

<sup>Written for commit 408c9730c1b30b16e8b43facd5523d558453eb52. Summary will update on new commits.</sup>


native-api:
Thank you! Our autogeneration script malfunctioned because they also released a prerelease 3.14.5rc1 at the same time -- which is not supposed to happen for non-initial releases.

jsirois:
> Thank you! Our autogeneration script malfunctioned because they also released a prerelease 3.14.5rc1 at the same time -- which is not supposed to happen for non-initial releases.

Aha! I was wondering about that.

Issue #3451, opened 2026-05-22
Streamline and automate the installation procedure


native-api:
**Describe the intended use case**
* What your general environment is if it's relevant to the feature request and is not a generic console with a typical Pyenv installation (CI, server with a custom setup, cloud environment, IDE)

An arbitrary system

* What you are trying to achieve

Get Pyenv up and running with minimal effort

* What specifically you are doing for that regarding Pyenv

- `curl https://pyenv.run | bash`
- If it errors out because Git isn't installed, find out the system's package manager and install Git with it
- `~/.pyenv/bin/pyenv init 2>&1 | { [[ -e ~/.bash_profile ]] && cat >>~/.bash_profile || cat >>~/.profile; }`
- `~/.pyenv/bin/pyenv init 2>&1 | cat >>~/.bashrc`
- go to https://github.com/pyenv/pyenv/wiki#suggested-build-environment and find a command to install build requirements
- install build requirements with the system's package manager
- restart the shell

* Where you are stuck

That's a lot of work and many steps are nontrivial.

**Describe the solution you'd like**

Perform those steps automatically as part of the installer as much as possible.
Take things like Homebrew and OhMyZsh as example.

**Describe alternatives you've considered**
[x] In particular, did you consider [writing a plugin](https://github.com/pyenv/pyenv/blob/master/README.md#pyenv-plugins)?

**Additional context**

PR #3454, opened 2026-05-24
init: add --install for shell setup

macayu17:
### Prerequisite
  * This belongs in `pyenv init`, since it reuses pyenv's existing shell detection and startup-file guidance.
  * This is pyenv setup behavior, but the same option may be useful upstream too.
  - Addresses https://github.com/pyenv/pyenv/issues/3451

### Description

This adds `pyenv init --install`.

It configures the shell startup files detected by `pyenv init`. If a startup file already contains Pyenv-related code, it refuses to edit that file and tells the user to review it manually and run `pyenv init <shell>` to see the suggested setup.

This covers Bash, Zsh, Ksh, Fish, and PowerShell. Fish also runs the universal-variable setup interactively before writing `config.fish`.

The inserted setup snippets share the same helper functions used by the existing `pyenv init` guidance, and the tilde paths are expanded only when the files are about to be written.

The usage text now keeps `--install` and `--detect-shell` separate from the `eval "$(pyenv init ...)"` form, since those modes are meant to be run directly.

### Tests

Added tests for:

- `--install` appearing in completions
- no-argument shell detection with `pyenv init --install`
- Bash `.bashrc` and profile updates
- existing `.bash_profile` selection
- Zsh, Fish, and PowerShell startup files
- Fish universal-variable setup
- startup files that already contain Pyenv-related code
- unreadable startup files without partial writes
- unsupported shells failing cleanly
- Fish startup files that already contain a generic `end` line

Checks run:

- `PATH=./bats/bin:$PATH bats test/init.bats`
- `PATH=./bats/bin:$PATH bats test/help.bats test/commands.bats`
- `PATH="$PWD/libexec:$PWD/bin:$PATH" pyenv-help --usage init`
- `git diff --check`


libexec/pyenv-init
```diff
@@ -186,6 +203,75 @@ function help_() {
...
+
+function append_lines() {
+  local file="$1"
+  local lines="$2"
+  local sentinel_line
+
+  sentinel_line="$(printf '%s\n' "$lines" | tail -n 1)"
+  if grep -Fqx "$sentinel_line" "$file" 2>/dev/null; then
+    return 0
+  fi
```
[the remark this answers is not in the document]
native-api: 1. I agree with the AI here. We would break the user's shell setup if they have an old or customized version of the setup code!  
I would stay on the safe side and bail out simply if there's a word "pyenv" anywhere in the file. Saying smth like: _"Cannot automatically apply changes to &lt;file&gt;: it appears to already contain Pyenv-related code. Review the file's contents and apply changes manually if necessary. Run `pyenv init <shell name>` to see the suggested setup."_
2. Silencing `grep` errors here seems dangerous. That would create an illusion that the installation went well -- when in actuality, no changes were made!  
According to the manpage, in `-q` mode, `grep` returns 0 if a match is found -- or if there's another error.

libexec/pyenv-init
```diff
@@ -186,6 +203,75 @@ function help_() {
...
+
+  mkdir -p "${file%/*}"
+  touch "$file"
```
native-api: Seems redundant. If it doesn't exist, `[ -s "$file" ]` will fail and the `printf` line will work normally.

libexec/pyenv-init
```diff
@@ -186,6 +203,75 @@ function help_() {
...
+
+  if [ -s "$file" ] && [ -n "$(tail -c 1 "$file" 2>/dev/null)" ]; then
```
native-api: Is there a compelling reason for `[` instead of `[[`? `[[` has fewer quirks and richer syntax. In particular, no need to quote arguments in most cases and you can combine conditions.

Again, silencing errors seems questionable.

libexec/pyenv-init
```diff
@@ -95,21 +105,28 @@ function detect_profile() {
 
   case "$shell" in
   bash )
-    if [ -e '~/.bash_profile' ]; then
+    if [ -e "${HOME}/.bash_profile" ]; then
       profile='~/.bash_profile'
+      profile_path="${HOME}/.bash_profile"
```
native-api: It's possible to expand the tilde manually later [with `"${var/#\~/$HOME}"`](https://stackoverflow.com/questions/3963716/how-to-manually-expand-a-special-variable-ex-tilde-in-bash) without the need to introduce duplicating variables in all branches.

libexec/pyenv-init
```diff
@@ -9,6 +11,7 @@ set -e
 if [ "$1" = "--complete" ]; then
   echo -
   echo --path
+  echo --append
```
native-api: If it's supposed to fully implement the shell setup step (as per my comment in the review message), "`--install`" is probably a name that conveys the intention more clearly.

native-api (commented):
What we need for https://github.com/pyenv/pyenv/issues/3451 for the shell setup step is a replacement for:

>     ~/.pyenv/bin/pyenv init 2>&1 | { [[ -e ~/.bash_profile ]] && cat >>~/.bash_profile || cat >>~/.profile; }
>     ~/.pyenv/bin/pyenv init 2>&1 | cat >>~/.bashrc

(or the counterpart for the user's shell as per the README) to be called from the installer.

In particular:

* For `fish`, we need to run some commands interactively in addition to adding to shell configuration files
* Since it makes changes that cannot be previewed, it absolutely has to be as close to `pyenv init`-produced code as possible (possibly barring instructional comments about destination files since they are irrelevant in inserted code) -- preferably, sharing the payload code to avoid duplication
* In the installer, the shell detection code works incorrectly. It detects the shell that has invoked `pyenv init` -- which, in the case of the installer, is always `bash`. So in the installer, we'll probably be running it with `${SHELL##*/}` as an argument (it's supposed to be the user's default shell). Detection however is useful when running interactively.
  * So it needs to work both with and without the `<shell>` argument, and also act gracefully case of an unsupported shell.
* When run interactively, it would probably be useful to indicate failure with an exit code -- but that's not a priority. OTOH, shell setup failure should definitely not terminate the installer, only produce an error message. We'll likely run it with `|| true`.

macayu17:
Updated based on review feedback:

- renamed the mode from `--append` to `--install`
- made startup-file edits bail out if a file already contains Pyenv-related code
- stopped treating `grep` inspection errors as a successful no-op
- added Fish universal-variable setup before writing `config.fish`
- kept the inserted setup snippets shared with the existing init output
- added coverage for no-argument shell detection, unsupported shells, existing Pyenv config, and file inspection failures

Checks run:

- `PATH=./bats/bin:$PATH bats test/init.bats`
- `PATH=./bats/bin:$PATH bats test/help.bats test/commands.bats`
- `PATH="$PWD/libexec:$PWD/bin:$PATH" pyenv-help --usage init`
- `git diff --check`

libexec/pyenv-init
```diff
@@ -186,6 +194,149 @@ function help_() {
...
+  fi
+
+  for ((index = 0; index < ${#files[@]}; index++)); do
+    append_lines "${files[$index]}" "${lines[$index]}"
+  done
+}
+
+function check_startup_file() {
+  local file="$1"
+  local grep_status
```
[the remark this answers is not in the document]
native-api: :laughing: :laughing: :laughing: 
One AI complained about one solution, the other complains about the opposite one!

Since the installer is not supposed to run repeatedly, I'm fine with that.

native-api (approved):

Issue #3365, opened 2025-12-05
pyenv rehash hangs for 60 seconds in sandboxed environments (Landlock/seccomp)


vlasky:
# pyenv rehash hangs for 60 seconds in sandboxed environments (Landlock/seccomp)

## Problem

`pyenv rehash` waits up to 60 seconds before failing when run inside a sandboxed environment that restricts filesystem writes (e.g., Landlock on Linux, Seatbelt on macOS).

This affects users of:
- OpenAI Codex CLI (uses Landlock sandbox)
- Flatpak applications
- Firejail
- Docker with seccomp profiles
- Any environment using Linux Security Modules that restrict writes

## Root Cause

In `libexec/pyenv-rehash`, the writability check on line 36 uses shell's `-w` test:

```bash
if [ ! -w "$SHIM_PATH" ]; then
  echo "pyenv: cannot rehash: $SHIM_PATH isn't writable"
  exit 1
fi
```

The `-w` test only checks Unix file permissions, **not** kernel-level sandbox restrictions like Landlock or seccomp. The check passes, but the subsequent `acquire_lock` fails silently because the actual write is blocked by the sandbox.

Since `acquire_lock` fails, pyenv interprets this as "another process holds the lock" and enters a 60-second wait loop:

```bash
while (( SECONDS <= start + ${PYENV_REHASH_TIMEOUT:-60} )); do
  if acquire_lock 2>/dev/null; then
    acquired=1
    break
  else
    sleep 0.1 2>/dev/null || sleep 1
  fi
done
```

## Impact

When `eval "$(pyenv init -)"` runs in a user's `.profile` or `.bashrc`, it includes `command pyenv rehash 2>/dev/null`. In a sandboxed environment, every shell command hangs for 60 seconds before timing out.

This is particularly problematic because:
1. The `2>/dev/null` hides the error message
2. Users have no indication why commands are slow
3. Many modern dev tools use sandboxing by default

## Proposed Fix

Detect sandbox restrictions by attempting an actual test write before entering the wait loop:

```bash
if [ ! -w "$SHIM_PATH" ]; then
  echo "pyenv: cannot rehash: $SHIM_PATH isn't writable"
  exit 1
fi

# Detect sandbox restrictions that -w can't see (Landlock, seccomp, etc.)
SANDBOX_TEST_FILE="$SHIM_PATH/.pyenv-sandbox-test-$$"
if ! touch "$SANDBOX_TEST_FILE" 2>/dev/null; then
[...]

Issue #2893, opened 2024-01-31
pyenv: cannot rehash: ~/.pyenv/shims/.pyenv-shim exists on startup


rsantellan:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [X] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [X] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [X] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [X] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [X] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [X] Platform information (e.g. Ubuntu Linux 16.04): LMDE 6 (Linux Mint Debian Edition)
- [X] OS architecture (e.g. amd64): amd64
- [X] pyenv version:  pyenv 2.3.32
- [X] Python version: Python 3.11.2
- [X] C Compiler information (e.g. gcc 7.3):  gcc (Debian 12.2.0-14) 12.2.0
- [ ] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`. E.g. if you have a problem with installing Python, run `env PYENV_DEBUG=1 pyenv install -v <version> 2>&1 | tee trace.log` (note the `-v` option to `pyenv install`).
- [ ] If you have a problem with installing Python, please also attach `config.log` from the build directory
  * The build directory is reported after the "BUILD FAILED" message and is usually under `/tmp`.
- [ ] If the build succeeds but the problem is still with the build process (e.g. the resulting Python is missing a feature), please attach
  * the debug trace from reinstalling the faulty version with `env PYENV_DEBUG=1 pyenv install -f -k -v <version> 2>&1 | tee trace.log`
[...]

Issue #2829, opened 2023-10-27
Race condition when starting multiple terminals? pyenv: cannot rehash: .pyenv-shim exists


fsc-eriker:
This morning when I logged in after an OS update (MacOS Sonoma 14.1) I got `-bash: python: command not found` in all my restarted terminals.

In one of them, I could also see

```
pyenv: cannot rehash: /Users/fsc-eriker/.pyenv/shims/.pyenv-shim exists
```

Of course, the file no longer existed by the time I checked.

Running `pyenv rehash` in one of the terminals resolved the problem in all of them.

My suspicion is that there is a race condition when many terminals start up in quick succession after a reboot, though I have been running `pyenv` for several years and this is the first time I see this problem.

### Prerequisite
* [X] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [X] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [X] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [ ] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [ ] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
- [X] Platform information: MacOS 14.1
- [X] OS architecture (e.g. amd64): Intel
- [X] pyenv version: 2.3.17
- [X] Python version: 3.11.3 (among others)
- [ ] C Compiler information (e.g. gcc 7.3):  clang I guess? I have 15.0.0
- [ ] Please attach the debug trace of the failing command as a gist:
  * Run `env PYENV_DEBUG=1 <faulty command> 2>&1 | tee trace.log` and attach `trace.log`. E.g. if you have a problem with installing Python, run `env PYENV_DEBUG=1 pyenv install -v <version> 2>&1 | tee trace.log` (note the `-v` option to `pyenv install`).
- [ ] If you have a problem with installing Python, please also attach `config.log` from the build directory
[...]

PR #3469, opened 2026-06-03
fix(rehash): prevent terminal hang caused by stale or sandbox-blocked lock file

### LINKED ISSUE #3457: Intermittently hangs new terminal windows for ~30 s when doing rehash or other operations
I have uninstalled pyenv due to the above issue and now terminal windows open immediately. Please fix ASAP. It used to be great but now the performance is so bad I'd rather not use it at all.

anupddas:
## Problem

Closes #3457. Related: #3365, #2829, #2893.

`pyenv rehash` uses a file-based mutex (`$SHIM_PATH/.pyenv-shim`) to prevent concurrent runs. If that lock file is left behind by a previously killed/crashed rehash, every subsequent call spins for the full `PYENV_REHASH_TIMEOUT`
(default 60 s). This blocks `pyenv init -` in `.bashrc`/`.zshrc`, hanging new terminal windows for 30–60 s on startup.

The same hang occurs on Linux ≥ 6.14 (Landlock) and macOS Seatbelt environments where the `-w` writability test passes but actual writes are blocked by the kernel.

## Changes

**`libexec/pyenv-rehash`**
1. **Stale lock TTL**: lowered from 10 min → 2 min inside `acquire_lock()`
2. **Pre-loop stale sweep**: `find -mmin +2` runs before the retry loop — stale locks are removed instantly without waiting
3. **Pre-loop sandbox check**: `mktemp`-based write test before the loop so sandboxed environments exit immediately with a clear error

**`test/rehash-stale-lock.bats`** — new bats tests covering all three scenarios

## Testing

```bash
# Run the new tests
bats test/rehash-stale-lock.bats

# Manual regression test
touch -d "3 minutes ago" ~/.pyenv/shims/.pyenv-shim
time pyenv rehash  # should complete in < 3 s, not hang 30–60 s
```

---
## Summary by cubic
Fixes terminal hangs caused by stale or sandbox-blocked rehash lock files. `pyenv rehash` now removes >2 min stale locks before the first write, fails fast in sandboxed `shims` dirs, and sets cleanup traps ASAP to avoid leftovers.

- **Bug Fixes**
  - Consolidated in `acquire_lock()`: remove stale lock files (>2 min) with errors ignored; run a one-time `mktemp` check to exit immediately when `shims` isn’t writable (Landlock/Seatbelt); set `trap EXIT` right after locking; simplify loop else-branch to sleep only; `release_lock` unsets the trap.
  - Tests: add stale-lock clearance case in `test/rehash.bats`; use portable `touch -t` with docs; remove a redundant test.

<sup>Written for commit 9e6006e66f79c4c84dc25b21b4164579d0526872. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libexec/pyenv-rehash
```diff
@@ -39,6 +39,19 @@ if [ ! -w "$SHIM_PATH" ]; then
   exit 1
 fi
 
+# Remove any stale lock file (> 2 min old) before entering the wait loop.
+# A healthy rehash completes in well under 2 minutes on any normal system.
+# This prevents a full 60 s hang caused by a lock left by a previously crashed process.
+find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \; 2>/dev/null || true
```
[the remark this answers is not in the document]
native-api: We assume that a rehash completes well beyond 2 minutes. This is tightening of the previous assumption of 10 minutes.

libexec/pyenv-rehash
```diff
@@ -21,7 +21,7 @@ acquire_lock() {
   # Assuming an old lockfile is stale
   # Unknown why this happens for some users but this at least unblocks them
   last_acquire_error="$( { ( echo -n > "$PROTOTYPE_SHIM_PATH"; ) 2>&1 1>&3 3>&1-; } 3>&1)" \
-    || { find "$PROTOTYPE_SHIM_PATH" -mmin +10 -exec rm -f {} \; ; ret=1; }
+    || { find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \; ; ret=1; }
```
native-api: This, and the code that this duplicates. should probably be refactored into `acquire_lock`.

libexec/pyenv-rehash
```diff
@@ -39,6 +39,19 @@ if [ ! -w "$SHIM_PATH" ]; then
...
+# This prevents a full 60 s hang caused by a lock left by a previously crashed process.
+find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \; 2>/dev/null || true
+
+# Verify actual write capability before the loop (catches Landlock/Seatbelt
+# sandbox restrictions that fool the -w test above).
+if ! _tmpfile="$(TMPDIR="$SHIM_PATH" mktemp 2>/dev/null)" || \
+   ! rm -f "$_tmpfile" 2>/dev/null; then
+  echo "pyenv: cannot rehash: $SHIM_PATH isn't writable" >&2
+  exit 1
+fi
```
native-api: Likewise, this duplicates another nontrivial code. AFAICS, it also belongs in `acquire_lock` if the intention is to call it before every lockfile acquisition attempt.

native-api (changes requested):

anupddas (commented):
Thanks for the review, @native-api . Both issues are resolved in the latest commit (`fix(rehash): consolidate stale lock and sandbox checks into acquire_lock()`).

---

**1. Pre-loop `find` sweep → consolidated into `acquire_lock()`**

The standalone pre-loop block has been removed. The stale-lock cleanup now lives exclusively inside `acquire_lock()` with the TTL lowered from 10 min to 2 min:

```bash
acquire_lock() {
  local ret
  set -o noclobber
  last_acquire_error="$( { ( echo -n > "$PROTOTYPE_SHIM_PATH"; ) 2>&1 1>&3 3>&1-; } 3>&1)" \
    || {
      find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \;
      ...
      ret=1
    }
  set +o noclobber
  [[ -z "${ret}" ]]
}
```

---

**2. Pre-loop `mktemp` check → moved into `acquire_lock()`**

The sandbox writability check is now inside `acquire_lock()`, guarded by `tested_for_other_write_errors` so it runs at most once and exits immediately on failure. The duplicate copy has been removed from the loop's `else` branch:

```bash
      if [[ -z $tested_for_other_write_errors ]]; then
        ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm "$t" ) \
          && tested_for_other_write_errors=1 \
          || { echo "pyenv: cannot rehash: $SHIM_PATH isn't writable" >&2
               set +o noclobber
               exit 1; }
      fi
```

The loop's `else` branch is now `sleep` only. No behaviour change for the normal concurrent-rehash path.

Happy to adjust further if anything looks off.

libexec/pyenv-rehash
anupddas: Thank you — you're right that this was duplication. The pre-loop sweep has been removed. The TTL tightening (10 min → 2 min) and stale cleanup are now consolidated inside acquire_lock() only, alongside the sandbox check. Both concerns now live in one place per your suggestion.

libexec/pyenv-rehash
anupddas: Agreed — moved into acquire_lock(). It runs on the first acquisition failure, guarded by the existing tested_for_other_write_errors flag so it only fires once and exits immediately on sandbox restriction rather than retrying. The duplicate block has been removed from the loop's else branch.

libexec/pyenv-rehash
```diff
@@ -13,15 +13,24 @@ mkdir -p "$SHIM_PATH"
...
   local ret
   set -o noclobber
-  # Assuming an old lockfile is stale
-  # Unknown why this happens for some users but this at least unblocks them
   last_acquire_error="$( { ( echo -n > "$PROTOTYPE_SHIM_PATH"; ) 2>&1 1>&3 3>&1-; } 3>&1)" \
-    || { find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \; ; ret=1; }
+    || {
+      # Remove stale lock file if older than 2 minutes.
+      # No healthy rehash takes this long; cleanup prevents hangs from crashed processes.
+      find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \;
```
native-api: Shouldn't we delete a stale lockfile _before_ trying to create our own?

This is how it was in your original logic. too.

anupddas: Good catch, thanks. Moved the `find` cleanup to the top of `acquire_lock()`, before the noclobber write attempt:

```bash
acquire_lock() {
  local ret
  # Remove stale lock file (> 2 min) before attempting to create our own.
  # No healthy rehash takes this long; avoids a failed write on the first attempt.
  find "$PROTOTYPE_SHIM_PATH" -mmin +2 -exec rm -f {} \;
  set -o noclobber
  last_acquire_error="$( { ( echo -n > "$PROTOTYPE_SHIM_PATH"; ) 2>&1 1>&3 3>&1-; } 3>&1)" \
    || {
      if [[ -z $tested_for_other_write_errors ]]; then
        ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm "$t" ) \
          && tested_for_other_write_errors=1 \
          || { echo "pyenv: cannot rehash: $SHIM_PATH isn't writable" >&2
               set +o noclobber
               exit 1; }
      fi
      ret=1
    }
  set +o noclobber
  [[ -z "${ret}" ]]
}
```

Previously the `find` was in the `||` failure branch, so the sequence was: try write → fail against stale file → clean up → return retry. That wasted one full loop iteration. With the cleanup before the noclobber attempt, a stale lock is removed first and the write succeeds immediately on the same call.

The sandbox `mktemp` check stays in the failure branch — it only makes sense to run it when the write has already failed.

anupddas (commented):
The only structural change from the previous version is that find moves from the || failure branch to the top of the function body, before set -o noclobber. The sandbox check stays in the failure branch where it belongs — it only makes sense to run it when the write has already failed.

test/rehash-stale-lock.bats
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+setup() {
+  export PYENV_ROOT="${BATS_TMPDIR}/pyenv-test"
+  mkdir -p "${PYENV_ROOT}/shims"
+  PROTOTYPE="${PYENV_ROOT}/shims/.pyenv-shim"
+}
```
native-api: Clashes with and seems superfluous to `setup()` in `test_helper`

test/rehash-stale-lock.bats
```diff
@@ -0,0 +1,36 @@
...
+
+teardown() {
+  rm -rf "${PYENV_ROOT}"
+}
```
native-api: Not needed: `setup()` from `test_helper` creates directories under `BATS_TEST_TMPDIR` which is recreated for each test.

native-api (changes requested):
The tests overlap with existing ones

anupddas:
> The tests overlap with existing ones

Fixed in the latest two commits:

- Deleted `test/rehash-stale-lock.bats` entirely.
- Removed the overlapping "unwritable shims" and timing tests — those are already covered by the existing "non-writable shims directory" and "rehash in progress" tests in `rehash.bats`.
- Added one genuinely new test directly to `test/rehash.bats`, immediately after "rehash in progress", with no custom `setup()`/`teardown()` — relying on `test_helper` as the rest of the file does:

```bash
@test "rehash clears stale lock file older than 2 minutes and succeeds" {
  mkdir -p "${PYENV_ROOT}/shims"
  touch -d "3 minutes ago" "${PYENV_ROOT}/shims/.pyenv-shim"
  run pyenv-rehash
  assert_success
}
```

This is the only scenario without existing coverage: a fresh lock (< 2 min) should still block, while a stale lock (> 2 min) should be cleared and the rehash should succeed.

test/rehash.bats
```diff
@@ -29,6 +29,13 @@ pyenv: cannot rehash: couldn't acquire lock ${PYENV_ROOT}/shims/.pyenv-shim for
 !
 }
 
+@test "rehash clears stale lock file older than 2 minutes and succeeds" {
+     mkdir -p "${PYENV_ROOT}/shims"
+     touch -d "3 minutes ago" "${PYENV_ROOT}/shims/.pyenv-shim"
```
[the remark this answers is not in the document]
anupddas: > P2: Uses a non-portable relative-date `touch -d` invocation in test setup; BSD/BusyBox environments may reject it, so the test can fail before exercising `pyenv-rehash`.
> 
> Prompt for AI agents
> ```
> Check if this issue is valid — if so, understand the root cause and fix it. At test/rehash.bats, line 34:
> 
> <comment>Uses a non-portable relative-date `touch -d` invocation in test setup; BSD/BusyBox environments may reject it, so the test can fail before exercising `pyenv-rehash`.</comment>
> 
> <file context>
> @@ -29,6 +29,13 @@ pyenv: cannot rehash: couldn't acquire lock ${PYENV_ROOT}/shims/.pyenv-shim for
>  
> +@test "rehash clears stale lock file older than 2 minutes and succeeds" {
> +     mkdir -p "${PYENV_ROOT}/shims"
> +     touch -d "3 minutes ago" "${PYENV_ROOT}/shims/.pyenv-shim"
> +     run pyenv-rehash
> +     assert_success
> </file context>
> ```
> 
> ✅ Addressed in [`e4f855e`](https://github.com/pyenv/pyenv/commit/e4f855efc518ec9c10f84125f2dbe972648b1089)

Valid flag. `touch -d "3 minutes ago"` is GNU coreutils-only and fails on macOS/BSD and BusyBox.

Fixed in the latest commit by replacing it with POSIX `touch -t`:

```bash
touch -t 200001010000.00 "${PYENV_ROOT}/shims/.pyenv-shim"
```

This sets mtime to 2000-01-01 00:00:00 — always older than 2 minutes — and is supported on GNU coreutils, BSD/macOS, and BusyBox without any shell arithmetic or platform detection.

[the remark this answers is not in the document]
native-api: I researched this topic when adding the previous version of the test, even asked at StackExchange: https://unix.stackexchange.com/questions/806015/portably-set-a-files-time-to-n-minutes-ago -- and a portable version of the code is very unwieldy.

So what the AI suggests -- to use a fixed timestamp far in the past -- is indeed by far the simplest solution for our use case.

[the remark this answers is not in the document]
anupddas: Thanks for the clarification!

native-api (approved):

anupddas:
Thanks @native-api

PR #3475, opened 2026-06-11
Fix #3474: CI: add_version enhancements

### LINKED ISSUE #3474: CI: add_version: copy patches, detect existing auto PRs, don't bump OpenSSL over major versions
Subj.

[ ] Copy patches to the new version
  * only within a version line?
 
[ ] Detect existing PRs for some of the detected versions
  * To prevent cases like https://github.com/pyenv/pyenv/pull/3473 : 2 releases are done upstream in quick succession, PR for one 1 is created, the 2nd check adds a PR for both

[ ] Don't update OpenSSL to a next major version?

macayu17:
### Prerequisite
* [x] Considered implementing the feature as a hook script or plugin first.
  * pyenv provides powerful support for plugins and hook scripts. See the Authoring Plugins guide for details. This change was evaluated with plugin compatibility in mind.
* [x] Considered contributing the patch upstream to rbenv.
  * Since pyenv periodically imports changes from rbenv and aims to maintain compatibility, core modifications were kept minimal.
* [x] This PR addresses the following issue:
  * Closes #3474

### Description
This PR implements the three improvements requested for the `add_cpython.py` CI script:
1. **Prevent OpenSSL major-version upgrades**
   * `get_store_latest_release()` now accepts a major-version prefix, ensuring older Python build definitions are not unintentionally upgraded across OpenSSL major versions (for example, from `1.1.1` to `3.x`).
2. **Detect existing automated PRs**
   * The script now checks:
     ```bash
     git ls-remote origin "refs/heads/auto_add_version/*"
     ```
   * This prevents duplicate pull requests when multiple upstream releases are published in quick succession (as observed in #3473).
3. **Copy patches for patch-version updates**
   * When creating a new patch release (for example, `3.12.2` → `3.12.3`), the script automatically copies the existing `patches/` directory using `shutil.copytree()` when both versions belong to the same release line.
   * For prerelease upgrades, it falls back to `git mv` to preserve file history.

### Tests
* [x] No additional unit tests were added.
  * This change modifies the maintainer-only GitHub Actions automation script (`add_cpython.py`) and does not affect pyenv core functionality or shell commands.

---
## Summary by cubic
Makes the `add_cpython.py` CI script safer by locking OpenSSL to the same major, skipping versions with pending auto PRs, and preserving patches when adding new releases. Improves reliability with paginated release lookups, stricter parsing, and a fixed OpenSSL cached lookup. Addresses #3474.

- **Bug Fixes**
  - OpenSSL: constrain updates to the same major parsed from the install line; use paginated GitHub release lookup with tag validation; skip drafts/prereleases; add request timeouts; fix cached lookup to honor the major and return the latest matching release.
  - Pending PRs: exclude versions already present in `auto_add_version/*` branches via `git ls-remote` with strict output parsing.
  - Patches: for same-series releases, copy patches for patch bumps and `git mv` for prereleases; update inner `Python-<ver>` paths and `*t` symlinks.

<sup>Written for commit bd944510f1fad5ee95c878ce80991486cb09de9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3475?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/python-build/scripts/add_cpython.py
+++ b/plugins/python-build/scripts/add_cpython.py
@@ -488,23 +488,38 @@ class OpenSSLVersionsDirectory(KeyedList[_OpenSSLVersionInfo, packaging.version.
             #already retrieved
             return self[max(self.keys())]
 
-        response = requests.get("https://api.github.com/repos/openssl/openssl/releases", timeout=30)
-        response.raise_for_status()
-        j = response.json()
-        if not isinstance(j, list):
-            raise ValueError(f"Unexpected GitHub API response: {j}")
+        url = "https://api.github.com/repos/openssl/openssl/releases?per_page=100"
+        releases = []
+        releases_prefix = []
+        
+        while url:
+            response = requests.get(url, timeout=30)
+            response.raise_for_status()
+            j = response.json()
+            if not isinstance(j, list):
+                raise ValueError(f"Unexpected GitHub API response: {j}")
+                
+            page_releases = [r for r in j if not r.get('prerelease')]
+            releases.extend(page_releases)
             
-        releases = [r for r in j if not r.get('prerelease')]
-        if not releases:
+            if prefix:
+                releases_prefix = [r for r in page_releases if r.get('tag_name', '').startswith(f"openssl-{prefix}")]
+                if releases_prefix:
+                    break
+            elif releases:
+                break
+                
+            if 'next' in response.links:
+                url = response.links['next']['url']
+            else:
+                url = None
+
+        if prefix and not releases_prefix:
+            raise ValueError(f"No OpenSSL release found matching prefix {prefix}")
+        if not releases and not prefix:
             raise ValueError("No non-prerelease OpenSSL versions found")
             
-        if prefix:
-            releases_prefix = [r for r in releases if r.get('tag_name', '').startswith(f"openssl-{prefix}")]
-            if not releases_prefix:
-                raise ValueError(f"No OpenSSL release found matching prefix {prefix}")
-            releases = releases_prefix
-        
-        j_release = releases[0]
+        j_release = releases_prefix[0] if prefix else releases[0]
 
         # noinspection PyTypeChecker
         # urlparse can parse str as well as bytes
--- a/plugins/python-build/scripts/add_cpython.py
+++ b/plugins/python-build/scripts/add_cpython.py
@@ -17,6 +17,7 @@ import os.path
 import pathlib
 import pprint
 import re
+import shutil
 import subprocess
 import sys
 import typing
@@ -45,6 +46,14 @@ EXCLUDED_VERSIONS= {
 
 here = pathlib.Path(__file__).resolve()
 OUT_DIR: pathlib.Path = here.parent.parent / "share" / "python-build"
+
+AUTO_ADD_VERSION_REF_RE = re.compile(
+    r"^(?P<object_id>[0-9a-f]{40,64})\t"
+    r"refs/heads/auto_add_version/(?P<versions>\S+)$"
+)
+OPENSSL_RELEASE_TAG_RE = re.compile(
+    r"^openssl-(?P<major>\d+)\.\d+(?:\.\d+)*(?:[a-z]+\d*)?$"
+)
 
 T_THUNK=\
 '''export PYTHON_BUILD_FREE_THREADING=1
@@ -83,15 +92,13 @@ def adapt_script(version: packaging.version.Version,
                                  url=new_package_url+'#'+new_package_hash,
                                  verify_py_suffix=verify_py_suffix)
 
-        elif m:=re.match(r'\s*install_package\s+"(?P<package>openssl-\S+)"\s+'
-                       r'"(?P<url>\S+)"\s.*$',
+        elif m:=re.match(r'\s*install_package\s+'
+                         r'"(?P<package>openssl-(?P<openssl_major>\d+)\.\S+)"\s+'
+                         r'"(?P<url>\S+)"\s.*$',
                          line):
-            existing_version_str = m.group('package').split('-')[1]
-            if existing_version_str.startswith('1.1.1'):
-                prefix = '1.1.1'
-            else:
-                prefix = '.'.join(existing_version_str.split('.')[:2])
-            item = VersionDirectory.openssl.get_store_latest_release(prefix)
+            item = VersionDirectory.openssl.get_store_latest_release(
+                int(m.group('openssl_major'))
+            )
 
             line = Re.sub_groups(m,
                                package=item.package_name,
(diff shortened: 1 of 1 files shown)


plugins/python-build/scripts/add_cpython.py
```diff
@@ -129,6 +134,18 @@ def add_version(version: packaging.version.Version):
...
+        prev_patches = OUT_DIR / "patches" / str(previous_version)
+        new_patches = OUT_DIR / "patches" / str(version)
+        if prev_patches.exists() and not new_patches.exists():
+            if is_prerelease_upgrade:
+                logger.info(f"Git moving patches from {previous_version} to {version}")
+                subprocess.check_call(("git", "-C", OUT_DIR, "mv", f"patches/{previous_version}", f"patches/{version}"))
+            else:
+                logger.info(f"Copying patches from {previous_version} to {version}")
+                shutil.copytree(prev_patches, new_patches)
+
```
native-api: Need to also rename the package name directory under `patches/<version>`
And replicate a symlink for the `t` version

plugins/python-build/scripts/add_cpython.py
```diff
@@ -129,6 +134,18 @@ def add_version(version: packaging.version.Version):
...
     VersionDirectory.existing.append(_CPythonExistingScriptInfo(version,str(new_path)))
 
+    if previous_version.major == version.major and previous_version.minor == version.minor:
+        import shutil
+        prev_patches = OUT_DIR / "patches" / str(previous_version)
+        new_patches = OUT_DIR / "patches" / str(version)
+        if prev_patches.exists() and not new_patches.exists():
+            if is_prerelease_upgrade:
+                logger.info(f"Git moving patches from {previous_version} to {version}")
+                subprocess.check_call(("git", "-C", OUT_DIR, "mv", f"patches/{previous_version}", f"patches/{version}"))
```
native-api: I don't think we ever had patches for a prerelease -- but good foresight!
<sub>(checking the history -- we did once, [for 3.2.6rc1](https://github.com/pyenv/pyenv/commit/6790e3e18a2402406c4d1beee87d629d11ab4529))</sub>

plugins/python-build/scripts/add_cpython.py
```diff
@@ -211,7 +228,28 @@ def main():
...
+                continue
+            branch = parts[1].replace("refs/heads/auto_add_version/", "")
+            for v_str in branch.split("_"):
+                try:
+                    pending_versions.add(packaging.version.Version(v_str))
+                except ValueError:
+                    pass
+    except Exception as e:
+        logger.warning(f"Failed to fetch pending PR branches: {e}")
+
```
native-api: The AI warned against swallowing exceptions -- but that is exactly what this code does!

Is there a reason why we should ignore a fault like this rather than let it crash the program?

plugins/python-build/scripts/add_cpython.py
```diff
@@ -211,7 +228,28 @@ def main():
...
+        for line in ls_remote.splitlines():
+            parts = line.split("\t", 1)
+            if len(parts) < 2:
+                continue
+            branch = parts[1].replace("refs/heads/auto_add_version/", "")
+            for v_str in branch.split("_"):
+                try:
+                    pending_versions.add(packaging.version.Version(v_str))
+                except ValueError:
+                    pass
```
native-api: What's the reason for swallowing this silently?

plugins/python-build/scripts/add_cpython.py
```diff
@@ -211,7 +228,28 @@ def main():
...
+        ls_remote = subprocess.check_output(
+            ["git", "-C", OUT_DIR, "ls-remote", "origin", "refs/heads/auto_add_version/*"],
+            text=True,
+            timeout=30
+        )
+        for line in ls_remote.splitlines():
+            parts = line.split("\t", 1)
+            if len(parts) < 2:
+                continue
+            branch = parts[1].replace("refs/heads/auto_add_version/", "")
```
native-api: This ad-hoc handling makes the expected format of the input very unclear.
Can we properly parse it into meaningful parts (e.g. with a regex)?

(OTOH `branch.split()` is okay 'cuz regex cannot handle delimited lists.)

Wrong-formatted output from `git ls-remote` is something unexpected, I doubt we should swallow it.

plugins/python-build/scripts/add_cpython.py
```diff
@@ -211,7 +228,28 @@ def main():
         VersionDirectory.available.get_store_available_source_downloads(release, True)
         del release
 
-    versions_to_add = sorted(VersionDirectory.available.keys() - VersionDirectory.existing.keys())
+    import subprocess
+    pending_versions = set()
```
native-api: Looks like this logic should be in a function 'cuz it's a separate concern.

plugins/python-build/scripts/add_cpython.py
```diff
@@ -211,7 +228,28 @@ def main():
         VersionDirectory.available.get_store_available_source_downloads(release, True)
         del release
 
-    versions_to_add = sorted(VersionDirectory.available.keys() - VersionDirectory.existing.keys())
+    import subprocess
```
native-api: PEP8 violation

plugins/python-build/scripts/add_cpython.py
```diff
@@ -440,18 +478,54 @@ class _OpenSSLVersionInfo(typing.NamedTuple):
...
+                
+            if 'next' in response.links:
+                url = response.links['next']['url']
+            else:
+                url = None
+
+        if prefix and not releases_prefix:
+            raise ValueError(f"No OpenSSL release found matching prefix {prefix}")
+        if not releases and not prefix:
+            raise ValueError("No non-prerelease OpenSSL versions found")
```
native-api: First, it retrieves all the releases ever every time, only to take just the first one found in the end.

Again, makes the expected structure very unclear.
Can probably be refined into a bunch of list comprehensions.

No need to handle every error separately, it should be enough to let the data retrieval fail. Printing a relevant part of the faulty data, too, would be useful for diagnostics but not necessary for now 'cuz presumably, you can just get it yourself by doing the same request locally.

plugins/python-build/scripts/add_cpython.py
```diff
@@ -86,7 +86,12 @@ def adapt_script(version: packaging.version.Version,
         elif m:=re.match(r'\s*install_package\s+"(?P<package>openssl-\S+)"\s+'
                        r'"(?P<url>\S+)"\s.*$',
                          line):
-            item = VersionDirectory.openssl.get_store_latest_release()
+            existing_version_str = m.group('package').split('-')[1]
```
native-api: Could avoid ad-hoc'ing by adding more groups to the regex

plugins/python-build/scripts/add_cpython.py
```diff
@@ -86,7 +86,12 @@ def adapt_script(version: packaging.version.Version,
...
+            if existing_version_str.startswith('1.1.1'):
+                prefix = '1.1.1'
```
native-api: Does this warrant a special case? We only need the latest major release.

Will this even be ever triggered? AFAICS we have no non-EOL version lines that use OpensSSL 1.

native-api (changes requested):
The code seems operational but rather sloppy at times.
I hope it's not an AI slop 'cuz I surely spent much more time reviewing it that an AI would generating it!

macayu17:
The earlier version was overcomplicated. I’ve reworked it based on your review. Sorry for the noise.

native-api (commented):
Now that's a different talk! Much better! 👍

I [forked the PR branch](https://github.com/native-api/pyenv/tree/fix/issue-3474) and [did some testing](https://github.com/native-api/pyenv/actions/workflows/add_version.yml) which has revealed a bunch of problems.

Some look unrelated to the current changes:
* `ftpmirror.gnu.org` sometimes produces a 5xx
* Unhandled exceptions do not cause a run failure -- looks like `sys.excepthook` has no effect in Github Actions (I tested that earlier when implemeting it, so maybe it's due to some later changes in Actions)

But [the latest one](https://github.com/native-api/pyenv/actions/runs/27471463760/job/81203032671) seems relevant:
* `KeyError` trying to retrieve an OpenSSL download info from the directory object. Looks like it's not being added correctly.

Could you look into that?

I'll try to address the unrelated ones and will look into the revelant one after that if you won't have done so by then.

native-api (approved):
LGTM. Cannot test actual PR handling now, it requires to install our private Github App to my fork which I cannot find how to do.
So I'll add fixes for unrelated issues separately.

Issue #3477, opened 2026-06-19
Release: add recent sponsor information to autogenerated release notes


native-api:
**Describe the intended use case**
* What your general environment is if it's relevant to the feature request and is not a generic console with a typical Pyenv installation (CI, server with a custom setup, cloud environment, IDE)

Pyenv Github

* What you are trying to achieve

Create a new release

* What specifically you are doing for that regarding Pyenv

Generate release notes with the stock button

* Where you are stuck

The autogenerated info has no info on sponsors

**Describe the solution you'd like**

Add a section on recent sponsors from Guthub Sponsors AND from OpenCollective directly.

* At least: since the last release or within the last month (=since the same day of the previous month or the end of the previous month if it's shorter), whichever is longer
  * The preferred phrasing looks like: "Sponsors since &lt;date&gt;"
  * [ ] Check the sponsorship stats to see if there's a need to pick a longer period at some points so that there is always someone to mention
* No mentioning of tiers for now
  * We'll probably reconfigure tiers later into something reasonable. So mentioning them now only to reconfigure shortly after would look like moving the goalposts.

**Describe alternatives you've considered**

Generating by hand is labor-intensive, inconvenient and error-prone.

**Additional context**

Plan:

* [ ] check if there's a way to customize Github release notes generation
* check if there's API to get sponsor information
  * [ ] from Github Sponsors
  * [ ] from OpenCollective
* [ ] write a script that would generate the sponsors section in ReST format using the above API
* [ ] integrate it with the "Generate release notes" Github button if customization is possible; if not, add running it as a step in MAINTENANCE.md

Issue #2334, opened 2022-04-22
Provide prebuilt CPython for common environments


fourlastor:
Too many issues will kill our team's development velocity, drastically.
Make sure you have checked all steps below.

### Prerequisite
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues). You should look for closed issues, too.
* [x] Make sure you are not asking us to help solving your specific issue.
  * GitHub issues is opened mainly for development purposes. If you want to ask someone to help solving your problem, go to some community site like [Gitter](https://gitter.im/yyuu/pyenv), [StackOverflow](https://stackoverflow.com/questions/tagged/pyenv), etc.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
  * Please refer to the package documentation for the installation issues, etc.
* [x] Make sure your problem is not derived from plugins.
  * This repository is maintaining `pyenv` and the default `python-build` plugin only. Please refrain from reporting issues of other plugins here.

### Description
Hello! I recently installed pyenv on Linux and found the whole process of downloading+compiling the source code somewhat complicated, so I created a [workflow](https://github.com/fourlastor/python-builds/blob/main/.github/workflows/main.yml) that does it for me (note: there is a `Dockerfile` which I initially used to build python locally, it's not actually used in the workflow). 

I thought it could be nice to have this on each release of pyenv, so that people can (if they want to) download a precompiled binary and not have to install python build dependencies. Is this something you'd be interested in? In case I can make a PR :slightly_smiling_face:

native-api:
Could you please describe your use case? This sounds like a strange thing to be causing a serious UX problem.

Installing build dependencies is a one-time step so this shouldn't be a problem.
Source releases are much more widely usable since they are not tied to a specific environment. They also have to be hosted somewhere. So it's more effort and more problems for questionable gain. And for you as a user, that shouldn't make a difference, either, because you just type a short invocation, Pyenv does all the heavy lifting for you.

fourlastor:
I agree, this isn't a serious UX issue, it's a nice to have.

My use case is the sporadic use as a user (instead of say a CI server), where I rather not have build dependencies installed on my desktop machine and keep everything as contained as possible (so for example I just need to delete the `.pyenv` folder in my home and everything is gone after using).

[...]

PR #3478, opened 2026-06-21
Add script to generate sponsors section for release notes

macayu17:
### Prerequisite
- [x] Please consider implementing the feature as a hook script or plugin as a first step.
  - This is a release-maintainer utility, not a user-facing pyenv command, so a script under `.github/scripts` is the simplest fit.
- [x] Please consider contributing the patch upstream to [rbenv](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.
  - This is specific to pyenv's release process and sponsorship setup.
- [x] My PR addresses the following pyenv issue (if any)
  - Addresses #3477

### Description
Adds `.github/scripts/generate_release_notes_sponsors.py`, a release-maintainer script that prints a Markdown sponsors section for GitHub Releases.

The script lists new GitHub Sponsors and OpenCollective backers created on or after the earlier of the latest release date or one month ago. It excludes the OpenCollective `github-sponsors` aggregate entry to avoid listing GitHub Sponsors twice.

`MAINTENANCE.md` now includes the script as a release checklist step.

### Tests
   - This is a standalone release utility that calls external APIs, so I verified it manually:
     - `python -m py_compile .github/scripts/generate_release_notes_sponsors.py`
     - `python .github/scripts/generate_release_notes_sponsors.py --help`
     - `python .github/scripts/generate_release_notes_sponsors.py --since 2026-05-21`
       - Verified clean failure when `gh` lacks the required `read:user` scope.

### DIFF
--- a/.github/scripts/generate_release_notes_sponsors.py
+++ b/.github/scripts/generate_release_notes_sponsors.py
@@ -176,22 +176,41 @@ def github_sponsors(since: datetime.date) -> typing.List[typing.Dict]:
             raise SponsorDataError(
                 "GitHub Sponsors query returned an unexpected response shape."
             ) from exc
-        for node in sponsorships["nodes"]:
-            created = datetime.datetime.fromisoformat(
-                node["createdAt"].replace("Z", "+00:00")
-            )
+        try:
+            nodes = sponsorships["nodes"]
+            page_info = sponsorships["pageInfo"]
+        except (TypeError, KeyError) as exc:
+            raise SponsorDataError(
+                "GitHub Sponsors query returned incomplete pagination data."
+            ) from exc
+
+        for node in nodes:
+            try:
+                created = datetime.datetime.fromisoformat(
+                    node["createdAt"].replace("Z", "+00:00")
+                )
+                entity = node["sponsorEntity"]
+                login = entity["login"]
+            except (AttributeError, KeyError, TypeError, ValueError) as exc:
+                raise SponsorDataError(
+                    "GitHub Sponsors query returned an unexpected sponsor record."
+                ) from exc
             if created < since_dt:
                 return sponsors
-            entity = node["sponsorEntity"]
             sponsors.append({
-                "login": entity["login"],
-                "name": entity.get("name") or entity["login"],
+                "login": login,
+                "name": entity.get("name") or login,
             })
 
-        page_info = sponsorships["pageInfo"]
-        if not page_info["hasNextPage"]:
+        try:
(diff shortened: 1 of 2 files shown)


native-api:
As per https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes, it looks impossible to customize the "generate release notes" button with custom code.

So it looks like all we can do is to edit release notes after the fact with an action workflow [triggered on `release`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release).

native-api (commented):
Looks good for a start but doesn't seem very usable as of now.

Tried to play with it, and all that it says is:

```
>gh auth status
github.com
  ✓ Logged in to github.com account native-api (keyring)
  - Active account: true
  - Git operations protocol: https
  - Token: ghp_************************************
  - Token scopes: 'read:org', 'read:user', 'repo', 'workflow'

>python generate_release_notes_sponsors.py --since 1970-01-01
error: HTTP Error 403: Forbidden

>python generate_release_notes_sponsors.py
error: HTTP Error 403: Forbidden
```

It doesn't even say where exactly the error is.
And I see lots of swallowed exceptions in code.

native-api:
Okay, I've debugged the code (because error messages produce no diagnostic information whatsoever) and found out that instead of an API response, I'm getting Cloudflare's challenge page -- with no way to complete the challenge.
Tried to use an API token -- no difference.

I've contancted OpenCollective support asking what my options are. But for now, I have no way to run the script to completion.

Could you make the OpenCollective step optional?

native-api:
OpenCollective did not reply yet on their stance about the challenge page (but it's weekend for them) -- but with an ability to make the query with the browser and pass the challenge, I can then supply the result to the script by hand.
Thank you!

native-api:
We can pay out, say, 15$ for the last two PRs if you're interested (because adding new versions ASAP is a user-requested feature). For that, you need to register at opencollective.com and provide your handle name for us to set up the payment.

The most important feature request right now is https://github.com/pyenv/pyenv/issues/2334. For that, a total reward would be several hundred.

Performance issues for Pyenv and Pyenv-Virtualenv are probably the second most important topic. We solved the most pressing ones in the previous months but some users claim that some still remain.

Implementing other highly requested features (you can order issues by the number of reactions) would also be eligible for a payout.

macayu17:
Thanks, I registered on OpenCollective. My handle is `macayu17`  .
I'll take a look at #2334 and the high-reaction issues before picking the next one.

PR #3483, opened 2026-06-26
rehash: fix race condition in landlock writability check

### LINKED ISSUE #3482: Race condition in pyenv-rehash writable check when many shells start concurrently
### Prerequisites
* [x] Make sure your problem is not listed in [the common build problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems).
* [x] Make sure no duplicated issue has already been reported in [the pyenv issues](https://github.com/pyenv/pyenv/issues?q=is%3Aissue). For build errors, a reported issue typically mentions a key error message. This key error message is often not in the 10 last build log lines reported to the console but is rather earlier in the build log -- typically, it's the first error message encountered in the log.
* [x] Make sure you are reporting a problem in Pyenv and not seeking consultation with Pyenv usage.
* [x] Make sure your problem is not derived from packaging (e.g. [Homebrew](https://brew.sh)).
* [x] Make sure your problem is not derived from plugins.

### Describe the bug
I use `tmux` to run many shells, and restore the session state after rebooting the machine using `tmux-resurrect`.

When the session is restored, more than 50 shells may be started concurrently. In rare cases, some of them fail at the writable check inside `acquire_lock`.

The output in the shell seems to be the following:
```
rm: cannot remove '/home/ec2-user/.pyenv/shims/tmp.3y7AMLWeI2': No such file or directory
pyenv: cannot rehash: /home/ec2-user/.pyenv/shims isn't writable
```

The problematic code seems to be the following:
```bash
( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm "$t" )
```

After `mktemp` creates a temporary file, another concurrently running `pyenv-rehash` process may remove that file before this process reaches `rm "$t"`. In that case, `rm` fails, and the writable check incorrectly reports that `$SHIM_PATH` is not writable.
From what I was able to observe, the temporary file was removed by `remove_stale_shims`. It may also be possible that `remove_outdated_shims` removes it as well.

#### Reproduction steps
The issue is rare, but I was able to reproduce it occasionally by starting around 60 shells concurrently while running tmux 3.6a:

```bash
for n in $(seq 1 60); do; tmux split-window -h -p 99 &; done
```

Depending on machine performance and timing, the issue may or may not occur. In my environment, around 1 or 2 shells sometimes failed.
To make the race condition easier to reproduce, I added a wait between `mktemp` and `rm`. With this change, I was able to reproduce the issue consistently:

```bash
( t="$(TMPDIR="$SHIM_PATH" mktemp)" && echo 'start sleeping' && sleep 10 && echo 'finish sleeping' && rm "$t" ) \
```

Then start two shells concurrently:
```
for n in $(seq 1 2); do; tmux split-window -h &; done
```

#### Diagnostic details
- [x] Platform information: Ubuntu 24.04.4 LTS (Also reproduce in Amazon Linux 2023)
- [x] OS architecture: amd64
- [x] pyenv version: pyenv 2.7.3-2-gebd788d
- [x] Python version: Python 3.13.6
- [x] C Compiler information (e.g. gcc 7.3): gcc 13.3.0
- [ ] Please attach the debug trace of the failing command as a [gist](https://gist.github.com/):
  * Can't get log due to inconsistent behavior

Sheile:

### Prerequisite
  - Closes https://github.com/pyenv/pyenv/issues/3482

### Description
Since the writable check has already succeeded once `mktemp` creates the temporary file, I think it should be safe to ignore a failure from `rm`.

I also considered other possible approaches, such as:

- excluding temporary files from removal in `remove_stale_shims` and similar functions
- using a different directory for temporary files created inside `acquire_lock`

However, these approaches may have a wider impact, so I think using `rm -f` is the smallest and least invasive change.

---
## Summary by cubic
Fixes a race in the rehash writable check that could falsely report `"$SHIM_PATH" isn't writable` during concurrent rehashes by tolerating the temp file being removed by another process.

- **Bug Fixes**
  - In `pyenv-rehash`’s `acquire_lock`, use `rm -f` on the temp file in `"$SHIM_PATH"` to avoid false failures if another rehash deletes it first; added and clarified an inline comment documenting this concurrency case.

<sup>Written for commit 19ae3820bdd612b982b64edbbec2bcf4803d5381. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3483?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -28,6 +28,8 @@ acquire_lock() {
       # So check for writablity by trying to write to a different file,
       # in a way that taxes the usual use case as little as possible.
       if [[ -z $tested_for_other_write_errors ]]; then
+        # if lots of (50+) rehashes run concurrentlty, another concurrent rehash
+        # may delete the temporary file in remove_*_shims() before we do so
         ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm -f "$t" ) \
           && tested_for_other_write_errors=1 \
           || { echo "pyenv: cannot rehash: $SHIM_PATH isn't writable" >&2
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -28,7 +28,7 @@ acquire_lock() {
       # So check for writablity by trying to write to a different file,
       # in a way that taxes the usual use case as little as possible.
       if [[ -z $tested_for_other_write_errors ]]; then
-        # if lots of (50+) rehashes run concurrentlty, another concurrent rehash
+        # if lots of (50+) rehashes run concurrently, another concurrent rehash
         # may delete the temporary file in remove_*_shims() before we do so
         ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm -f "$t" ) \
           && tested_for_other_write_errors=1 \
--- a/libexec/pyenv-rehash
+++ b/libexec/pyenv-rehash
@@ -28,7 +28,7 @@ acquire_lock() {
       # So check for writablity by trying to write to a different file,
       # in a way that taxes the usual use case as little as possible.
       if [[ -z $tested_for_other_write_errors ]]; then
-        ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm "$t" ) \
+        ( t="$(TMPDIR="$SHIM_PATH" mktemp)" && rm -f "$t" ) \
           && tested_for_other_write_errors=1 \
           || { echo "pyenv: cannot rehash: $SHIM_PATH isn't writable" >&2
                set +o noclobber


native-api (approved):

native-api:
Thank you! I don't run dozens of consoles, not to say launch them at the same moment, so I would have never run into this issue myself!
That's what users are for! 🙂

PR #3487, opened 2026-07-01
Add an experimental pyenv-binary plugin with a save command

macayu17:
This is a first step towards prebuilt CPython binaries (#2334), following the plan @native-api sketched out in that thread.

It adds a new experimental plugin under `plugins/pyenv-binary`, alongside python-build. The plugin is deliberately decoupled from python-build: it does not touch `pyenv install` or any other command. You drive it explicitly through `pyenv binary`, so regular usage is unaffected.

This PR is just the scaffold plus the first subcommand, `save`:

    pyenv binary save 3.12.7 ./dist

It packs an installed version into a relocatable `.tar.gz` (relative paths) and writes a `.meta` file next to it describing the build platform (OS, arch, distro and libc version) and the external system libraries the build links against. The metadata is a plain line-based format with no dependencies, so an installer can later check compatibility before unpacking.

I have kept this small on purpose. Generating an installer that consumes the archive and metadata is the next piece; I will send it as a separate PR so each part can be reviewed on its own.

---
## Summary by cubic
Experimental `pyenv-binary` plugin with a `save` command to package an installed Python into a relocatable archive plus metadata, laying groundwork for prebuilt CPython binaries (#2334) without changing existing commands. Tightens version validation and adds tests that verify dependency parsing on Linux and macOS.

- **New Features**
  - `save`: creates `<version>-<platform>.tar.gz` (relative paths) and `<version>-<platform>.meta` with OS, arch, distro/libc, and only external system libs (via `ldd` on Linux, `otool -L` on macOS); outputs to the given dir (default `.`); metadata is plain line-based.

- **Bug Fixes**
  - Reject invalid version names containing `/`, `.` or `..` to prevent path escapes.
  - Drop `readlink -f` for broader macOS/BSD compatibility; `ldd`/`otool` follow symlinks already.

<sup>Written for commit 09c31624f49e9303c596e29fb4b780d7b3a94080. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/libexec/pyenv-binary-save
+++ b/plugins/pyenv-binary/libexec/pyenv-binary-save
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+#
+# Summary: Save an installed Python version as a relocatable archive
+#
+# Usage: pyenv binary save <version> [<output-dir>]
+#
+#   Packs an installed version into a relocatable .tar.gz (relative paths) and
+#   writes a metadata file listing the build platform and the system libraries
+#   it links against, so an installer can check compatibility before unpacking.
+#
+#   <version>      An installed version, as listed by `pyenv versions --bare'.
+#   <output-dir>   Where to write the archive and metadata (default: `.').
+#
+set -e
+[ -n "$PYENV_DEBUG" ] && set -x
+
+# Provide pyenv completions
+if [ "$1" = "--complete" ]; then
+  exec pyenv-versions --bare
+fi
+
+version="$1"
+output_dir="${2:-$PWD}"
+
+if [ -z "$version" ]; then
+  echo "Usage: pyenv binary save <version> [<output-dir>]" >&2
+  exit 1
+fi
+
+prefix="${PYENV_ROOT}/versions/${version}"
+if [ ! -d "${prefix}/bin" ]; then
+  echo "pyenv-binary: version \`${version}' is not installed" >&2
+  exit 1
+fi
+
+os="$(uname -s)"
+arch="$(uname -m)"
+platform="$(printf '%s' "$os" | tr '[:upper:]' '[:lower:]')-${arch}"
+
+# Record the distro and libc version. Platform and arch alone are too coarse to
+# judge compatibility: a build is only portable to a matching libc (e.g. a
+# glibc 2.36 build will not load on an older glibc, nor on musl at all).
+distro=""
+libc=""
+if [ -r /etc/os-release ]; then
+  distro="$( . /etc/os-release && printf '%s %s' "${ID:-}" "${VERSION_ID:-}" )"
+elif [ "$os" = "Darwin" ]; then
+  distro="macos $(sw_vers -productVersion 2>/dev/null)"
+fi
+if [ "$os" = "Linux" ]; then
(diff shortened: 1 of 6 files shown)


plugins/pyenv-binary/libexec/pyenv-binary-save
```diff
@@ -0,0 +1,104 @@
...
+
+if [ -z "$version" ]; then
+  echo "Usage: pyenv binary save <version> [<output-dir>]" >&2
+  exit 1
+fi
+
+# A version is a single directory name under versions/. Reject a name with a
+# slash or a dot-dot component so it cannot point outside there.
+case "$version" in
+*/* | *..* )
```
native-api: If `/` is prohibited, only bare `..` (or `.`) would be a foul.

```suggestion
*/* | .. | . )
```

For reference, https://github.com/pyenv/pyenv/blob/46cc3e552b7f3b4c4bb996b162d0596ae2ff85e4/libexec/pyenv-version-file-read#L8-L21 guards against the same security issue -- but it does accept relative paths

plugins/pyenv-binary/libexec/pyenv-binary-save
```diff
@@ -0,0 +1,104 @@
...
+# List the external shared libraries the install links against: those that
+# resolve outside its own prefix, so they must already exist on the target.
+# The interpreter plus every bundled shared object are inspected.
+system_deps() {
+  local f
+  {
+    for f in "${prefix}"/bin/python*; do
+      [ -e "$f" ] && printf '%s\n' "$f"
+    done
+    find "${prefix}" -type f \( -name '*.so' -o -name '*.so.*' -o -name '*.dylib' \)
```
native-api, line 75: Do we need to test for `*.so.*` separately? AFAICS, CPython doesn't produce any `*.so.*` whatsoever and whenever I've seen builds that do, they always also produce an equivalent `*.so` alongside.<br>
I guess there's no harm... Worth pointing out in a comment that the "`*.so.*`" is here just in case as per above so that it doesn't raise further eyebrows.

native-api (commented):
Great code.

Definitely need a test for parsing `otool` and `ldd` output -- it's the most intricate part of the code. Otherwise, superb work!

macayu17:
Thanks for the review! Pushed a commit that covers all three:

- Tightened the version guard to reject just . and .. as you suggested, since a slash is already refused.
- Added the tests for the dependency parsing. There's one feeding a realistic `ldd` listing and one for `otool` (faking the platform), both checking that only the libraries resolving outside the prefix make it into the metadata, and that the vdso, the loader and the `@rpath`/in-prefix entries get dropped.
- Left a note on the `find` explaining the `*.so.*` is there just in case a build ships a versioned copy, since CPython itself doesn't produce them.

native-api (approved):

PR #3488, opened 2026-07-03
pyenv-binary: add the `generate-installer` subcommand

macayu17:
Follow-up to #3487, the next piece of the prebuilt-binary work in #2334.

The `save` command from #3487 packs an installed version into an archive plus a `.meta` file. This adds the command that consumes those: `generate-installer` reads a `.meta` file and writes a self-contained installer script for that archive.

```sh
pyenv binary generate-installer ./dist/3.12.7-linux-x86_64.meta \
  --archive-url https://example.com/3.12.7-linux-x86_64.tar.gz \
  -o ./dist/install-3.12.7.sh

./dist/install-3.12.7.sh ~/.pyenv/versions/3.12.7
```

The archive location is a parameter, so the archive can be hosted anywhere and does not have to be a pyenv location. The generated installer:

- refuses to run on a different OS/arch, or an older glibc, than the archive was built for
- warns if compiler environment variables are set, since they are ignored
- checks the required system libraries and an rpath patching tool are present
- downloads and unpacks the archive into a target prefix
- rewrites rpaths to be `$ORIGIN`-relative so the copy runs from its new path

Relocation is Linux-only for now; macOS is not wired up yet and the command refuses non-Linux metadata rather than emit an installer that cannot relocate.

Like the last one this stays decoupled: it is a plugin command, nothing in `pyenv install` changes. The build-and-package command that ties `save` and this together is the next step and I will send it separately.

---
## Summary by cubic
Adds `pyenv binary generate-installer` to create a `python-build` definition from a `.meta` file and adjacent archive, plus `pyenv binary relocate` to rewrite rpaths so the unpacked Python runs from its prefix. The installer enforces OS/arch and glibc checks, verifies required system libraries, and refuses macOS; relocation is Linux-only.

- **New Features**
  - Generate a `python-build` definition with a baked-in SHA256 from the adjacent archive; configurable download URL; installs via `install_package ... copy relocate` with a distinct `Python-<version>-binary` name.
  - Add `pyenv binary relocate` (Linux, `patchelf`) to set rpaths to `<prefix>/lib` for ELF executables under `bin/` and CPython’s `lib-dynload` extensions.

- **Bug Fixes**
  - Validation and platform checks: strict arg/metadata parsing, quoted values, exact soname checks via `ldconfig -p` (warn if unavailable), refuse macOS, glibc floor enforced on Linux; fix tarball layout to match `fetch_tarball()`; make subcommand help work by exporting the plugin’s `libexec`.
  - Tests/tooling/docs: add `test-binary` and Docker targets (split `python-build` and `pyenv-binary` suites), build Docker images only when missing, link `bats` conditionally, reuse core `test_helper`, skip install test when `ldconfig -p` is missing, update test docs; ignore `*.swp`.

<sup>Written for commit 017ad4a57312b38213fb8e73b671fcbd9815fe2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,218 @@
...
+    echo "pyenv-binary: unknown option \`$1'" >&2; exit 1 ;;
+  * )
+    [ -z "$metadata" ] || { echo "pyenv-binary: unexpected argument \`$1'" >&2; exit 1; }
+    metadata="$1"; shift ;;
+  esac
+done
+
+if [ -z "$metadata" ] || [ -z "$archive_url" ]; then
+  echo "Usage: pyenv binary generate-installer <metadata-file> --archive-url <url> [-o <output>]" >&2
+  exit 1
```
native-api: We have `pyenv-help` for this that generates the message based on the text at the file's top. No need to duplicate that info.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,218 @@
...
+fi
+if [ ! -r "$metadata" ]; then
+  echo "pyenv-binary: cannot read metadata file \`${metadata}'" >&2
+  exit 1
+fi
```
native-api: Is this check necessary? If there's a problem reading it, it'll show futher on when attempting to. The additional check is also subject to TOCTOU (that's a negligible possibility alright but that just shows that the check is not fool-proof which further questions its necessity).

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,218 @@
...
+  fi
+done
+
+# Relocation is only implemented for Linux so far (rpath rewriting). Refuse other
+# platforms rather than emit an installer that unpacks but cannot relocate.
+if [ "$os" != "Linux" ]; then
+  echo "pyenv-binary: only Linux archives are supported for now (metadata says \`${os:-unknown}')" >&2
+  exit 1
+fi
+
```
native-api: Is this PR intended to be a partial solution?
That's okay if so, I reckon MacOS is "special" enough to warrant a separate effort.

In MacOS, the rpath change tool is `install_name_tool` (I mentioned it in [the plan message](https://github.com/pyenv/pyenv/issues/2334#issuecomment-4487836151)), and it's preinstalled.

In FreeBSD, it's `chrpath`/`patchelf` like in Linux.
So I think we can default to `chrpath`/`patchelf` everywhere that's not MacOS.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,218 @@
...
+emit() {
+  # The header carries the values known now. Quote them with printf %q so a
+  # stray quote or space in the URL or metadata cannot break the generated
+  # script or inject into it. The body below is a quoted here-doc, verbatim.
+  {
+    echo "#!/usr/bin/env bash"
+    printf '# Installer for prebuilt Python %s (%s/%s%s).\n' \
```
native-api: Aren't we supposed to generate a Python-Build instlalation script rather than a standalone script?
Python-Build scripts are sourced by Python-Build at https://github.com/native-api/pyenv/blob/b52a8e3f52c3be68cf46c751ed40a180dfc48ba4/plugins/python-build/bin/python-build#L2775. Their usual structure is outlined in https://github.com/pyenv/pyenv/blob/master/CONTRIBUTING.md#advanced-changes--adding-new-python-flavor-support

Major repeating parts (e.g. checking system dependencies) can be put into `pyenv binary` subcommands that the script would call.

native-api (commented):
Thanks for the contribution! Sorry for the delay, had to allocate time to have a proper sit down and take a good look. This is would-be critical stuff, I can't afford to slack off here.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
macayu17: Makes sense, this should be a python-build definition rather than a standalone script. I'll rework it to set the version/url and hand off to `install_package ... copy`, with the relocation and dependency check pulled into `pyenv binary` subcommands that the definition calls.

One thing before I do: `install_package` expects the url with a `#sha256` on the end, but `save` doesn't record a checksum in the `.meta` right now. Would you prefer I add the checksum to `save`'s metadata, or wire it up some other way? Happy to go either way, just want to match how you'd expect it done.

native-api: The `#sha256` URL suffix is optional. If it's missing, the logic won't check the checksum (at least, that's what's supposed to happen).

native-api: However, since we're going to be downloading the binary package in the script, we certainly have to add the checksum to the generated script.

I see that you're not requiring the package to be present locally in this subcommand.
You can of course add the checksum to the metadata -- but I think that's redundant since we'll most certainly have the package locally at this point so it seems reasonable to require it.

native-api: OTOH if we have the checksum in the metadata, we're shielded from accidentally linking a wrong package to the script... I don't think it's too big of a danger though.

So it's your call.

macayu17: Going with computing the checksum from the archive rather than putting it in the metadata. save drops the archive and the .meta side by side, so it's right there at generate time. Having generate-installer hash it and bake in `url#sha256`, and error out if the archive isn't next to the .meta.

Also reworking it into a python-build definition instead of the standalone script. The plan is: check arch/glibc/system libs, hand off to `install_package ... copy`, then relocate. Moving the rpath rewriting into a `pyenv binary relocate` subcommand so the definition just calls it. Still ironing it out, will push once I've got it passing.

macayu17:
`generate-installer` now writes a python-build definition rather than a standalone installer. The definition checks the platform, glibc floor and required system libraries, then hands off to python-build's `install_package ... copy` to download, verify the checksum and unpack the archive. That drops the download and unpack code we were carrying ourselves.

The rpath rewriting moves into a new `pyenv binary relocate` subcommand that the definition calls once the tree is unpacked. The checksum is taken from the archive `save` writes next to the metadata and baked into the definition.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,188 @@
...
+    fi
+    ;;
+  esac
+  ;;
+esac
+
+# Required system libraries must already be present on the target. ldconfig -p
+# lists each library as "<soname> (...) => <path>"; match the soname column
+# exactly so a longer name like libc.so.6.1 does not satisfy libc.so.6.
+if [ "$os" = "Linux" ] && command -v ldconfig >/dev/null 2>&1; then
```
native-api: Earlier test checks for non-MacOS instead.

Altered to check for `ldconfig` with `-p` according to the further logic (FreeBSD doesn't have it).

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,188 @@
...
+      "$version" "$os" "$arch" "${distro:+, $distro}"
+    echo "# Generated by \`pyenv binary generate-installer'."
+    echo
+    printf 'EXPECT_OS=%q\n' "$os"
+    printf 'EXPECT_ARCH=%q\n' "$arch"
+    printf 'EXPECT_LIBC=%q\n' "$libc"
+    printf 'VERSION=%q\n' "$version"
+    printf 'ARCHIVE_URL=%q\n' "$archive_url"
+    printf 'SHA256=%q\n' "$sha256"
+    printf 'DEPS=%q\n' "$deps"
```
native-api, line 128: Testing this, I found out that `DEPS` end up containing all the libraries that Python and modules load, not just direct dependencies.

This needs to be fixed but it can be considered outside of scope of this PR.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,188 @@
...
+#
+set -e
+[ -n "$PYENV_DEBUG" ] && set -x
+
+# Provide pyenv completions
+if [ "$1" = "--complete" ]; then
+  echo --archive-url
+  echo -o
+  exit
+fi
```
native-api, line 31: Found out that upon double-tab, I only get the list of options. So I cannot pick and complete the name of a `.meta` file instead.

Looking further, I see that `completions/pyenv.bash` doesn't pass the last word or the index of the word being completed to `pyenv completions`. So completing a file name seems impossible with the current logic.

plugins/pyenv-binary/libexec/pyenv-binary-relocate
```diff
@@ -0,0 +1,56 @@
...
+# The interpreter loads libpython from ../lib. Patch every ELF binary in bin/ so
+# this holds whatever the interpreter is named (python3, python2.7, ...); the
+# scripts alongside it (pip, idle) are not ELF, so `--print-rpath' fails on them
+# and they are skipped. A tree with no interpreter at all did not unpack the way
+# we expect, so treat that as an error rather than quietly relocating nothing.
+found=0
+for file in "$prefix"/bin/*; do
+  [ -f "$file" ] && [ ! -L "$file" ] || continue
+  patchelf --print-rpath "$file" >/dev/null 2>&1 || continue
+  patchelf --set-rpath '$ORIGIN/../lib' "$file"
```
native-api: The result is not what the original rpath was but "`$ORIGIN/../lib`".
It would break if e.g. the interpreter is hardlinked into a different location.

plugins/pyenv-binary/libexec/pyenv-binary-relocate
```diff
@@ -0,0 +1,56 @@
...
+  echo "pyenv-binary: found no interpreter to relocate under \`${prefix}/bin'" >&2
+  exit 1
+fi
+
+# The extension modules under lib-dynload load libpython from ../.. relative to
+# their own directory. Each match is an ELF object we expect to patch, so let a
+# failure stop us rather than swallow it.
+while IFS= read -r so; do
+  patchelf --set-rpath '$ORIGIN/../..' "$so"
+done < <(find "$prefix" -path '*/lib-dynload/*.so')
```
native-api, line 58: In `save`, you also search for `*.so.*`. Either remove it from there, or add here.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
...
+  # If the url were interpolated unquoted, sourcing the assignments would run the
+  # injected command. Stub the build calls so only the assignments matter.
+  run bash -c '
+    uname() { echo unmatched; }
+    install_package() { :; }
+    pyenv() { :; }
+    source "$1"
+  ' _ "$out"
+  assert [ ! -e "${BATS_TEST_TMPDIR}/pwned" ]
+}
```
native-api: What exactly does this achieve? This seems to be a vulnerability check -- but the thread model is unclear.
This only tests one possible way of incorrectly handling one of the inputs.
Inputs to `generate-installer` are supplied by the packager so they are supposed to be non-malicious.
If we defend against the packaging machine becoming compromised, other inputs, as well as the package, could be malicious as well.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env bats
```
native-api, line 1: Lots and lots of repeated and almost-repeated nontrivial code, would be impossible to maintain without refactoring first.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
...
+  assert_success
+  assert_output_contains "EXPECT_OS=FreeBSD"
+}
+
+@test "fails when metadata is missing the architecture" {
+  local meta="${BATS_TEST_TMPDIR}/no-arch.meta"
+  printf 'version=3.12.7\nos=Linux\nlibc=glibc 2.36\narchive=x.tar.gz\n' > "$meta"
+  run pyenv-binary-generate-installer "$meta" --archive-url http://x/a.tar.gz
+  assert_failure "pyenv-binary: metadata is missing \`arch'"
+}
```
native-api: Doesn't test other fields

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+# Write a minimal Linux metadata file next to the archive it describes (save
+# writes the two together) and echo the metadata path. The glibc floor is kept
+# low so the generated definition's check passes on any modern system.
+create_meta() {
```
native-api, line 5: Used for a few tests but in other tests, the same is done by hand.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
...
+  run pyenv-binary-generate-installer "$meta" --archive-url http://x/a.tar.gz
+  assert_failure "pyenv-binary: metadata is missing \`arch'"
+}
+
+@test "fails when Linux metadata is missing libc" {
+  local meta="${BATS_TEST_TMPDIR}/no-libc.meta"
+  printf 'version=3.12.7\nos=Linux\narch=x86_64\narchive=x.tar.gz\n' > "$meta"
+  run pyenv-binary-generate-installer "$meta" --archive-url http://x/a.tar.gz
+  assert_failure "pyenv-binary: metadata is missing \`libc'"
+}
```
native-api, line 83: Not consecutive to the other test for the same field.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
...
+
+@test "carries the platform and archive url into the definition" {
+  run pyenv-binary-generate-installer "$(create_meta)" \
+    --archive-url http://example.com/a.tar.gz
+  assert_success
+  assert_output_contains "EXPECT_OS=Linux"
+  assert_output_contains "EXPECT_ARCH=x86_64"
+  assert_output_contains "http://example.com/a.tar.gz"
+}
```
native-api: How is testing these particular things useful? Without context even.

plugins/pyenv-binary/test/generate-installer.bats
```diff
@@ -0,0 +1,169 @@
...
+  run pyenv-binary-generate-installer --complete
+  assert_success
+  assert_output "--archive-url
+-o"
+}
+
+@test "fails with no arguments" {
+  run pyenv-binary-generate-installer
+  assert_failure
+  assert_output_contains "Usage: pyenv binary generate-installer"
```
native-api: This redundantly tests `pyenv-help`. Better stub it. Same elsewhere.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
```diff
@@ -0,0 +1,188 @@
...
+  -* )
+    echo "pyenv-binary: unknown option \`$1'" >&2; exit 1 ;;
+  * )
+    [ -z "$metadata" ] || { echo "pyenv-binary: unexpected argument \`$1'" >&2; exit 1; }
+    metadata="$1"; shift ;;
+  esac
+done
+
+if [ -z "$metadata" ] || [ -z "$archive_url" ]; then
+  PATH="${BASH_SOURCE%/*}:$PATH" pyenv-help --usage binary-generate-installer >&2
```
native-api: `PATH="${BASH_SOURCE%/*}:$PATH"` seems unneeded. The `pyenv` launcher adds the necessary entries to `PATH`.

plugins/pyenv-binary/test/relocate.bats
```diff
@@ -0,0 +1,51 @@
...
+  run pyenv-binary-relocate
+  assert_failure
+  assert_output_contains "Usage: pyenv binary relocate"
+}
+
+@test "fails when patchelf is not available" {
+  local bash_path; bash_path="$(command -v bash)"
+  mkdir -p "${BATS_TEST_TMPDIR}/empty"
+  run env PATH="${BATS_TEST_TMPDIR}/empty" "$bash_path" \
+    "${BATS_TEST_DIRNAME}/../libexec/pyenv-binary-relocate" "${BATS_TEST_TMPDIR}/prefix"
```
native-api: We have `path_without` for this in Pyenv's `test_helper`.

Why not symlink `test_helper` from there?

In Python-Build, that is not done because Python-Build is supposed to be able to be installed independently from Pyenv. But for this plugin, that is not so.

plugins/pyenv-binary/test/relocate.bats
```diff
@@ -0,0 +1,51 @@
...
+  assert_failure
+  assert_output_contains "need patchelf"
+}
+
+@test "fails when the prefix has no interpreter" {
+  mkdir -p "${BATS_TEST_TMPDIR}/tools" "${BATS_TEST_TMPDIR}/prefix"
+  printf '#!/bin/sh\nexit 0\n' > "${BATS_TEST_TMPDIR}/tools/patchelf"
+  chmod +x "${BATS_TEST_TMPDIR}/tools/patchelf"
+  run env PATH="${BATS_TEST_TMPDIR}/tools:$PATH" \
+    pyenv-binary-relocate "${BATS_TEST_TMPDIR}/prefix"
```
native-api: We have `create_path_executable` for this. Same elsewhere.

plugins/pyenv-binary/README.md
```diff
@@ -25,3 +25,36 @@ distro and libc version) and the system libraries the build links against.
...
+must sit next to the `.meta` file so its checksum can be baked into the
+definition. The definition:
+
+- refuses to install on a different OS/architecture, or an older glibc, than the
+  archive was built for,
+- checks the required system libraries are present,
+- hands off to python-build's `install_package ... copy` to download, verify the
+  checksum and unpack the archive,
+- calls `pyenv binary relocate` to rewrite rpaths so the copy runs from its
+  prefix.
```
native-api: That's implementation details irrelevant to the user

plugins/pyenv-binary/README.md
```diff
@@ -25,3 +25,36 @@ distro and libc version) and the system libraries the build links against.
...
+pyenv install 3.12.7-linux-x86_64
+```
+
+### `pyenv binary relocate <prefix>`
+
+Rewrites the rpaths of a Python tree unpacked into `<prefix>` to be
+`$ORIGIN`-relative, using `patchelf`, so it runs from its new location. The
+generated definition calls this; you rarely run it by hand.
+
+Relocation is implemented for Linux and the BSDs; macOS is not wired up yet.
```
native-api: The BSD part is likely false as per above.

plugins/pyenv-binary/README.md
```diff
@@ -25,3 +25,36 @@ distro and libc version) and the system libraries the build links against.
...
+  --archive-url https://example.com/3.12.7-linux-x86_64.tar.gz \
+  -o "$(pyenv root)/plugins/python-build/share/python-build/3.12.7-linux-x86_64"
+
+pyenv install 3.12.7-linux-x86_64
+```
+
+### `pyenv binary relocate <prefix>`
+
+Rewrites the rpaths of a Python tree unpacked into `<prefix>` to be
+`$ORIGIN`-relative, using `patchelf`, so it runs from its new location. The
```
native-api: relative won't do as per above

native-api (commented):
AI once again rears its ugly head. This was clearly generated as it doesn't contain blunders typical for humans, and does contain blunders non-typical for humans.

macayu17:
Extremely sorry for the noise. I used AI help on this and didn't scrutinise what came out, and the tests are where it shows worst. Sorry for the time you've spent on it.

I should have read your `test_helper` before writing a line of that file. `path_without` and `create_path_executable` were already sitting right there and I hand-rolled both, over and over. The injection test is guarding against a threat that doesn't exist, and a few of the others don't really assert anything worth asserting. I can't defend it.

Thanks for pushing the fixes, and sorry you had to. I'd like to put it right. Let me take the tests back and redo them on top of your commits, cut down to the ones that test something real, and I'll go through the rest of your comments properly rather than firing back patches.

I'm invested in seeing #2334 through, so I'd rather slow down and get it right than keep wasting your review time.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
macayu17: I think this one is still needed, just not here. The launcher adds plugins/*/bin to PATH, not plugins/*/libexec, and pyenv-help resolves the command with `command -v`, so without it the usage text prints nothing. I moved it into the dispatcher so it is set once and the subcommands do not each carry it. Let me know if you had something else in mind.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
macayu17: Noted. Happy to do that as a separate PR once this one lands.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
native-api: Then it looks like `pyenv-binary` -- which acts as a laucher for these subcommands -- needs to add `pyenv-binary/libexec` to `PATH`.

plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
native-api: I'm ambivalent about this so whichever is more convenient. I've already fixed an incompatibility in `save` within this PR.

macayu17: Separate PR then, if that is alright. It needs different tooling rather than a tweak: `ldd` walks the whole dependency tree, so the direct dependencies are really the `DT_NEEDED` entries, and the current in-prefix filter works off the paths `ldd` resolves, which those do not give you. Cleaner on its own.

macayu17:
Reworked the tests on top of your commits.

Sorry for the earlier test slop. I should have exercised the real python-build path from the start.

The plugin now uses pyenv’s shared test helper and its existing helpers for stubs and PATH setup. I removed the tests that only checked generated text, plus the FreeBSD case that no longer applies after the `ldconfig` change. Required metadata fields are covered in one table-driven test.

The definition test now runs `python-build` against a generated definition and a real archive. Previously it only sourced the definition with `install_package` and `pyenv` stubbed, so it missed the tarball-layout and package-name issues. The interpreter and versioned shared-library relocation paths now have coverage too.

I moved the plugin `libexec` PATH setup into the dispatcher, so subcommands and their help work without repeating it in each command.

Two follow-ups I left alone:

  - `save` still prints its own usage text instead of using `pyenv-help`. It is already merged.
  - `make test` and `make test-docker` only run `plugins/python-build/test`. I did not change those shared targets in this PR.

Ran all 26 tests in a Linux checkout with the helper symlink materialized and `patchelf` unavailable. I do not have macOS available to run the suite there; its two save paths are covered with stubs.

plugins/pyenv-binary/README.md
```diff
@@ -25,3 +25,33 @@ distro and libc version) and the system libraries the build links against.
...
+```
+
+### `pyenv binary relocate <prefix>`
+
+Rewrites the rpaths of a Python tree unpacked into `<prefix>` so the interpreter
+and its extension modules load the bundled libraries from there rather than from
+the path the archive was built at. Uses `patchelf`. The generated definition
+calls this; you rarely run it by hand.
+
+Relocation is implemented for Linux; macOS is not wired up yet.
```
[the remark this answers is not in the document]
native-api: @cubic-dev-ai BSD doesn't have `ldconfig -p` required by further logic -- so while BSD is  technically accepted, the code is likely broken for it as of now.

Makefile
```diff
@@ -81,32 +81,22 @@ $(TEST_BATS_IMAGE_TARGETS):
 		-t $(DOCKER_IMAGE):$(DOCKER_TAG) \
 		./
 
-.PHONY: test test-build test-unit test-plugin
+.PHONY: test test-unit test-python-build test-binary
```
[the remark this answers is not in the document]
native-api: Running parts of Pyenv tests is an internal facility and as such, is not stable for external use.

native-api (approved):
The UX is questionable. `.meta` name is autogenerated but has to be supplied to `generate-installer`. Script name is not autogenerated but probably should.

But let's postpone this to the point when we are actually using them: until then, we don't quite know what exact interface we need.

macayu17:
Thanks for all the work on this, especially wiring the tests into CI and sorting out the Docker matrix. That was a lot more than I'd have gotten right on my own.

On the UX point: agreed, let's postpone. Passing the `.meta` path in and naming the script by hand both feel awkward, but until the build side actually calls `save` and `generate-installer` we don't really know what the interface should be. Easier to get that right once something's using them than to guess now.

PR #3493, opened 2026-07-14
Miniforge3 uses 'Darwin' instead of 'MacOSX' for naming scheme

j9ac9k:

### Prerequisite
  - No relevant open issue

### Description

If you run `pyenv install miniforge3-latest` you'll get quite an old version (when on macOS).  Some years ago, the file naming scheme for miniforge3-latest changes on macOS.

### Tests
- [x] My PR adds the following unit tests (if any) - None

---
## Summary by cubic
Switch macOS Miniforge3 asset names from "MacOSX" to "Darwin" to match upstream, so `pyenv install miniforge3-latest` fetches the current macOS build on both arm64 and x86_64.

<sup>Written for commit 4ff711bf2d474625a0807af4022e1ce411e1067f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/python-build/share/python-build/miniforge3-latest
+++ b/plugins/python-build/share/python-build/miniforge3-latest
@@ -9,10 +9,10 @@ case "$(anaconda_architecture 2>/dev/null || true)" in
   install_script "Miniforge3-Linux-aarch64" "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-aarch64.sh" "miniconda" verify_py3_latest
   ;;
 "MacOSX-arm64" )
-  install_script "Miniforge3-MacOSX-arm64" "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-arm64.sh" "miniconda" verify_py3_latest
+  install_script "Miniforge3-Darwin-arm64" "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Darwin-arm64.sh" "miniconda" verify_py3_latest
   ;;
 "MacOSX-x86_64" )
-  install_script "Miniforge3-MacOSX-x86_64" "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-x86_64.sh" "miniconda" verify_py3_latest
+  install_script "Miniforge3-Darwin-x86_64" "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Darwin-x86_64.sh" "miniconda" verify_py3_latest
   ;;
 * )
   { echo


native-api:
Looking at https://github.com/conda-forge/miniforge/releases/#release-26.3.2-3, I see "MacOS" rather than "Darwin" in the list of assets.
and I also see that they dropped MacOS-x86_64 support in [26.3.2-3](https://github.com/conda-forge/miniforge/releases/tag/26.3.2-3)

Yet, your URLs seems to work despite not being present in the release assets. How is that possible?

j9ac9k:
> Looking at https://github.com/conda-forge/miniforge/releases/#release-26.3.2-3, I see "MacOS" rather than "Darwin" in the list of assets. and I also see that they dropped MacOS-x86_64 support in [26.3.2-3](https://github.com/conda-forge/miniforge/releases/tag/26.3.2-3)
> 
> Yet, your URLs seems to work despite not being present in the release assets. How is that possible?

I tried going through the miniforge history, I was unclear; here's the relevant bit of code in their CI config I believe:

https://github.com/conda-forge/miniforge/blob/3fbc4935f982120b923118c9c3319c3e04e5dc68/.github/workflows/ci.yml#L185-L188

What I'm unclear about is why `pyenv install miniforge3-latest` _right now_ installs an old version of miniforge3, which is what they say you should be doing in the README.

```
curl -fsSLo Miniforge3.sh "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-$(uname -m).sh"
```

native-api:
Looked at it closer.

* Looking closer at https://github.com/conda-forge/miniforge/releases/tag/26.3.2-3, it does have both "Darwin" and "MacOS" in assets, and has "MacOS-x86_64", too. Looks like the release list page was not showing all the artifacts.
* The alias was added in https://github.com/conda-forge/miniforge/commit/85754d8b64c6668e141a8373c979826cbe25d48c#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5, and the "MacOS" file is the primary artifact. The commit message, diff, [linked PR](https://github.com/conda-forge/miniforge/pull/434) and [issue](https://github.com/conda-forge/miniforge/issues/429) -- none has any explanation why this was done.
* [Miniforge README as of now](https://github.com/conda-forge/miniforge/blob/3fbc4935f982120b923118c9c3319c3e04e5dc68/README.md) still lists the "MacOS" file as the primary artifact. It's also linked to from other places like the [PKG section](https://github.com/conda-forge/miniforge/blob/3fbc4935f982120b923118c9c3319c3e04e5dc68/README.md#macos-pkg-installers)
  * [But it also lists the `uname`-based invocation later](https://github.com/conda-forge/miniforge/blob/3fbc4935f982120b923118c9c3319c3e04e5dc68/README.md#unix-like-platforms-macos-linux--wsl)
* Versioned release artifacts only feature the "MacOS" name.

---

Based on the above, I conclude:

* Neither naming scheme seems to be deprecated now
* The "MacOS" scheme is still the primary one, "Darwin" is syntactic sugar to support the `uname`-based invocation

---

As such, a rename like the current PR diff looks harmless but unwarranted.

If you're getting an old version, something must be wrong at your side. E.g.:

* you can check the package hash at https://api.github.com/repos/conda-forge/miniforge/releases/latest and see if it's equal to what you are getting on the corresponding URL
* you may have [Pyenv package cache](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#package-download-cache) enabled and it has an old arttifact

j9ac9k:
@native-api I appreciate you diving into this, I suspect your assumption about the pyenv package cache is correct.  While I didn't see a Miniforge3-MacOS-arm64.sh file there, I removed all the files there and re-installed `miniforge3-latest` and it grabbed the current release (as verified by both the hash and the contents of the installed environment).

I'm sorry to have taken so much of your time.

native-api:
We can still accept a simplification/future-proofing of the script based on the `uname`-based invocation if you're interested.

If Pyenv package cache interferes with installing rolling releases, that's a Pyenv bug.

j9ac9k:
> We can still accept a simplification/future-proofing of the script based on the uname-based invocation if you're interested.

Given that's what miniforge is doing, it is probably a good idea to adapt here.  I'll submit a new PR in a day or so as I've already closed this one out.

> If Pyenv package cache interferes with installing rolling releases, that's a Pyenv bug.

This could certainly be the case.  Before I make any declaration one way or another, I'll want to do more testing

PR #3498, opened 2026-07-22
pyenv-binary: add the `package` subcommand; add `pyenv install --list --bare`

macayu17:
### Prerequisite

    * This is entirely inside the `pyenv-binary` plugin. Nothing in the core or `python-build` changes.
  * [x] Please consider contributing the patch upstream to rbenv.
  * [x] My PR addresses the following pyenv issue.
    * Part of #2334, step 4 of the plan. It does not close the issue.

### Description

Follows #3487 (`save`) and #3488 (`generate-installer`).

  `package` combines the existing steps for producing a private binary package:

  ```sh
  pyenv binary package 3.13.14:3.13.14-debian-12 \
    --archive-base-url https://example.com/binaries
  ```

  It installs the source version under the explicit entry name, saves that installation, and generates a matching python-build definition. The example writes:

  ```text
  3.13.14-debian-12.tar.gz
  3.13.14-debian-12.meta
  3.13.14-debian-12
  ```

  The first two files are the archive and metadata. The third is the python-build definition. Its `ARCHIVE_URL` points to:

  ```text
  https://example.com/binaries/3.13.14-debian-12.tar.gz
  ```

  `save` now accepts `--name` so `package` can use the explicit entry for the archive and metadata names. Standalone `save` keeps its existing `<version>-<platform>` default.

  Completion delegates to `pyenv install --list` instead of duplicating its plugin definition discovery. Invalid entry names are rejected before compilation begins.

  The command currently refuses macOS before compiling because `generate-installer` does not support macOS archives yet.

  The README documents the private package workflow and the updated interfaces.

  ### Tests

  `plugins/pyenv-binary/test/package.bats` covers:

  * completion through `pyenv install --list`
  * missing and invalid arguments
  * invalid entry names
  * refusing macOS before compilation
  * creating the archive, metadata, and definition under the explicit entry name
  * constructing the full archive URL from `--archive-base-url`
  * handling a trailing slash in the base URL

  `plugins/pyenv-binary/test/save.bats` also covers explicit package names and invalid `--name` values.

  All 40 `pyenv-binary` tests pass in Debian 12, including installation through the generated python-build definition. The repository’s Ubuntu and macOS checks also pass.

---
## Summary by cubic
Adds `pyenv binary package` to install a Python version, save it as a relocatable archive, and emit a matching `python-build` definition in one step. Adds `--bare` to `pyenv install --list` and uses it for completion; Linux only, with macOS refused early.

- **New Features**
  - New command: `pyenv binary package <version>:<entry> --archive-base-url <url>` runs install, save, and generate-installer; writes `<entry>.tar.gz`, `<entry>.meta`, and `<entry>` with `ARCHIVE_URL=<url>/<entry>.tar.gz` (trims a trailing `/`).
  - Validates entry names; rejects `latest`, names with `/`, `.`, `..`, or control chars; refuses macOS before compiling.
  - Completion supports `--archive-base-url` and lists buildable versions via `pyenv install --list --bare`, including plugin definitions.
  - `pyenv binary save` adds `--name` to set a custom archive/metadata base name.

<sup>Written for commit 1c2c8dd66e9e28ee8e81d4bafe15684e644a15c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/libexec/pyenv-binary-build
+++ b/plugins/pyenv-binary/libexec/pyenv-binary-build
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+#
+# Summary: Build a Python version and package it as an installable binary
+#
+# Usage: pyenv binary build <version>:<entry> --archive-url <url>
+#
+#   Builds <version> from source under the separate name <entry>, packages the
+#   result with `pyenv binary save', then emits a python-build definition for
+#   it with `pyenv binary generate-installer'. The archive, metadata and
+#   definition are written to the current directory, named after <entry>.
+#
+#   <version>             A version `pyenv install' knows how to build.
+#   <entry>               The name to build under and to install the binary
+#                         as, e.g. `3.13.14-debian-12'. Keeping it distinct
+#                         from <version> lets the binary sit alongside a
+#                         normal source install of the same version.
+#   --archive-url <url>   Where the archive will be hosted; the definition
+#                         downloads it from <url>/<archive-name>.
+#
+set -e
+[ -n "$PYENV_DEBUG" ] && set -x
+
+# Provide pyenv completions. Only the definitions are useful here: the options
+# `pyenv install' takes are not accepted.
+if [ "$1" = "--complete" ]; then
+  echo --archive-url
+  exec python-build --definitions
+fi
+
+spec=""
+archive_url=""
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+  --archive-url )
+    [ $# -ge 2 ] || { echo "pyenv-binary: --archive-url needs a value" >&2; exit 1; }
+    archive_url="$2"; shift 2 ;;
+  -* )
+    echo "pyenv-binary: unknown option \`$1'" >&2; exit 1 ;;
+  * )
+    [ -z "$spec" ] || { echo "pyenv-binary: unexpected argument \`$1'" >&2; exit 1; }
+    spec="$1"; shift ;;
+  esac
+done
+
+if [ -z "$spec" ] || [ -z "$archive_url" ]; then
+  pyenv-help --usage binary-build >&2
+  exit 1
+fi
+
+case "$spec" in
+*?:?* ) ;;
+* )
+  echo "pyenv-binary: expected <version>:<entry>, e.g. \`3.13.14:3.13.14-debian-12'" >&2
+  exit 1
+  ;;
+esac
+entry="${spec##*:}"
+
+case "$entry" in
+# `pyenv install' reads a trailing `:latest' as part of the version rather than
+# as an alias, so nothing would end up installed under that name.
+latest )
+  echo "pyenv-binary: \`latest' cannot be used as an entry name" >&2
+  exit 1
+  ;;
+# The entry becomes a directory name under versions/. `pyenv install' does not
+# validate the alias, and `save' only checks once the build is done, so refuse
+# a name that could point elsewhere before compiling anything.
+*/* | .. | . )
+  echo "pyenv-binary: invalid entry name \`${entry}'" >&2
+  exit 1
+  ;;
+esac
+
+# `pyenv install' puts a `<version>:<alias>' build under versions/<alias>.
+pyenv-install "$spec"
+pyenv-binary-save "$entry"
+
+# Name the metadata and the archive the way `save' just named them.
+platform="$(uname -s | tr '[:upper:]' '[:lower:]')-$(uname -m)"
+pyenv-binary-generate-installer "${entry}-${platform}.meta" \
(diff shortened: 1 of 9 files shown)


plugins/pyenv-binary/libexec/pyenv-binary-build
```diff
@@ -0,0 +1,100 @@
...
+
+# Provide pyenv completions. Only the definitions are useful here: the options
+# `pyenv install' takes are not accepted. Include definitions from other
+# plugins just as `pyenv install' does.
+if [ "$1" = "--complete" ]; then
+  shopt -s nullglob
+  for plugin_path in "$PYENV_ROOT"/plugins/*/share/python-build; do
+    PYTHON_BUILD_DEFINITIONS="${PYTHON_BUILD_DEFINITIONS}:${plugin_path}"
+  done
+  export PYTHON_BUILD_DEFINITIONS
```
native-api: This duplicates `pyenv-install` code. Need to either refactor the duplicating logic out (e.g. to a shared module in `python-build/lib`), or somehow delegate to existing public interface (what about `pyenv-install --list`?).

plugins/pyenv-binary/libexec/pyenv-binary-build
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+# `pyenv install' puts a `<version>:<alias>' build under versions/<alias>.
+pyenv-install "$spec"
+pyenv-binary-save "$entry"
+
+# Name the metadata and the archive the way `save' just named them.
+platform="$(printf '%s' "$os" | tr '[:upper:]' '[:lower:]')-$(uname -m)"
+pyenv-binary-generate-installer "${entry}-${platform}.meta" \
+  --archive-url "${archive_url%/}/${entry}-${platform}.tar.gz" -o "$entry"
```
native-api: As per https://github.com/pyenv/pyenv/issues/2334#issuecomment-4991313248, you are currently only implementing Scenario 1 (private custom builds with an explicit name). If so -- why are these names autogenerated by some unrelated pattern? When making a custom build script under a specific name, the user probably wants them named the same.

plugins/pyenv-binary/libexec/pyenv-binary-build
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+#
+# Summary: Build a Python version and package it as an installable binary
+#
+# Usage: pyenv binary build <version>:<entry> --archive-url <url>
```
native-api: Could you explain the logic behind the "`build`" name? In programming, "building" usually means converting source code into object code for testing and debugging. That's what `pyenv install` and `python-build` do. Creating a distribution package from that object code is typically a separate step after that -- because it needs to only be done occasionally, when the development is complete. So it looks confusing to me.
Maybe I'm just too old, and nowadays, in modern ecosystems, development is done differently?

native-api: `--archive-url` looks misleading. `generate-installer` also accepts "`--archive-url`" but it's a full URL while this one is a partial one.
Maybe "`--archive-base-url`"?

plugins/pyenv-binary/libexec/pyenv-binary-build
macayu17: Switched this to `pyenv-install --list` and strip only its heading and indentation for completion. This removes the duplicated plugin definition discovery.

plugins/pyenv-binary/libexec/pyenv-binary-build
macayu17: Renamed the command to `package`. It still invokes pyenv install first, but its user-facing result is a distributable package, so package describes the operation more clearly.

plugins/pyenv-binary/libexec/pyenv-binary-build
macayu17: Changed this to use the explicit entry name for all three outputs. save now accepts --name, so an entry such as 3.12.7-test produces 3.12.7-test.tar.gz, 3.12.7-test.meta, and the 3.12.7-test definition. Standalone save keeps its existing platform-suffixed default.

plugins/pyenv-binary/libexec/pyenv-binary-build
macayu17: Renamed this to `--archive-base-url`. `package` appends `/<entry>.tar.gz` and passes the resulting full URL to `generate-installer`.

PR #3504, opened 2026-07-30
pyenv-binary: generate platform-specific package names

macayu17:
### Prerequisite

    * Part of #2334, Scenario 2 of step 4. It does not close the issue.
  ### Description

  Follows #3498.

  `pyenv binary package` now accepts a bare version and generates an entry name from the current platform:

  ```sh
  pyenv binary package 3.13.14 \
    --archive-base-url https://example.com/binaries
  ```

On Debian 12 x86_64, this creates `3.13.14-debian-12-x86_64`. The existing `<version>:<entry>` form remains available for custom package names.

The naming logic is also available through `pyenv binary package-name <version>`. Linux names use the distribution and version, macOS names use the macOS version, and other systems use the name and release reported by `uname`. The architecture is included in every generated name. Libc compatibility remains in the metadata.

### Tests

`plugins/pyenv-binary/test/package-name.bats` covers completion, argument validation, Linux, macOS, FreeBSD, and invalid version names.

`plugins/pyenv-binary/test/package.bats` covers the bare-version flow through the real naming command, including the archive, metadata, and generated definition.

The repository's Bash 4.1 GNU Docker run completed with 46 passing tests and 1 skip because `ldconfig -p` was unavailable. I also built and reinstalled CPython 3.13.14 through the generated definition on Debian 12.

---
## Summary by cubic
Adds automatic, platform-specific package names to `pyenv-binary`, so `pyenv binary package <version>` works without a custom `<entry>`. Also adds `pyenv binary package-name` to print the generated name. Part of #2334.

- New Features
  - `pyenv binary package <version>[:<entry>]`: when `<entry>` is omitted, the name is built from OS, platform version, and arch (Linux uses distro/version; macOS uses macOS version; others use `uname`).
  - `pyenv binary package-name <version>`: outputs the computed entry name without building, for scripting/CI.
  - Custom names via `<version>:<entry>` remain supported; libc compatibility stays in metadata.

<sup>Written for commit 668aa6b44e07364e2fb3ae87f9775185790069d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/README.md
+++ b/plugins/pyenv-binary/README.md
@@ -16,21 +16,38 @@ and dependency metadata exist to catch that.
 
 ## Commands
 
-### `pyenv binary package <version>:<entry> --archive-base-url <url>`
+### `pyenv binary package <version>[:<entry>] --archive-base-url <url>`
 
-Installs `<version>` from source under the separate name `<entry>`, packages
-that install with `save`, then emits a python-build definition for it with
-`generate-installer`. The archive, metadata and definition land in the current
-directory, named after the entry; host the archive under `<url>` and drop the
-definition into python-build's definition directory. Keeping the entry name
-distinct from the version lets the binary sit alongside a normal source
-install of the same version.
+Installs `<version>` from source under a separate name, packages that install
+with `save`, then emits a python-build definition for it with
+`generate-installer`. With no explicit entry, the name is generated from the
+current platform, platform version and architecture. An explicit entry keeps
+the existing custom-build workflow.
 
 ```sh
-pyenv binary package 3.12.7:3.12.7-debian-12 \
+pyenv binary package 3.12.7 \
   --archive-base-url https://example.com/binaries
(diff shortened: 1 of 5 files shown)


native-api (approved):

native-api:
LGTM. `slug()` is especially impressive!

native-api:
Sorry for a late review. I've been busy with a tough job here and took a break before the final stint to rest up and catch up with the project. Hope to be finished with it tomorrow.

native-api:
And thanks for supporting MacOS here, too, and not putting it off! Reward well earned.

PR #3506, opened 2026-08-03
versions: fix --executables portability regression

timotheosh:
### Prerequisite

  * This fixes a portability regression in the internal `pyenv versions --executables` fast path used by `pyenv rehash`. A plugin or hook cannot correct the failing core command before it is invoked.

* [x] Please consider contributing the patch upstream to [[rbenv](https://github.com/rbenv/rbenv)](https://github.com/rbenv/rbenv), since we have borrowed most of the code from that project.

  * The affected `--executables` fast path is specific to pyenv and was introduced in pyenv commit `8037f226`; `rbenv-versions` has no equivalent code path. However, this change is consistent with the implementation style already used by `rbenv-versions`, which uses Bash parameter expansion (`"${path##*/}"`) rather than relying on non-portable utility options.

  * This regression was introduced by pyenv-specific commit `8037f226`, which added the `pyenv versions --executables` path. The affected code does not appear to originate from rbenv.

  * No existing issue found.

### Description

* [x] Here are some details about my PR

Commit `8037f226` introduced a fast path for executable discovery using:

```sh
xargs -0 -r basename -a
```

These options are not portable across non-GNU userlands. The native illumos implementations support neither `xargs -r` nor `basename -a`. OpenBSD also does not support `basename -a`.

As a result, `pyenv rehash` fails, including when it is invoked during shell initialization.

This change replaces the non-portable utility options with Bash 3.2-compatible parameter expansion:

```sh
"${path##*/}"
```

The optimized behavior is retained: executable paths are discovered directly from the versions directory, filtering and link resolution are skipped, and the resulting basenames are sorted and deduplicated.

### Tests

* [x] My PR adds the following unit tests (if any)

No new unit tests were added because the existing executable-discovery behavior is unchanged.

The change was tested on:

* OmniOS/illumos
* OpenBSD

Testing included:

* Running `pyenv versions --executables`
* Running `pyenv rehash`
* Verifying shell initialization completes without utility-option errors
* Comparing the output against the previous implementation using GNU `xargs` and GNU `basename`
* Testing duplicate executable names, virtual-environment executables, dotfiles, filenames containing spaces, and an empty versions directory

---
## Summary by cubic
Fixes a portability regression in `pyenv versions --executables` that broke `pyenv rehash` on non-GNU systems by replacing GNU-specific tooling with Bash 3.2-safe parameter expansion while keeping the fast path.

- **Bug Fixes**
  - Replaced `xargs -r`/`basename -a` with array-based `"${paths[@]##*/}"` plus `sort -u`, and added a length check to avoid running with no matches.
  - Kept direct discovery from `versions/*/bin` and `versions/*/envs/*/bin`, restoring `pyenv rehash` during shell init on illumos and OpenBSD.

<sup>Written for commit bb07d2e18a5d62a03a1053cd62d2ff9dc74026de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -42,9 +42,15 @@ versions_dir="${PYENV_ROOT}/versions"
 if [[ -n "$executables" ]]; then
   if [ -d "$versions_dir" ]; then
     shopt -s dotglob nullglob
-    # MacOS 12+ and FreeBSD 15 support `xargs -r -0' and `basename -a'
-    # `sort -u` is simpler and a bit faster than `awk '!seen[$0]++'`, with the same result for rehash purposes
-    printf '%s\0' "$versions_dir"/*/bin/* "$versions_dir"/*/envs/*/bin/* | xargs -0 -r basename -a | sort -u
+    # Fast path for rehash: skip filtering and link resolution.
+    # Discover executables directly from the versions directory and
+    # deduplicate their basenames. Use bash 3.2 conventions to keep code portable.
+    for path in \
+      "$versions_dir"/*/bin/* \
+      "$versions_dir"/*/envs/*/bin/*
+    do
+      printf '%s\n' "${path##*/}"
+    done | sort -u
     shopt -u dotglob nullglob
   fi
   exit 0
--- a/libexec/pyenv-versions
+++ b/libexec/pyenv-versions
@@ -45,12 +45,10 @@ if [[ -n "$executables" ]]; then
     # Fast path for rehash: skip filtering and link resolution.
     # Discover executables directly from the versions directory and
     # deduplicate their basenames. Use bash 3.2 conventions to keep code portable.
-    for path in \
-      "$versions_dir"/*/bin/* \
-      "$versions_dir"/*/envs/*/bin/*
-    do
-      printf '%s\n' "${path##*/}"
-    done | sort -u
+    paths=( "$versions_dir"/*/bin/* "$versions_dir"/*/envs/*/bin/* )
+    if [ "${#paths[@]}" -gt 0 ]; then
+      printf '%s\n' "${paths[@]##*/}" | sort -u
+    fi
     shopt -u dotglob nullglob
   fi
   exit 0


native-api (commented):
1. What's the performance difference in realistic scenarios (can use scripts from https://github.com/pyenv/pyenv/pull/3411#issuecomment-3988017858 to simulate them)? The only reason we've implemented this is user complaints about `pyenv rehash` poor performance.
2. Do we need to support the OSes this backwards and broken? A lack of the necessary switches for reliable scripting is a major PINA, source of unfixable bugs, security vulnerabilities and a performance hit. You can just install GNU coreutils, you know!
    * we use these switches elsewhere, too, and [planning to replace the `realpath` builtin with `readlink -f`](https://github.com/pyenv/pyenv/pull/3389)

timotheosh:
I swiched to bash 3.2 compatible arrays and squeezed out some performance.

Using the profiling tools from #3411, I measured a 3% regression (+6.2 ms) on the standard fixture (3 versions, 12 virtualenvs, 320 executables) and an 11% regression (+27.4 ms) on the large fixture (60 virtualenvs, 1466 executables). The --executables output is identical in both cases. In exchange, the fast path works on illumos and OpenBSD without requiring GNU xargs or basename.

```
pyenv-tune bench
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
shell:      zsh
iterations: 20
flags:      (none — default init)
targets:    upstream portable

▸ Benchmarking: upstream/
  warming up...
  running 20 iterations...
  5/20
  10/20
  15/20
  20/20

  upstream/
    min:    181.3 ms
    median: 202.2 ms
    mean:   201.5 ms
    p95:    208.5 ms
    max:    211.3 ms

▸ Benchmarking: portable/
  warming up...
  running 20 iterations...
  5/20
  10/20
  15/20
  20/20

  portable/
    min:    204.5 ms
    median: 208.5 ms
    mean:   209.1 ms
    p95:    214.0 ms
    max:    217.1 ms

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
comparison (median)
  upstream: 202.2 ms
  portable: 208.5 ms
  delta:    6.2 ms slower (3%)

```

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
shell:      zsh
iterations: 20
flags:      (none — default init)
targets:    upstream portable

▸ Benchmarking: upstream/
  warming up...
  running 20 iterations...
  5/20
  10/20
  15/20
  20/20

  upstream/
    min:    203.6 ms
    median: 230.9 ms
    mean:   226.0 ms
    p95:    244.2 ms
    max:    246.4 ms

▸ Benchmarking: portable/
  warming up...
  running 20 iterations...
  5/20
  10/20
  15/20
  20/20

  portable/
    min:    248.3 ms
    median: 258.4 ms
    mean:   257.7 ms
    p95:    264.6 ms
    max:    268.2 ms

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
comparison (median)
  upstream: 230.9 ms
  portable: 258.4 ms
  delta:    27.4 ms slower (11%)
```

native-api:
> Using the profiling tools from #3411, I measured a 3% regression (+6.2 ms) on the standard fixture (3 versions, 12 virtualenvs, 320 executables) and an 11% regression (+27.4 ms) on the large fixture (60 virtualenvs, 1466 executables).

With us [receiving complaints about a 100 ms startup delay](https://github.com/pyenv/pyenv/issues/2918), that's not good.

timotheosh:
Using the profiling tools from #3411, I benchmarked several portable alternatives (shell loop, array expansion, and sed). The submitted implementation was the fastest portable option I could find, with a 3% regression (+6.2 ms) on the standard fixture (3 versions, 12 virtualenvs, 320 executables) and an 11% regression (+27.4 ms) on the large fixture (60 virtualenvs, 1466 executables).

It is also worth noting that this regression is isolated to pyenv-versions --executables; it is not equivalent to a 3–11% regression in shell startup as a whole.

Additionally, this regression only exists because the GNU-specific optimization introduced in 8037f226 broke compatibility on illumos and OpenBSD.

native-api:
CC @pyenv/pyenv-core-maintainers @pyenv/pyenv-core-committers @VorpalBlade @aalbaali @varikin @dg-pb @swarajban @brandonrwin @Sheile 

Any thoughts on this? Is supporting OpenSolaris derivatives and BSD flavors with more archaic stock utils worth the slowdown?

dg-pb:
Couldn't we simply branch out and only use slower path only when needed?

Something like:
```bash
if xargs -r </dev/null true 2>/dev/null; then
    echo "Supports -r"
else
    echo "Does NOT support -r"
fi

if basename -a /etc/passwd /etc/group >/dev/null 2>&1; then
    echo "Supports -a"
else
    echo "Does NOT support -a"
fi
```

timotheosh:
I’m open to this approach. A more prudent implementation may be to detect the platform during make—for example with uname -s—and generate or persist the appropriate fast-path choice, rather than probing utility capabilities every time pyenv-versions --executables runs.

The main reason I did not take that route in the second revision is that this repository places a strong emphasis on compatibility with rbenv. The current PR remains consistent with the implementation style used by rbenv-versions, which uses Bash parameter expansion for pathname handling rather than relying on GNU-specific utility options.

That said, I’m open to preserving the optimized path on platforms where those options are available and selecting the portable path only where needed. A make-time capability or platform check may be preferable to probing on every invocation.

VorpalBlade:
This mentions using bash specific code in the PR description. Will the code in question always run under bash (or what happens if the user uses zsh for example)?

native-api:
@VorpalBlade Pyenv is written in Bash. It needs Bash to be present on the system but you can use a different shell.

native-api:
> A more prudent implementation may be to detect the platform during make—for example with uname -s—and generate or persist the appropriate fast-path choice, rather than probing utility capabilities every time pyenv-versions --executables runs.

That sounds promising. Generating/loading the cache seems like a job for either the `pyenv` loader or `pyenv init` (if shell integration is enabled, we can add some envvars to the user's shell).

It however boils down to how much expense that extra machinery would incur. We're talking about dozens of milliseconds difference. A couple of extra forks can ruin all the effect.

Note also that Pyenv uses e.g. `xargs -0` elsewhere, too. We do not test changes in those niche OSes. Making everyone suffer due to a bunch of folks set in their ways seems unfair -- and detrimental to the projects's future.

timotheosh:
This has more to do with technical diversity than with users being “set in their ways.” There are real production environments using illumos and OpenBSD because those platforms solve technical problems differently from Linux.

That said, I agree that Linux, macOS, and FreeBSD users should not pay a performance penalty for platforms they are not using. There may be a narrower compatibility path that preserves the current fast path everywhere else.

The incompatibilities are:

illumos: native xargs lacks -r, and native basename lacks -a
OpenBSD: native xargs -0 -r works, but native basename lacks -a

illumos already provides GNU variants under /usr/gnu/bin, so pyenv could select /usr/gnu/bin/xargs and /usr/gnu/bin/basename explicitly without changing the user’s global PATH.

OpenBSD provides GNU coreutils package with g-prefixed names such as gbasename; the OpenBSD package intentionally prefixes the binaries to distinguish them from the native utilities. Since native OpenBSD xargs already supports the required options here, only basename would need substitution.

One possible path would be:

retain the current command by default;
allow platform-specific command names or paths for xargs and basename;
require GNU coreutils only on platforms where the native implementation is insufficient.

That avoids imposing the portable Bash slowdown on existing platforms while keeping the compatibility logic narrowly scoped.

VorpalBlade:
> @VorpalBlade Pyenv is written in Bash. It needs Bash to be present on the system but you can use a different shell.

Sure, i just wanted to make sure this wasn't code loaded into the shell completion for other shells or similar.

native-api:
> The main reason I did not take that route in the second revision is that this repository places a strong emphasis on compatibility with rbenv.

That note is obsolete. We've not been syncing with Rbenv for years. I've removed the item from the template.

native-api:
As this PR's suggestion currently stands, it's not about a single incompatibility but about adding support for OpenSolaris/illumos and OpenBSD -- because those OSes were never officially supported.

Whose job it is to add that support and how it should be approached depends on the rationale for the current situation that @timotheosh described in https://github.com/pyenv/pyenv/pull/3506#issuecomment-5205244767 .

My theories for why they did not add backward-compatible features that are absolutely required for reliable shell scripting and have been out for a few decades now are the following:
1. They are "set in their ways", or rather, uninterested in portability aka supporintg running any 3rd-party software that is not specifically built with their OSes in mind -- for whatever reasons
    * In this case, a user who wants to run such software anyway is acting in outright defiance of the OS maintainers' wishes. Which means that the maintainers may choose to go as far as actively hamper such a user's efforts -- like making them run afoul of security checks that prohibit tampering with the execution environment.
2. They do not think highly of shell scripting and don't give that area any attention, likely having obsoleted it and the related tools (officially or in their minds) in favor of something else
    * In this case, the related OS packages are effectively EOL and are not expected to be used by anything production-grade. A user who needs working versions of these tools is expected to roll out their own ecosystem and run their software that needs them inside it.
3. Their highest priority is indefinitely keeping support for a number of existing legacy (read: untestable and with code maintained on a shoestring budget) production systems. As such, they believe they cannot afford to make any changes to the tools whatsoever, even backward-compatible ones, out of fear that somewhere, something might just break as a result
    * In this case, the entire OS is effectively a frozen-in-time EOL environment that is not expected to be able to run anything designed for newer systems. A user who needs that is on their own.

timotheosh:
I don't think the rationale for the design decisions made by OpenBSD or illumos needs to be resolved here.

The intent of this PR is simpler: the proposed implementation requires no external tooling beyond what pyenv already requires, and works with Bash 3.2+. It avoids depending on extensions provided by a particular implementation of xargs or basename.

That also seems consistent with the heritage of the *env tools. rbenv-versions, for example, uses Bash parameter expansion for pathname manipulation rather than requiring GNU-specific variants of those utilities.

There is a measurable performance trade-off, which I've benchmarked and documented. Whether that trade-off is acceptable is ultimately a project decision, but I think portability through the existing Bash dependency is preferable to adding userland-specific requirements.

native-api:
Here's the situation. I hope this clarifies why the PR as it is now screams "big trouble" to me. If you have a way to rework it to address these concerns, I'll be happy to merge it.

1. We're having user complaints about performance. We are losing users over performance issues.  
In particular, 2 scenarios seem to be performance-critical: shell startup time and `PROMPT_COMMAND` time.
`uv` provides strong competition here. As a shell-based project, we cannot complete with `uv`, a monolithic compiled tool, on performance. But we can retain users by providing satisfactory performance and flexibility/extensibility unachievable for a monolithic compiled tool.
We seem to have mostly achieved satisfactory performance in the above scenarios but any noticeable degradation of it would likely be a pain point for the users.
Plus, we don't yet have a framework to reliably measure, monitor and optimize performance, only a few ad-hoc tools.
So this is probably the worst time to be suggesting a change with 10% performance degradation in one of those critical scenarios.
2. The features we're using are not in "some implementations", they are in all the mainstream implementations. OpenBSD and illumos are the odd ones out. The lack of features necessary for reliable shell scripting has been a major PINA for shell script developers since there was shell scripting. Thus we have zero reason to condone such neglect from distribution providers or attempts to push the resulting trouble onto us. Unfixable bugs aside, in Pyenv's case, security vulnerabilities are also a concern.
It's a one-time 2 lines fix for you, but it'll be an ongoing task to check everything against 2 obscure environments for us.
3. I've brought up possible OS maintainers' stance because depending on it, the best way to fix this is not in Pyenv codebase:
  * In case 1, the best would be to make an OS-specific package for Pyenv that would add compliance with all the OS specifics and local requirements -- 'cuz there are likely to be much more than just command-line tools.
  * In case 2, since you mentioned there are GNU tools available outside of PATH, the easiest would probably be to run Pyenv with them prepended to `PATH`.
  * In case 3, both of the above may apply because CPython is "newer" software and may also require OS-specific downstream patches.

PR #3508, opened 2026-08-04
fix: the realpath_builtin function contains a use-af... in realpath.c

anupamme:
## Summary
Fix high severity security issue in `src/realpath.c`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `src/realpath.c:15` |
| **Assessment** | Likely exploitable |

**Description**: The realpath_builtin function contains a use-after-free vulnerability where the realbuf pointer is freed but not set to NULL. In subsequent loop iterations, if the same memory address is reused by the allocator, the dangling pointer could be dereferenced, leading to memory corruption.

## Evidence

**Exploitation scenario**: Attacker provides multiple path arguments to trigger the realpath_builtin function, crafting paths that cause realpath() to fail on some iterations.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a local CLI tool - exploitation requires the attacker to control command-line arguments or input files.

## Changes
- `src/realpath.c`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

---
## Summary by cubic
Reverts the previous `realbuf = NULL` change in `realpath_builtin` (`src/realpath.c`) because it was unnecessary. `realbuf` is overwritten by `realpath(p, NULL)` on each loop iteration, so there is no reachable use-after-free; behavior is unchanged.

<sup>Written for commit c969a79c492ead49809489d939a1861ba63c4e9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3508?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/realpath.c
+++ b/src/realpath.c
@@ -21,6 +21,7 @@ int realpath_builtin(WORD_LIST *list)
 		} else {
 			printf("%s\n", realbuf);
 			free(realbuf);
+			realbuf = NULL;
 		}
 	}
 	return es;
--- a/src/realpath.c
+++ b/src/realpath.c
@@ -21,7 +21,6 @@ int realpath_builtin(WORD_LIST *list)
 		} else {
 			printf("%s\n", realbuf);
 			free(realbuf);
-			realbuf = NULL;
 		}
 	}
 	return es;


src/realpath.c
native-api, line 17: In each successive iteration, `readbuf` is unconditionally overwritten -- so the dangling value can never be used. According to https://linux.die.net/man/3/realpath, it's set either to a result or to `NULL` (if there's a failure).
If the allocator reuses an existing address -- so what? It now is a different, independently allocated, and initialized with a new value, block.

Looks like your AI thinks that if a deallocated block has been reused and allocated to something else, it's still dangerous to use because we cannot say without context, by just looking at a pointer, if it's a use-after-free or not.
That AI needs to track data flow. If a variable is reassigned, it doesn't matter if the new value happens to be the same as old or not -- for the purpose of the code logic, it's a new value.

native-api: I've checked if it's a current best practice to always null freed pointers in C.
According to https://stackoverflow.com/questions/1879550/should-one-really-set-pointers-to-null-after-freeing-them and https://www.reddit.com/r/C_Programming/comments/15byc2b/should_you_set_freeed_pointers_to_null, it's a contested topic.

anupamme: You're right on both counts, thank you.

The data-flow analysis confirms there is no reachable use-after-free here: `realbuf` is unconditionally overwritten by `realpath(p, NULL)` at the top of every loop iteration, so the freed value is never read. If the loop does not iterate again, execution reaches `return es;` and `realbuf` is not accessed there either.

On the defensive-null idiom: as the linked threads show, the practice is contested. It can catch double-free bugs in longer-lived or aliased contexts, but in a tight loop where the variable is immediately reassigned (or goes out of scope), it adds noise with no real safety gain.

Reverted in c969a79c.

PR #3511, opened 2026-08-05
pyenv-binary: check FreeBSD system libraries

macayu17:
### Prerequisite

* [x] Please consider implementing the feature as a hook script or plugin as a first step.
    * This is entirely inside the `pyenv-binary` plugin. Nothing in the core or `python-build` changes.
* [x] Please consider contributing the patch upstream to rbenv, since we have borrowed most of the code from that project.
    * Not applicable. This builds on `python-build`, which rbenv does not have.
* [x] My PR addresses the following pyenv issue (if any)
    * Part of #2334. It does not close the issue.

### Description

- [x] Here are some details about my PR
    * Generated binary definitions now use `ldconfig -r` on FreeBSD and normalize the resolved paths to sonames before the existing exact dependency check. Other systems keep the existing `ldconfig -p` behavior.
    * FreeBSD does not support `ldconfig -p`, so generated definitions previously skipped dependency validation even when its library cache was available.

### Tests

- [x] My PR adds the following unit tests (if any)
    * Added a regression test with one library present in a FreeBSD cache and one missing.
    * The full `pyenv-binary` suite completed under Bash 3.2 and 4.1, with GNU mode enabled and disabled. Each run had 47 passing tests and the existing `ldconfig -p` integration skip.
    * A FreeBSD 14.4 VM run generated a definition that checked the actual cache entries for `libc.so.7`, `libedit.so.8`, and `libtinfow.so.9`.

---
## Summary by cubic
Enable required-library checks on FreeBSD by reading `ldconfig -r` and normalizing entries, and add version prefix resolution in `pyenv-binary-package-name`. Linux keeps `ldconfig -p`; when the cache can’t be read, we show “ldconfig cache unavailable”.

- **New Features**
  - `pyenv-binary-package-name` resolves version prefixes via `pyenv-latest -f -k` before generating names.

- **Bug Fixes**
  - FreeBSD: parse `ldconfig -r`, handle spaced or unspaced `=>`, strip paths to sonames, then exact-match required libs.

<sup>Written for commit a6ef1e70baa7b4b9a6a04fb7eab41007fdf769ef. Summary will update on new commits.</sup>

[...]

PR #3516, opened 2026-08-10
pyenv-binary: check patchelf early and add --verbose

macayu17:
### Prerequisite
    * This is implemented entirely inside the `pyenv-binary` plugin. Nothing in the core or `python-build` changes.
  * [x] Please consider contributing the patch upstream to rbenv, since we have borrowed most of the code from that project.
    * Part of #2334. It does not close the issue.

  ### Description
    * Generated definitions now check for `patchelf` after checking system libraries and before calling `install_package`, so a missing prerequisite is reported before the archive is downloaded.
    * `package` now calls `pyenv-install -v` so build progress is visible.
    * This follows the feedback from testing #3511.
    * `package` now accepts `-v` and forwards it to `pyenv-install` so build progress can be enabled when needed.

  ### Tests
    * Added coverage that runs a generated definition without `patchelf` and checks that it fails before installation.
    * Added coverage that checks the arguments passed from `package` to `pyenv-install`.
    * The full `pyenv-binary` suite passed under Bash 3.2 and Bash 4.1 with GNU tools enabled and disabled. Each run had 50 tests with one existing `ldconfig -p` integration skip.

---
## Summary by cubic
Adds a fail-fast `patchelf` check to generated installers and makes build progress opt-in via `-v`/`--verbose` in `pyenv-binary`. Prevents unnecessary downloads and only shows progress when requested; stabilizes tests across CI and Docker.

- **New Features**
  - Generated installers check for `patchelf` before downloading; they fail early with a clear message if missing.
  - `pyenv-binary-package` accepts `-v`/`--verbose` and passes `--verbose` to `pyenv-install`; completions and README updated.

- **Bug Fixes**
  - Test helpers support here-doc inputs and glob assertions, and show whitespace-aware diffs to reduce flakiness across environments.

<sup>Written for commit 9962f8cb3960d1fff99d84b5753d288cc4ac48b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/libexec/pyenv-binary-package
+++ b/plugins/pyenv-binary/libexec/pyenv-binary-package
@@ -2,7 +2,7 @@
 #
 # Summary: Create an installable binary package from a Python version
 #
-# Usage: pyenv binary package <version>[:<entry>] --archive-base-url <url>
+# Usage: pyenv binary package [-v|--verbose] <version>[:<entry>] --archive-base-url <url>
 #
 #   Installs <version> from source under a separate entry name, saves it as a
 #   binary package, then emits a python-build definition for that package.
@@ -13,6 +13,7 @@
 #   <entry>               The optional name to build under and install the
 #                         binary as. If omitted, a name is generated from the
 #                         current platform, platform version and architecture.
+#   -v,--verbose          Show build progress from `pyenv install'.
 #   --archive-base-url <url>
 #                         Where the archive will be hosted; the definition
 #                         downloads it from <url>/<entry>.tar.gz.
@@ -23,23 +24,29 @@ set -e
 # Provide pyenv completions
 if [ "$1" = "--complete" ]; then
   echo --archive-base-url
+  echo --verbose
   exec pyenv-install --list --bare
 fi
 
 spec=""
 archive_base_url=""
+verbose=""
 
 while [ $# -gt 0 ]; do
   case "$1" in
   --archive-base-url )
-    [ $# -ge 2 ] || { echo "pyenv-binary: --archive-base-url needs a value" >&2; exit 1; }
-    archive_base_url="$2"; shift 2 ;;
+    [ $# -lt 2 ] && { echo "pyenv-binary: --archive-base-url needs a value" >&2; exit 1; }
+    archive_base_url="$2"; shift ;;
+  -v|--verbose)
+    verbose=1 ;;
   -* )
     echo "pyenv-binary: unknown option \`$1'" >&2; exit 1 ;;
   * )
     [ -z "$spec" ] || { echo "pyenv-binary: unexpected argument \`$1'" >&2; exit 1; }
-    spec="$1"; shift ;;
+    spec="$1"
+    ;;
   esac
+  shift
 done
 
 if [ -z "$spec" ] || [ -z "$archive_base_url" ]; then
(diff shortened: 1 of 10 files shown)


plugins/pyenv-binary/libexec/pyenv-binary-package
```diff
@@ -78,7 +78,7 @@ if [ "$os" = "Darwin" ]; then
 fi
 
 # `pyenv install' puts a `<version>:<alias>' build under versions/<alias>.
-pyenv-install "$spec"
+pyenv-install -v "$spec"
```
native-api: I meant adding `-v` to `package` that would cause running `pyenv install` with `-v`.

macayu17: Changed `package` so `-v` is opt-in and forwarded to `pyenv-install`. Without `-v`, it uses the default install output. Added coverage for both paths and completion.

native-api:
Refactored the tests to be more human-readable.
As part of that, added a few other long-desired test features.

When adding an ability to supply output to `assert_success/failure` from stdin, I discovered that a number of tests were already "using" this "feature" (which was effectively a no-op) -- so they now needed a fix!

macayu17:
Thanks for merging this and for cleaning up the tests. I reviewed the changes to the package tests and shared assertions, and I’ll follow that pattern in the next work.

native-api:
> I reviewed the changes to the package tests and shared assertions

The "shared assertions" in a loop save heavy duplication between tests. They are problematic in that upon failure, it's unclear from the output which of the assertions failed. I had to run `bats --trace` to see that.
If you come up with a better solution, I'm all for that.

native-api:
> They are problematic in that upon failure, it's unclear from the output which of the assertions failed.

That said, normally, assertion messages contain parts of the invocation to deduce that.
But in this particular case, the invoked program failed with an error message that didn't contain any hints of which invocation it was from.

So this problem is not acute. Unlike e.g. the fixed problem with `assert_equal`: I've been pretty tired for some time from having to play "spot the difference", but this time, I ran into a case that was outright impossible: the difference ended up being two trailing spaces on an intermediate line!

macayu17:
Got it. I meant the assertion helpers generally, but I see the trade-off with the looped cases. They reduce duplication, but the failing invocation is less obvious. I’ll keep that in mind when writing similar tests.

native-api:
> I meant the assertion helpers generally

That's just your old standard refactoring. Any DRY principle violation is a problem whenever you need to make a further change.

native-api:
> I've been pretty tired for some time from having to play "spot the difference", but this time, I ran into a case that was outright impossible

Hilariously, just yesterday, I watched a [Super Mario Maker 2 Youtube video](https://www.youtube.com/watch?v=YKHqEDFa974) where the player also got a level which was a nigh-impossible "spot the difference" challenge! Talk about a coincidence!! 😆

PR #3520, opened 2026-08-15
pyenv-binary: record direct system dependencies in Linux/FreeBSD

macayu17:
### Prerequisite

    - Part of #2334. It does not close the issue.

  ### Description

    - `pyenv binary save` now records only direct shared-library dependencies in package metadata.
    - For ELF files, it reads `DT_NEEDED` entries with `readelf` and matches them against the paths resolved by `ldd`. Libraries inside the packaged prefix and transitive dependencies are excluded.
    - The existing macOS `otool` path is unchanged.
    - The command fails before creating an archive when `readelf` is unavailable.

  ### Tests

    - Added coverage for a missing `readelf` executable.
    - Updated the dependency test to distinguish direct, transitive, packaged, and system libraries.
    - The full `pyenv-binary` suite passed with Bash 3.2 and Bash 4.1 using both BusyBox and GNU tools: 200 passing tests and 4 expected `ldconfig -p` skips.
    - Verified on FreeBSD 15.1 that direct dependencies are retained and a transitive dependency is excluded.

---
## Summary by cubic
Records only direct system shared-library dependencies in `pyenv-binary` metadata on ELF platforms. Previously all `ldd`-resolved libraries were recorded; now `pyenv-binary-save` filters by `readelf` DT_NEEDED and exits early if `readelf` is unavailable.

- Non-macOS builders must have `readelf` on PATH (install via `binutils`); the command fails before archiving if it is missing.
- Tests: add a failure case for missing `readelf`, update dependency expectations to direct-only, and stub `readelf` in package tests.
- CI: install `binutils` in the test Dockerfile and broaden `path_without` essentials; the script comments document the `readelf`/`ldd` formats parsed.

<sup>Written for commit 7e8a1b740f1e5640248e2fb70787f9b3a759028c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/libexec/pyenv-binary-save
+++ b/plugins/pyenv-binary/libexec/pyenv-binary-save
@@ -92,12 +92,16 @@ fi
 if [ "$os" = "Linux" ]; then
   libc="$(getconf GNU_LIBC_VERSION 2>/dev/null || true)"
 fi
+if [ "$os" != "Darwin" ] && ! LC_ALL=C readelf --version >/dev/null 2>&1; then
+  echo "pyenv-binary: need readelf to inspect shared libraries" >&2
+  exit 1
+fi
 
 # List the external shared libraries the install links against: those that
 # resolve outside its own prefix, so they must already exist on the target.
 # The interpreter plus every bundled shared object are inspected.
 system_deps() {
-  local f
+  local f needed
   {
     for f in "${prefix}"/bin/python*; do
       [ -e "$f" ] && printf '%s\n' "$f"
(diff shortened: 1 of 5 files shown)


native-api (approved):

plugins/pyenv-binary/libexec/pyenv-binary-save
```diff
@@ -111,8 +115,15 @@ system_deps() {
       otool -L "$f" 2>/dev/null | tail -n +2 | awk -v pfx="${prefix}/" \
         '$1 !~ /^@/ && substr($1, 1, length(pfx)) != pfx { print $1 }'
     else
-      ldd "$f" 2>/dev/null | awk -v pfx="${prefix}/" \
-        '$2 == "=>" && $3 ~ /^\// && substr($3, 1, length(pfx)) != pfx { print $1 }'
+      needed="$(LC_ALL=C readelf -dW "$f" 2>/dev/null | awk \
+        '$2 == "(NEEDED)" || $2 == "NEEDED" { sub(/^.*\[/, ""); sub(/\].*$/, ""); print }' | tr '\n' ' ')"
```
native-api, line 121: Need to add a comment on what format we're parsing here.
Which `readelf` versions produce what format? Is it documented anywhere?

plugins/pyenv-binary/libexec/pyenv-binary-save
```diff
@@ -111,8 +115,15 @@ system_deps() {
...
+      LC_ALL=C ldd "$f" 2>/dev/null | awk -v needed="$needed" -v pfx="${prefix}/" '
+        BEGIN {
+          split(needed, deps, " ")
+          for (i in deps) direct[deps[i]] = 1
+        }
+        $1 in direct && $2 == "=>" && $3 ~ /^\// && substr($3, 1, length(pfx)) != pfx { print $1 }
+      '
```
native-api, line 130: Same as above on expected `ldd` format.

native-api (approved):

PR #3522, opened 2026-08-23
CI: test on all runners when the build logic changes

### LINKED ISSUE #3304: CI: test everywhere if the build logic is changed
**Describe the intended use case**
* What your general environment is if it's relevant to the feature request and is not a generic console with a typical Pyenv installation (CI, server with a custom setup, cloud environment, IDE)
In PR CI checks, if there are changes to the build logic but no script changes, they are only tested against `macos-latest` and `ubuntu-latest`
* What you are trying to achieve
Those should be tested against all available Ubuntu and MacOS versions

**Describe the solution you'd like**
Add to the `ubuntu-build` and `macos-build` jobs additional jobs that would run if `plugins/python-build/bin/python-build` is changed, in other (non-latest) Ubuntu/MacOS runners.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

sujeito-operator:
### Prerequisite
* [x] My PR addresses the following pyenv issue
  - Closes https://github.com/pyenv/pyenv/issues/3304

### Description

`ubuntu_build` and `macos_build` only exercise `ubuntu-latest` and `macos-latest`.
Changes to a *version definition script* are already tested on every runner by
`modified_scripts`, but changes to `python-build` itself — which affect every version
on every platform — were not covered anywhere beyond the two `latest` runners.

The two workflows are merged into `.github/workflows/build.yml`, which gains a
`discover_build_logic_change` job. That job **selects the `os` matrix** for
`macos_build` and `ubuntu_build`: the `*-latest` list normally, the full list of
available runners when the PR touches the build logic. The build steps themselves live
in one place, `.github/actions/build-python`, used by both jobs.

> **Revised twice after review.** The first version added a second, copy-pasted build job
> to each of the two workflows
> ([1](https://github.com/pyenv/pyenv/pull/3522#discussion_r3839504119),
> [2](https://github.com/pyenv/pyenv/pull/3522#discussion_r3839530296),
> [3](https://github.com/pyenv/pyenv/pull/3522#discussion_r3839543169)): driving the
> existing job's matrix instead, and merging the workflows, removed all of that
> duplication. The second round
> ([4](https://github.com/pyenv/pyenv/pull/3522#discussion_r3839911596),
> [5](https://github.com/pyenv/pyenv/pull/3522#discussion_r3839877458)) restored the full
> Python version list, made the discover job `pull_request`-only, and fused the two build
> jobs' steps into a composite action. **+137/−86** against `master`, in four separable
> commits.

Design notes:

* **The two runner lists are alternatives, not additions.** `*-latest` is an alias for one
  of the specific environments and which one is not under our control, so running both
  lists would build that environment twice — and would need watching whenever GitHub
  repoints the alias. The specific lists match `modified_scripts_build.yml`
  (`ubuntu-22.04`, `ubuntu-22.04-arm`, `ubuntu-24.04`, `ubuntu-24.04-arm`;
  `macos-14`, `macos-15`, `macos-15-intel`, `macos-26`, `macos-26-intel`).
* **The version axis does not narrow.** Every non-EOL line is built on every runner in the
  selected list. An earlier revision narrowed it to `3.10`/`3.14` to keep the job count
  down; that was rejected, correctly, because installing the latest CPython in each line
  is the primary use case. **The number that buys: a build-logic PR is 25 macOS + 20
  Ubuntu jobs**, against 5 + 5 for every other PR.
* **Detection uses a `git diff` pathspec** and asks only whether the result is empty, so
  the changed-file list is never parsed or piped. This replaced a
  `git diff -z | tr | grep -qxF` pipeline that had a real bug: `grep -q` exits at the
  first match, `git diff` takes `SIGPIPE`, and `pipefail` — on by default in Actions —
  made the condition false. On a large enough diff it silently reported "not changed"
  and skipped every job this PR adds.
* **The discover job is `pull_request`-only.** On `push` it is skipped; the build jobs run
  on `!cancelled() && !failure()` — success *or* skipped, but not failure — and fall back
  to the `*-latest` list, which is what `master` runs today. `!cancelled()` alone would
  have run them through a *failed* discovery on an empty matrix.
* **Both the workflow and the composite action are in the pathspec**, so a change to
  either is exercised by the PR that makes it — you should see the wide fan-out running
  below rather than having to take my word for it.

### Tests

CI config isn't reachable from the bats suite, so there is no unit test to add here.
What I did verify locally:

* **The detection step, against a real git repo, 13/13:** `python-build` changed → wide;
  this workflow changed → wide; `action.yml` added, changed, or deleted → wide (added is
  this PR's own case, where the path is absent from the base); **only a version definition
  script changed → not wide** (that's `modified_scripts`' job, and the case I most wanted
  to be sure of); unrelated file → not wide; `python-build` + unrelated → wide; a
  different workflow only → not wide; `plugins/python-build/bin/python-build.bak` → not
  wide (pathspec matching is exact, not a prefix); all three watched paths deleted → wide;
  a **5,000-file diff with `python-build` among the changes → wide**, which is the case
  the old pipeline got wrong (reproduced the old failure at 20,000 files before fixing
  it); and a tree where none of the watched paths exist on either side, which does **not**
  make `git diff` exit non-zero.
* **`actionlint` 1.7.7, diffed against `master`. No new class of finding** — the one on
  `build.yml` is the unknown `ubuntu-slim` label, which `master` already has seven of in
  `modified_scripts_build.yml` (actionlint's bundled runner list is simply older). Five
  `SC2086` infos disappear from the total, and **that is not an improvement**: those lines
  moved verbatim into `action.yml` and actionlint doesn't follow composite actions. I left
  them unquoted deliberately, so the diff shows the steps moving rather than changing.
* No CHANGELOG entry: the last 12 CI-only commits here don't carry one.

**Two things to check on your side, since I can't see either:**

1. The *job* names `macos_build` and `ubuntu_build` are unchanged, so the check names are
   unchanged — but this PR deletes `macos_build.yml` and `ubuntu_build.yml`. If branch
   protection or an integration refers to them by workflow or file name, that needs
   updating (or drop the workflow-merge commit).
2. Now that the steps are shared, collapsing the two jobs into a single `build` job with
   one combined matrix is about four more lines. I haven't, because it *would* rename
   every check — same question as above. Say the word.

I can't exercise the macOS or ARM runners from my end, so their first real run will be
the one in this PR. The runs on the latest push are sitting at `action_required` and need
approving before that fan-out is visible.

---
## Summary by cubic
Test all Ubuntu and macOS runners when build logic changes; otherwise run only on `ubuntu-latest` and `macos-latest`. This replaces the latest-only behavior, reduces platform regressions, and fixes a detection bug that sometimes skipped fanout on large diffs.

- Merge to a single `build` workflow and collapse the OS-specific jobs into one `build` job; remove `macos_build.yml` and `ubuntu_build.yml`. Check names change to `build`.
- Add `discover_build_logic_change` (PR-only) to select the OS matrix; on push it’s skipped and the `build` job falls back to `["macos-latest","ubuntu-latest"]`.
- Fanout triggers when any of these change: `plugins/python-build/bin/python-build`, `.github/workflows/build.yml`, `.github/actions/build-python/action.yml`. Detection uses a `git diff` pathspec to avoid the prior SIGPIPE false negatives.
- On fanout, run the mirrored runner list from `modified_scripts_build.yml` (`macos-14`, `macos-15`, `macos-15-intel`, `macos-26`, `macos-26-intel`, `ubuntu-22.04`, `ubuntu-22.04-arm`, `ubuntu-24.04`, `ubuntu-24.04-arm`) and build CPython `3.10`–`3.14` on each.
- Extract shared build steps into `.github/actions/build-python` with OS-specific prerequisites and consistent `PYENV_ROOT` usage; dump `config.log` on failure.

- Migration: Update branch protections and integrations to require the `build` check instead of `macos_build`/`ubuntu_build`.

<sup>Written for commit 282e5cfc2f60d00e26f176ad4dae64e08abe1e43. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3522?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sujeito-operator:
One correction to the last line of my description: the four workflow runs are sitting at `action_required` rather than running, since this is my first PR here and first-time contributor runs need a maintainer to approve them. So the self-triggering can't demonstrate itself yet — whenever you approve the run, `ubuntu_build_other_runners` and `macos_build_other_runners` should appear alongside the existing jobs, which is the check I'd most want you to see before merging.

.github/workflows/macos_build.yml
```diff
@@ -41,3 +41,73 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v7
+      - run: git fetch origin "$GITHUB_BASE_REF"
+      - id: check
+        shell: bash
+        # This workflow file is included so that changes to the jobs below are
+        # exercised by the very PR that makes them.
+        run: |
+          changed=false
+          if git diff "origin/$GITHUB_BASE_REF" --name-only -z | tr '\0' '\n' \
```
native-api: `tr '\0' '\n'` defeats the purpose of `git diff -z`. Both GNU and MacOS `grep` have `-z`.

.github/workflows/macos_build.yml
```diff
@@ -41,3 +41,73 @@ jobs:
...
+                        -e .github/workflows/macos_build.yml
+          then
+            changed=true
+          fi
+          echo "build logic changed: $changed"
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
+
+  macos_build_other_runners:
+    needs: discover_build_logic_change
+    if: needs.discover_build_logic_change.outputs.changed == 'true'
```
native-api: Rather than duplicating all the build logic, can we instead have `discover_build_logic_change` set/affect `matrix: os:` for the build jobs?

So that we have 2 lists: `*-latest` and a list of specific environments, and one of them is selected depending on the detection result.

This would also future-proof the script because we don't control which of the specific environments the `-latest` alias refers to. So as it is now, we'd have to either watch for its changes by Github, or the job for the specific environment corresponding to `latest` would run redundantly.

.github/workflows/ubuntu_build.yml
```diff
@@ -41,3 +41,74 @@ jobs:
         env:
           EXPECTED_PYTHON: ${{ matrix.python-version }}
         run: import os, sys ; assert sys.version.startswith(os.getenv("EXPECTED_PYTHON"))
+
+  # The job above only covers `ubuntu-latest`. When the build logic itself is
+  # changed -- as opposed to a version definition script, which
+  # `modified_scripts` already tests everywhere -- also test on the other
+  # available Ubuntu runners.
+  discover_build_logic_change:
```
native-api: Again, rather than duplicate, we can instead fuse `macos_build` and `ubuntu_build` together.
They are only separate for historical reasons because it was good enough at that time and Github Actions were new and unfamiliar.

Ideally, the build logic from `modified_scripts` should be used everywhere -- but I don't know if it's possible to reuse chunks of YAML between workflows.

native-api (commented):
I agree with AIs considerations, and also see other comments.

.github/workflows/macos_build.yml
sujeito-operator: You're right, and it turned out to be worse than a style point — thank you.

Rather than swapping in `grep -z` I've removed the parsing altogether. The two paths go to `git diff` as a pathspec and the step only asks whether the output is empty, so nothing splits the list and there is nothing left for `-z` to protect:

```bash
changed=false
if [[ $GITHUB_EVENT_NAME == pull_request ]] &&
   [[ -n $(git diff --name-only "origin/$GITHUB_BASE_REF" -- \
             plugins/python-build/bin/python-build \
             .github/workflows/build.yml) ]]
then
  changed=true
fi
```

That also disposes of the `grep -q` problem the bot flagged alongside you, which I checked rather than assumed and which is real: `grep -q` exits at the first match, `git diff` takes SIGPIPE, and `pipefail` — on by default in Actions — makes the whole condition false. On a 20,000-file diff with `python-build` among the changed paths, the old snippet reports `false` and skips every job this PR adds. It failed silently and in the "don't test" direction, which is the bad one.

Re-ran the detection cases against a real repo, 9/9: `python-build` changed; this workflow changed; **only** a version definition script changed (must *not* fire — that's `modified_scripts`' job); unrelated file; `python-build` + unrelated; a different workflow only; `plugins/python-build/bin/python-build.bak` (pathspec matching is exact, so it doesn't fire); a `push` event with no base ref; and the large-diff case above.

.github/workflows/macos_build.yml
sujeito-operator: Done — this is a better design than what I had, and it made the patch smaller: the PR is now +125/−86 against `master`, where the version you reviewed was +141/−0.

`discover_build_logic_change` emits the runner list and the existing build job consumes it, so the duplicated `*_other_runners` jobs are gone entirely:

```yaml
  discover_build_logic_change:
    runs-on: ubuntu-slim
    outputs:
      macos_os: ${{ steps.detect.outputs.macos_os }}
      macos_python_version: ${{ steps.detect.outputs.macos_python_version }}
      ...

  macos_build:
    needs: discover_build_logic_change
    strategy:
      fail-fast: false
      matrix:
        os: ${{ fromJson(needs.discover_build_logic_change.outputs.macos_os) }}
        python-version: ${{ fromJson(needs.discover_build_logic_change.outputs.macos_python_version) }}
    runs-on: ${{ matrix.os }}
```

The two lists are alternatives rather than additions, which is your point about `-latest`: when the build logic changes we run the specific environments *instead of* `macos-latest`, so whichever one it currently aliases gets built once rather than twice and we don't have to track what GitHub repoints it to.

Two consequences worth your eye:

* **The discover job no longer skips on `push`.** It has to run, because the build job now `needs:` it; on a push it takes the `*-latest` branch. That's one of the 9 cases I re-tested.
* **I kept the narrowed version axis**, so when the OS list widens the version list becomes `3.10`/`3.14`. That is a real trade and I'd rather name it than bury it: a `python-build` PR now gets 3.11–3.13 on no macOS runner at all, where previously it got them on `-latest`. In exchange it's 10 + 8 jobs instead of 25 + 20. Both lists are one line each in the same step — say the word and I'll make the version axis unconditional.

.github/workflows/ubuntu_build.yml
sujeito-operator: Done, and kept as its own commit so you can drop it with a single revert if you'd rather keep the two filenames.

`.github/workflows/build.yml` now holds one `discover_build_logic_change` feeding both `macos_build` and `ubuntu_build`. After the matrix change above, that discover job was the only thing still duplicated between the two files — so this removes the last copy, and leaves one self-reference in the pathspec instead of two.

The **job** names `macos_build` and `ubuntu_build` are unchanged, so the check names they produce are unchanged; the **workflow** name becomes `build`. One thing to check on your side, since I can't see it: if anything refers to these by workflow or file name rather than job name — branch protection, required checks, an external integration — this PR deletes `macos_build.yml` and `ubuntu_build.yml`.

> Ideally, the build logic from `modified_scripts` should be used everywhere -- but I don't know if it's possible to reuse chunks of YAML between workflows.

Three facts, since they decide how far that can go:

* **YAML anchors and aliases: not supported.** Actions parses each workflow independently and won't expand them. This is the mechanism people reach for first and it's the one that doesn't exist.
* **Composite actions: yes, and this is the one that fits.** A `.github/actions/<name>/action.yml` with `runs: using: composite` holds a block of *steps*, used from any job in the repo as `- uses: ./.github/actions/<name>`. What's duplicated between these jobs and `modified_scripts` is steps, so this is the right tool.
* **Reusable workflows: yes, for whole jobs.** `on: workflow_call` in the callee and `jobs.<id>.uses: ./.github/workflows/<file>.yml` in the caller, with typed `inputs`/`outputs`. A matrix can be passed in as a JSON string input. Heavier than a composite action here.

So the end state you're describing is reachable: extract each OS's build steps into a composite action and have both this workflow and `modified_scripts_build.yml` use it, leaving one copy of the prerequisites/build/check sequence per OS. I've left that out of this PR because it edits `modified_scripts_build.yml` and this one is meant to be about #3304 — happy to open it as a follow-up if you want it, or to fold it in here if you'd rather review it once.

`actionlint` 1.7.7 on the merged file reports the same two classes as `master` and no new one: the unknown `ubuntu-slim` label (`modified_scripts` already uses it) and SC2086 on the build steps carried over unchanged.

native-api: Composite action it is.
It also seems possible to fuse MacOS and Ubuntu build logic into one set of steps: they only differ in the dependency installation step which can be made conditional on the OS.

.github/workflows/macos_build.yml
native-api: I would rather not narrow down Python version list. Installing latest CPython versions in all current non-EOL version lines is Pyenv's primary use case. We cannot let that break.

Can you make `needs:` on build jobs conditional on `pull_request`? Then you can skip the discover job on `push`.

sujeito-operator: Both done in `3fc8731`, and you were right to stop me — thank you.

**The version list is unconditional again.** Every non-EOL line builds on every runner in the list, whichever list is selected, so nothing narrows when the OS axis widens. That also removed two of the four outputs the discover job was carrying:

```yaml
    outputs:
      macos_os: ${{ steps.detect.outputs.macos_os }}
      ubuntu_os: ${{ steps.detect.outputs.ubuntu_os }}
```

The bill for that is the one number worth having in front of you before you approve a run: when the build logic changes it is **25 macOS + 20 Ubuntu jobs**, against 10 + 8 under the narrowed axis and 5 + 5 today. This PR touches `build.yml`, which is in the pathspec, so the PR itself is one of those 45-job runs.

**`needs:` conditional on `pull_request`.** `needs:` itself can't take an `if:`, but the effect you asked for is available and the discover job is now `pull_request`-only:

```yaml
  discover_build_logic_change:
    if: github.event_name == 'pull_request'
...
  macos_build:
    needs: discover_build_logic_change
    if: ${{ !cancelled() && !failure() }}
    strategy:
      matrix:
        os: ${{ fromJson(needs.discover_build_logic_change.outputs.macos_os || '["macos-latest"]') }}
```

Two parts, because the default `if:` on a job is `success()` and a *skipped* dependency doesn't satisfy it:

* `!cancelled() && !failure()` runs the build jobs when discovery **succeeded or was skipped**, and not when it failed. Plain `!cancelled()` would have run them through a failed discovery on a stale or empty matrix, which is the failure mode worth avoiding here.
* A skipped job's outputs are the empty string, and empty is falsy in expressions, so `|| '["macos-latest"]'` supplies the push list. The fallback is the same list the `else` branch emits, so on `push` the matrix is what it is on master today.

This also removes the last reason the detection script needed to know about the event: the `[[ $GITHUB_EVENT_NAME == pull_request ]]` guard and the `if:` on the `git fetch` step are both dead now that the job is gated, so they're gone. One of the nine cases I'd tested was "a `push` event with no base ref" — that case can no longer reach the script at all, which is a better answer than handling it.

.github/workflows/ubuntu_build.yml
sujeito-operator: Done in `be4097d`, and it is one set of steps rather than two — `.github/actions/build-python/action.yml`, with the dependency install conditional on `runner.os` exactly as you describe:

```yaml
runs:
  using: composite
  steps:
    # The two OSes differ only in this step.
    - if: runner.os == 'macOS'
      shell: bash
      run: brew install openssl readline sqlite3 xz tcl-tk@8 libb2 zstd
    - if: runner.os == 'Linux'
      shell: bash
      run: sudo apt-get update -q; sudo apt install -yq make build-essential ...
```

and both jobs are now four lines:

```yaml
    steps:
      - uses: actions/checkout@v7
      - uses: ./.github/actions/build-python
        with:
          python-version: ${{ matrix.python-version }}
```

Checking your premise before relying on it: the dependency step was **not** the only difference. macOS took `PYENV_ROOT` from `$GITHUB_WORKSPACE` and called `pyenv` through the shim path; Ubuntu hardcoded `/home/runner/work/pyenv/pyenv` on a single step and called `bin/pyenv` directly; Ubuntu had a stray `pwd` and a `rehash`, macOS had the `config.log` dump on failure. **The fused version is the macOS shape, which is also what `modified_scripts_build.yml` already does for Ubuntu** — so it isn't a setup I invented, it's the one your other workflow has been running on Ubuntu all along. Ubuntu keeps `rehash` and gains the `config.log` dump, with `${TMPDIR:-/tmp}` so the glob resolves on Linux (today an Ubuntu build failure prints nothing).

**Three things to look at, since none of them are my call:**

1. **The jobs are still two, deliberately.** `macos_build` and `ubuntu_build` keep their names and their check names, and keep separate `os` matrices because the runner lists differ. Collapsing them into one `build` job with a combined matrix is now about four lines — but it renames every check, and I already owe you a question about whether branch protection names these. Say the word and I'll do it in the same commit.
2. **The action is build logic, so it's in the detection pathspec.** A change to `action.yml` now runs the wide matrix on the PR that makes it, the same self-reference `build.yml` already had. Re-ran the detection against a real repo, **13/13**, including the three new cases (`action.yml` added — which is this PR — changed, and deleted), the exact-pathspec case, the 20,000-file diff, and a tree where none of the watched paths exist, which does not make `git diff` exit non-zero.
3. **`actionlint` 1.7.7 shows no new class against master**, but I don't want to oversell that: five `SC2086` infos vanished from the count, and that is *not* an improvement — those five lines moved verbatim into `action.yml`, and actionlint doesn't follow composite actions. I left them unquoted on purpose so the diff shows the steps moving rather than changing. The one remaining finding on `build.yml` is `ubuntu-slim` as an unknown label, which master already has seven of in `modified_scripts_build.yml`.

> Ideally, the build logic from `modified_scripts` should be used everywhere

I looked, and it is closer than it sounds: `modified_scripts_build.yml`'s `macos_build` and `ubuntu_build` are also identical except for the dependency step — same envvars block, same build block, same micropython-aware version print and check, same shim cleanup at the end. `macos_build_bundled_dependencies` and `ubuntu_build_tar_gz` are the same steps again with one substitution each (a `brew-uninstall-cascade.sh` call; `_PYTHON_BUILD_FORCE_SKIP_XZ=1`). So a second composite action taking `python-version` plus a couple of optional inputs would collapse those four jobs the same way.

I've deliberately **not** done that here. It's a bigger change than this PR, it touches the workflow that gates every version-definition PR you merge, and it would bury the thing you're actually reviewing. Happy to open it as a separate PR on top of this one if you want it — but it should be yours to approve separately, not something that arrives inside a CI-matrix fix.

native-api: There are no branch protection rules referencing any Github Actions. So go ahead with merging `ubuntu_build` and `macos_build`.

sujeito-operator:
Thank you — that was the one thing I had no way to check from outside the repo.

It is already in the branch: `3d47a7c` fused the two workflows into `.github/workflows/build.yml`, and `be4097d` moved the shared steps into `.github/actions/build-python`. So the diff as it stands deletes `macos_build.yml` and `ubuntu_build.yml` and adds those two files — +137/−86 across four files.

The **job** names `macos_build` and `ubuntu_build` are unchanged, so the check names GitHub reports are unchanged; only the workflow name becomes `build`. With no branch protection referencing Actions, nothing else should notice the file rename.

The run on `be4097d` is green: 61 checks, 57 success, 4 skipped, none failed. Worth noting which run that is — this PR touches `build.yml`, which is in the detection pathspec, so it selected the *expanded* matrix rather than the `*-latest` one. The wide path is the one that just passed, not the cheap one.

Nothing outstanding from my side.

sujeito-operator:
Already in the branch — `be4097d`, pushed before your reply, so there is nothing outstanding on my side. `macos_build.yml` and `ubuntu_build.yml` are gone and `build.yml` is the only build workflow.

The run on that head is the part worth having in front of you: **45 build jobs, all green.** 25 macOS (`macos-14`, `macos-15`, `macos-15-intel`, `macos-26`, `macos-26-intel`) and 20 Ubuntu (`ubuntu-22.04`, `ubuntu-22.04-arm`, `ubuntu-24.04`, `ubuntu-24.04-arm`), each on the full 3.10-3.14 axis. That is the 25 + 20 I estimated for you earlier, now measured rather than predicted.

Four checks on that head read `skipped`, and I want to say plainly that none of them are this PR's: `macos_build`, `ubuntu_build`, `macos_build_bundled_dependencies` and `ubuntu_build_tar_gz` all belong to `modified_scripts_build.yml`, which this PR does not touch. They skip because no `python-build` version script changed here.

I checked the deleted files for the same reason rather than assuming: each held exactly one job, and both of those jobs now exist in `build.yml`. The fusion dropped no coverage.

sujeito-operator:
You said "go ahead with merging `ubuntu_build` and `macos_build`" and I answered that twice with the wrong thing — I read it as the workflow *files*, which were already fused, and told you there was nothing outstanding. You were answering the question I actually asked, about collapsing the two **jobs**. Two round trips that were mine to avoid.

Done in `282e5cf`. One job:

```yaml
  build:
    needs: discover_build_logic_change
    if: ${{ !cancelled() && !failure() }}
    strategy:
      fail-fast: false
      matrix:
        os: ${{ fromJson(needs.discover_build_logic_change.outputs.os || '["macos-latest","ubuntu-latest"]') }}
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@v7
      - uses: ./.github/actions/build-python
        with:
          python-version: ${{ matrix.python-version }}
```

The detection job emits a single `os` output instead of two, and the composite action was already picking its dependency step off `runner.os`, so nothing else had to move. −35/+16 in one file.

**The job count does not change**: 45 when the build logic changes (25 macOS + 20 Ubuntu) and 10 otherwise. The combined list is the concatenation of the two lists and the version axis is untouched.

**One thing I found while checking what the check names would become, which argues for this beyond your having asked for it.** I pulled the jobs of all three runs on `be4097d` rather than reading the flat list of 61 checks:

```
workflow 'build'             25  macos_build     success
                             20  ubuntu_build    success
                              1  discover_build_logic_change
workflow 'modified_scripts'   1  macos_build     skipped
                              1  ubuntu_build    skipped
                              1  macos_build_bundled_dependencies  skipped
                              1  ubuntu_build_tar_gz               skipped
```

`modified_scripts_build.yml` already has jobs named `macos_build` and `ubuntu_build`, so today two workflows emit check runs under the same two names — the flat list shows `26 macos_build` and `21 ubuntu_build` with no way to tell which workflow a given one came from. That is also where two of the four `skipped` checks I flagged for you earlier were hiding. After this commit the 45 are named `build`, and `macos_build`/`ubuntu_build` belong unambiguously to `modified_scripts`.

The run on `282e5cf` will need your approval again, same as the last one.

.github/workflows/ubuntu_build.yml
sujeito-operator: Done in `282e5cf` — one `build` job, job count unchanged at 45/10. Written up in [the comment below](https://github.com/pyenv/pyenv/pull/3522#issuecomment-5396947822), including a check-name collision with `modified_scripts_build.yml` that this happens to clear up.

sujeito-operator:
Those two P0s in the latest review are both from the in-progress test commits — `e5cfc23` commenting out `build`, and `ab73d16` adding `with:` to `discover_build_logic_change` — so they're expected rather than defects in the PR as proposed. Flagging in case the bot summary gets triaged later.

On the `with:` experiment itself: there's no `with` context on a plain `runs-on:` job — `with:` (and the context that reads it) only exists for `uses:` calls, i.e. reusable workflows and action steps — so `${{ with.full_os_matrix }}` resolves to empty and the key isn't valid there.

If the goal is to define the full runner list once and emit it from the detect step, job-level `env:` holding a JSON string does it without a reusable workflow, and it also lets the detect step emit the full matrix on a build-logic change (right now it writes an empty `os=`, which falls through to the two-runner fallback):

```yaml
  discover_build_logic_change:
    if: github.event_name == 'pull_request'
    runs-on: ubuntu-slim
    env:
      FULL_OS_MATRIX: '["macos-14","macos-15","macos-15-intel","macos-26","macos-26-intel","ubuntu-22.04","ubuntu-22.04-arm","ubuntu-24.04","ubuntu-24.04-arm"]'
    outputs:
      os: ${{ steps.detect.outputs.os }}
    steps:
      - uses: actions/checkout@v7
      - run: git fetch origin "$GITHUB_BASE_REF"
      - id: detect
        shell: bash
        run: |
          changed=false
          if [[ -n $(git diff --name-only "origin/$GITHUB_BASE_REF" -- \
                       plugins/python-build/bin/python-build \
                       .github/actions/build-python/action.yml \
                       .github/workflows/build.yml) ]]; then
            changed=true
          fi
          if [[ $changed == true ]]; then
            echo "os=$FULL_OS_MATRIX" >> "$GITHUB_OUTPUT"
          fi
```

The `build` job then reads it unchanged:

```yaml
    matrix:
      os: ${{ fromJson(needs.discover_build_logic_change.outputs.os || '["macos-latest","ubuntu-latest"]') }}
```

— full matrix when the build logic changed, two-runner fallback otherwise. Happy to push that onto the branch, or leave it to you if you're mid-experiment.

native-api:
@sujeito-operator Is it possible to define `FULL_OS_MATRIX` as a YAML list in the workflow file?

sujeito-operator:
The multiline form works — `ef4f1f5` expanded the matrix correctly: `discover_build_logic_change` is green and the `build` job fanned out to all 45 combinations (9 runners x 5 Python versions), 20 green and 0 failures so far with the rest still queued. So `os<<!` / `printf` writing the pretty-printed JSON is fine; nothing further needed there.

One thing worth catching before this merges, though: `1bbfd3f` ("(t) disable irrelevant") deletes four workflows, and three of them are outside what this PR touches —

- `.github/workflows/add_version.yml`
- `.github/workflows/no-response.yml`
- `.github/workflows/pyenv_tests.yml`

All three were still present at `282e5cf`, the last commit on the branch before the test commits. If the `(t)` commits get squashed in as-is, master loses the version-adding workflow, the stale-issue bot and the test workflow along with the CI matrix change.

The fourth, `modified_scripts_build.yml`, is a different case — it genuinely overlaps with what `build.yml` now does, so removing it may well be intended. If it is, it's probably worth doing in its own commit and saying so in the description, rather than leaving it inside a `(t)` one where it reads as scaffolding.

Happy to push a commit restoring the three, or to rebase the branch onto a clean history once you're done experimenting — whichever is less disruptive to what you have in flight.

PR #3528, opened 2026-08-29
pyenv-binary: support relocating macOS archives

macayu17:
### Prerequisite

      * Part of #2334. It does not close the issue.

  ### Description

      * `pyenv binary save` now records the original installation prefix in package metadata.
      * Generated macOS definitions require `otool` and `install_name_tool`, then pass the original and destination prefixes to `pyenv binary relocate`.
      * On macOS, relocation rewrites matching Mach-O dependencies, install IDs, and `LC_RPATH` entries to use `@rpath`, `@executable_path`, and `@loader_path`.
      * System and package-manager libraries, existing relative entries, and extensions under `site-packages` are left unchanged. The existing ELF relocation path is unchanged.
      * `pyenv binary package` can now produce macOS archives for standard non-framework CPython builds.
      * Package-manager-specific dependency validation remains separate work; generated macOS definitions currently warn when no `ldconfig` cache is available.

  ### Tests

      * Added coverage for macOS metadata, generated-definition quoting and tool checks, Mach-O dependency and `LC_RPATH` replacement, idempotence, failure handling, and `site-packages` exclusion.
      * The full `pyenv-binary` suite passed on macOS with Bash 3.2: 61 passing tests and one expected Linux-specific `ldconfig -p` integration skip.
      * Saved CPython 3.13.14 and installed it through a generated `python-build` definition into two different prefixes on macOS 26.5.2 arm64.
      * Verified native-module imports, `venv`, `ensurepip`, removal of all original-prefix Mach-O references, and valid signatures for all 79 runtime Mach-O files.

---
## Summary by cubic
Supports relocating macOS archives in `pyenv-binary`; previously macOS archives were rejected, now they can be packaged, installed, and relocated with Mach-O load paths rewritten.

- `pyenv binary save` records the original build prefix in metadata, and `pyenv binary generate-installer` requires it for macOS definitions.
- macOS relocation uses `otool` and `install_name_tool` to rewrite bundled-library dependencies, install IDs, and rpaths to `@rpath`, `@executable_path`, and `@loader_path`; files without an install ID (like extension modules) are handled gracefully.
- System libraries, existing relative paths, and `site-packages` extensions are left unchanged; the ELF relocation path still uses `patchelf`.
- `pyenv binary package` no longer exits early on macOS and can now produce non-framework CPython archives.

<sup>Written for commit 5cd4ea56d21d9490c15f9ac5553c944b9d785b69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3528?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
+++ b/plugins/pyenv-binary/libexec/pyenv-binary-generate-installer
@@ -55,7 +55,7 @@ if [ -z "$metadata" ] || [ -z "$archive_url" ]; then
   exit 1
 fi
 
-version="" os="" arch="" distro="" libc="" archive=""
+version="" os="" arch="" distro="" libc="" build_prefix="" archive=""
 deps=""
 while IFS='=' read -r key value; do
   case "$key" in
@@ -64,6 +64,7 @@ while IFS='=' read -r key value; do
   arch ) arch="$value" ;;
   distro ) distro="$value" ;;
   libc ) libc="$value" ;;
+  build_prefix ) build_prefix="$value" ;;
   archive ) archive="$value" ;;
   dep ) deps="${deps:+$deps }$value" ;;
   esac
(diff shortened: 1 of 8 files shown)


native-api (commented):
My machine that I used to run MacOS VMs on is out of order so I cannot make sense of this PR now. I assume that you've tested it so let's just roll with it. If there are any deficiencies, they will come up later at some point.

macayu17:
Yeah I tested it on one of my friend's macbook.

PR #3538, opened 2026-09-08
Incorporate the `pyenv-link` plugin

macayu17:
### Prerequisite

  - [x] Please consider implementing the feature as a hook script or plugin as a first step.
    - This imports the existing `pyenv-link` plugin into pyenv with its original history and license preserved.
  - [x] My PR addresses the following pyenv issue (if any)
    - Part of #2334. It does not close the issue.

  ### Description

    - Imports `real-yfprojects/pyenv-link` under `plugins/pyenv-link` while preserving its original commit history and authorship.
    - Extends `pyenv link version <path> [<name>]` so a platform-specific binary installation can be linked under a plain version name.
    - Uses relative links for installations already under `$(pyenv root)/versions`, so they continue working if `PYENV_ROOT` moves. External virtual environments keep absolute targets.
    - Refuses existing or dangling destinations and rejects names that cannot safely represent a pyenv version.
    - Keeps the existing virtual-environment workflow and command structure.

  ### Tests

    - Added Bats coverage for completion, argument validation, binary aliases, external paths containing spaces, dangling destinations, and invalid names.
    - The focused tests passed with Bash 3.2 and 4.1 using both BusyBox and GNU tools.
    - The full core suite passed in all four Windows/Docker environments: 261 tests in each.
    - The full core suite passed on macOS 26.5.2 arm64 under Bash 3.2: 253 passed with 8 environment-specific skips.
    - A native macOS runtime check confirmed that the linked version resolves through `pyenv exec`.

---
## Summary by cubic
Imports the existing `pyenv-link` plugin and extends `pyenv link version <path> [<name>]` so platform-specific binary installations can be linked under a plain version name. Part of #2334 but does not close it.

- Preserves the plugin's original commit history and MIT license.
- Links installations already inside `$(pyenv root)/versions` relatively so they stay valid if `PYENV_ROOT` moves; external virtual environments keep absolute targets.
- Refuses existing or dangling destinations and rejects names that cannot safely represent a pyenv version.
- Adds Bats tests for completions, argument validation, binary aliases, external paths with spaces, dangling destinations, and invalid names.
- Wires the plugin tests into the existing `Makefile` test and Docker targets, and fixes test PATH setup so plugin suites resolve the pyenv root properly.
- Fixes the PowerShell shell integration so pyenv is resolved once, avoiding a "pyenv pyenv" error when it appears multiple times on PATH.

<sup>Written for commit 88e9ab13e8fd603c732a4f43f1ff3d1c2d264b6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pyenv/pyenv/pull/3538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 yfprojects
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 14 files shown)


test/link.bats
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bats
+
+load test_helper
+
```
native-api: Why is this file in the core tests? It should be under `pluging/pyenv-link/test`, shouldn't it?

native-api (approved):

Issue #2852, opened 2023-11-26
"use openssl@<...> from homebrew" is displayed even if OpenSSL to use has been specified explicitly


Fares-Abubaker:
I am trying to install Python and build it using OpenSSL 3.1, but it always using OpenSSL 1.1 
I tried many ways to ensure that I configure OpenSSL 3.1 but nothing changed

I am using Mac M1 with Homebrew
latest version of pyenv and python-build

```
❯ CONFIGURE_OPTS="--with-openssl=$(brew --prefix openssl@3.1) --with-openssl-dir=$(brew --prefix openssl@3.1) --with-openssl-include=$(brew --prefix openssl@3.1)/include --with-openssl-lib=$(brew --prefix openssl@3.1)/lib" pyenv install 3.11.6

python-build: use openssl@1.1 from homebrew
python-build: use readline from homebrew
...
```

Is there an issue or what is the problem ?!

native-api:
Confirmed, the `use_homebrew_openssl()` function does not detect if OpenSSL-related arguments have been passed.

As a workaround, set envvar `PYTHON_BUILD_HOMEBREW_OPENSSL_FORMULA=<Homebrew formula name>`  to manually specify the Homebrew formula to pick.

native-api:
Though, user-supplied Configure options are put after Pyenv-supplied ones, so your command should actually be building against OpenSSL 3.1 despite that message.

Fares-Abubaker:
Great news @native-api thanks

but this shouldn't be fixed ?

native-api:
Yes, it should.

Fares-Abubaker:
in fact this worked now:

```
❯ PYTHON_BUILD_HOMEBREW_OPENSSL_FORMULA=openssl@3.1 pyenv install 3.10.13 

python-build: use openssl@3.1 from homebrew
python-build: use readline from homebrew
...
```

Fares-Abubaker:
i suggest to add this to the readme

Issue #3399, opened 2026-02-03
Building EOL Python versions, and locating/building dependencies for non-EOL versions, in a convenient way


safulop:
I have successfully used pyenv to install python 3.7 on this system, where it is running great.  I tried to install python 3.4.7 and I got these errors

Installing Python-3.4.7...
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/safulop/.pyenv/versions/3.4.7/lib/python3.4/curses/__init__.py", line 13, in <module>
    from _curses import *
ImportError: No module named '_curses'
WARNING: The Python curses extension was not compiled. Missing the ncurses lib?
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/safulop/.pyenv/versions/3.4.7/lib/python3.4/ssl.py", line 97, in <module>
    import _ssl             # if we can't import it, let the error propagate
ImportError: No module named '_ssl'
ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?

Clearly the ncurses devel package is actually installed so, I am at a loss to figure this out.

native-api:
Please add the necessary diagnostic details as per the issue template.

My best guess for now is that since this is an old Python version, your library versions may be incompatible with it.

safulop:
I'm sorry to be such a newb, but I don't know where the issue template is here, to provide the desired diagnostic details.

native-api:
When you create a new issue with a type "bug report", you get https://github.com/pyenv/pyenv/blob/master/.github/ISSUE_TEMPLATE/bug_report.md (the part under the table) as the template.

safulop:
Thanks.  I am now in the process of setting up a new Fedora system on a new computer.  I am hoping to get it set up in such a way that pyenv can be used to install older Python versions as smoothly as possible, do you have any particular recommendations in this regard?  Like, how I should install the system Python, and then pyenv, to ensure maximum compatibility with older Pythons as far back as 3.4?

native-api:
As a rule, to build EOL versions that are incompatible with your system, you'll have to:

* specify additional build arguments (via envvars, see https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#special-environment-variables)
  * in particular, compiler/linker/configure flags to link to older versions of dependencies if the current ones are incompatible
* possibly, also apply patches if the core code has bits that won't work in the current environment.

The easiest way to set envvars conditionally would be via a plugin:

[...]

PR #3541, opened 2026-09-11
python-build: honor explicitly configured OpenSSL on macOS

aprylewu:
When a macOS build supplies `--with-openssl`, Python-Build still selects an OpenSSL installation from Homebrew and prepends its include, library, and runtime search paths. The explicit configure option survives, but the injected flags can refer to a different installation.

Honor explicit OpenSSL options before the broken-system fallback and at the standard-build dependency-discovery call sites. When there is no explicit option, the standard Python build now probes Homebrew/MacPorts for OpenSSL alongside its other dependencies, including definitions that do not call `has_broken_mac_openssl`. The low-level selectors perform discovery when called directly; the explicit-option guard belongs at their call sites.

Recognize both argument forms in `CONFIGURE_OPTS`, `PYTHON_CONFIGURE_OPTS`, and definition-provided package options. A standalone `--with-openssl-rpath` still allows automatic selection. General dependency-prefix flags are unchanged.

### Prerequisite

* [x] Considered a hook or plugin. This corrects built-in dependency selection before it alters compiler flags or installs bundled OpenSSL.
* [x] Closes https://github.com/pyenv/pyenv/issues/2852.

### Tests

* [x] Regressions run the real `python-build` entry point with tiny local source fixtures and stubbed build commands. They cover explicit configuration, standard-build discovery without the broken-system predicate, repeated discovery, rpath-only options, and bundled fallback.
* [x] The two new standard-build discovery regressions fail on the previous PR head and pass after this update. All 14 OpenSSL-focused cases pass.
* [x] All 59 cases in `plugins/python-build/test/build.bats` pass on Bash 3.2 with Bats 1.10.0. Existing library tests explicitly account for the new OpenSSL probes.
* [x] Darwin tests provide both `uname` and `sw_vers` stubs. A local simulation of the missing macOS command reproduces the Linux CI failures before the fix and passes afterward.
* [x] Complete `make test`: 474 passed, 9 skipped (unavailable fish/PowerShell integrations and the Linux-only `ldconfig -p` case).
* [x] Bash 3.2 syntax check with `extglob` enabled and `git diff --check`.

The fixtures inspect configure arguments and compiler flags; no CPython or OpenSSL compilation is performed locally.

Implementation and regression tests were prepared with assistance from OpenAI Codex.


native-api (commented):
Will this also prevent picking OpenSSL from Homebrew/MacPorts?
IIRC after `has_broken_mac_openssl`, OpenSSL search triggers again at some point.

FWIW, scripts aren't required to call `has_broken_mac_openssl` at all.

aprylewu:
Good point: the original guard only covered `has_broken_mac_openssl`. I traced the later standard-build steps: they call the general `use_homebrew`/`use_macports` dependency-prefix setup, not the OpenSSL selectors again. But a definition can call either OpenSSL selector directly, bypassing that guard.

Commit b567c52 factors out the option check and applies it to both `use_homebrew_openssl` and `use_macports_openssl` too. They return success without probing or modifying options/flags when OpenSSL is already configured, matching the existing readline convention and avoiding fallback to another installation. The predicate still skips the bundled OpenSSL build. General dependency-prefix flags remain unchanged; this specifically prevents OpenSSL selection and its extra arguments/flags.

Added four regressions for direct calls without the predicate and repeated calls after it, covering both package managers. All four fail on the previous head and pass now. All 10 OpenSSL cases and the full suite pass on Bash 3.2 (470 passed, 9 skipped); no CPython/OpenSSL compilation was needed.

plugins/python-build/bin/python-build
```diff
@@ -1725,6 +1729,7 @@ has_broken_mac_openssl() {
 }
 
 use_homebrew_openssl() {
+  configured_with_openssl && return 0
```
native-api: This should be 1, no?

plugins/python-build/bin/python-build
```diff
@@ -1705,13 +1705,17 @@ build_package_mac_readline() {
...
 }
 
-has_broken_mac_openssl() {
-  is_mac || return 1
-  # Respect an explicit OpenSSL location instead of adding flags for another one.
+configured_with_openssl() {
   local arg
   for arg in ${CONFIGURE_OPTS} ${PYTHON_CONFIGURE_OPTS} "${PYTHON_CONFIGURE_OPTS_ARRAY[@]}"; do
-    [[ "$arg" = --with-openssl || "$arg" = --with-openssl=* ]] && return 1
+    [[ "$arg" = --with-openssl || "$arg" = --with-openssl=* ]] && return 0
```
native-api: Replace with `=~ --with-openssl[ =]` ?

plugins/python-build/bin/python-build
aprylewu: Yes, changed both to return `1` in f5c262e. `has_broken_mac_openssl` still has its own guard to skip the bundled build.

plugins/python-build/bin/python-build
aprylewu: Switched to `^--with-openssl(=|$)` per argument. That also handles `--with-openssl /path`, while keeping `--with-openssl-rpath` out.

native-api (commented):
Looking through the code, I confirmed that indeed, OpenSSL search is indeed currently only triggered from `has_broken_mac_openssl`.
We probably need to do that more like the Tcl-related code does.
Maybe even refactor all OpenSSL search logic into a subroutine like https://github.com/pyenv/pyenv/issues/3399 suggests.

aprylewu:
Agreed. OpenSSL discovery being a side effect of `has_broken_mac_openssl` is harder to follow than the Tcl/Tk build step.

I read through #3399. A shared selection routine needs to preserve the version constraints, source-build fallback and custom definitions. I'd suggest keeping that refactor separate and this PR focused on honoring explicit `--with-openssl` options. If you'd prefer the OpenSSL extraction first, I can adjust this PR.

plugins/python-build/bin/python-build
native-api:
```suggestion
if can_use_homebrew; then
configured_with_openssl || use_homebrew_openssl || true
use_custom_tcltk || use_homebrew_tcltk || true
```

native-api:
Here's the design I had in mind: add probing of OpenSSL within `package_build_standard` in the same fashion as other libraries:

```diff
 plugins/python-build/bin/python-build | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/plugins/python-build/bin/python-build b/plugins/python-build/bin/python-build
index 6a651ed0..0a802722 100755
--- a/plugins/python-build/bin/python-build
+++ b/plugins/python-build/bin/python-build
@@ -873,6 +873,7 @@ build_package_standard_build() {
 
   if [ "$package_var_name" = "PYTHON" ]; then
     if can_use_homebrew; then
+      configured_with_openssl || use_homebrew_openssl || true
       use_custom_tcltk || use_homebrew_tcltk || true
       use_homebrew_readline || true
       use_homebrew_ncurses || true
@@ -885,6 +886,7 @@ build_package_standard_build() {
       fi
     fi
     if can_use_macports; then
+      configured_with_openssl || use_macports_openssl || true
       use_custom_tcltk || true
       use_macports_readline || true
       use_macports_ncurses || true
@@ -1729,7 +1731,6 @@ has_broken_mac_openssl() {
 }
 
 use_homebrew_openssl() {
-  configured_with_openssl && return 1
   can_use_homebrew || return 1
   command -v brew >/dev/null || return 1
   for openssl in ${PYTHON_BUILD_HOMEBREW_OPENSSL_FORMULA:-openssl}; do
@@ -1768,7 +1769,6 @@ use_homebrew_openssl() {
 }
 
 use_macports_openssl() {
-  configured_with_openssl && return 1
   can_use_macports || return 1
   command -v port >/dev/null || return 1
   local port_location="$(command -v port)"
```

aprylewu:
Thanks, that makes the intended call-site behavior clear. I've applied the layout you suggested in b8f41fd00e: the standard Python build now performs guarded Homebrew/MacPorts OpenSSL discovery, and the low-level selectors no longer contain the explicit-option guard. The guard in `has_broken_mac_openssl` remains so an explicit installation still skips the bundled build.

I replaced the direct-selector tests with standard-build regressions covering explicit options, discovery without a preceding broken-system check, and avoiding a second probe when the earlier selection has already added `--with-openssl`. The two new discovery cases fail on the previous head and pass with this update. Existing dependency tests now explicitly expect the additional OpenSSL probes.

I also traced the failing Linux jobs to four tests that mocked Darwin but omitted `sw_vers`. The tests now stub both commands; the missing-command scenario is reproduced locally before the fix and passes afterward. The full `build.bats` file passes all 59 cases on Bash 3.2 with Bats 1.10.0. No Python or OpenSSL compilation was needed.

The complete `make test` run also passes: 474 passed, 9 skipped (unavailable fish/PowerShell integrations and the Linux-only `ldconfig -p` case).

native-api:
Okay, that does it! Thank you for your contribution!

aprylewu:
Thanks for the review and for clarifying the design. Glad to see this merged!

